PR #42, opened 2015-04-29
Adding mouseOut and mouseOver native events

jayudey-wf:
While working with the dart react test utils it was discovered that it was necessary to simulate some native events in order to initiate a mouseEnter or mouseLeave event during testing.

In this work, the necessary events events were added in order to create those events within the dart react test utils.

issue originally documented - http://stackoverflow.com/questions/24140773/could-not-simulate-mouseenter-event-using-react-test-utils

### DIFF
--- a/lib/react_test_utils.dart
+++ b/lib/react_test_utils.dart
@@ -16,6 +16,8 @@ JsObject _TestUtils = _getNestedJsObject(
 
 JsObject _Simulate = _TestUtils['Simulate'];
 
+JsObject _SimulateNative = _TestUtils['SimulateNative'];
+
 _getNestedJsObject(
     JsObject base, List<String> keys, [String errorIfNotFound='']) {
   JsObject object = base;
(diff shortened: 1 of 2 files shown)


hleumas:
I am not sure @jayudey-wf , isn't this already solved by https://github.com/facebook/react/pull/1366 ? And how does SimulateNative work in the react code exactly? Does it support only these two methods?

I haven't dived into the topic deeply, so I would appreciate if you could provide better explanation of the situation.

Thanks:)

jayudey-wf:
I don't believe that https://github.com/facebook/react/pull/1366 will solve my issue.  That was an update to the way that the react synthetic event is handled. 

Synthetic events are the result of using `Simulate` and are created from a cross browser wrapper of an event, where as the `SimulateNative` events are generic browser events.  Synthetic events are used within React and allow it to be more browser compatible ( https://facebook.github.io/react/docs/events.html ).

The component that I am testing however responds to an actual browser emitted event which is why I need a `SimulateNative` event rather then a `Native` synthetic event.

I went in and added as well what seem to be all the supported native events.

hleumas:
@jayudey-wf thanks for the patch, merged and published.

PR #44, opened 2015-05-05
Ref errors on null

dominicfrost-wf:
### PROBLEM
If you try to access a ref that doesn't exist, this code will error and crash.

### SOLUTION
Add a null check for ref before trying to access properties on it.

### TESTING
1. Create a component.
2. Try to access a ref that doesn't exist on the component.
3. Verify that there is no error and you get null back.

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -90,6 +90,7 @@ ReactComponentFactory _registerComponent(ComponentFactory componentFactory, [Ite
 
     var getRef = (name) {
       var ref = jsThis['refs'][name] as JsObject;
+      if (ref == null) return null;
       if (ref[PROPS][INTERNAL] != null) return ref[PROPS][INTERNAL][COMPONENT];
       else return ref.callMethod('getDOMNode', []);
     };


hleumas:
@dominicfrost-wf isn't it good that you get error when accessing bad ref? I mean, program should rather crash than silently not working, doesn't it?

dominicfrost-wf:
It will still crash, just in the consumers application instead of this library. If someone wants a ref, they are going to use it for something. And if it's null they will know something is wrong. I have since fixed this another way in our application, but I think it's still a good practice to null check objects that could potentially be null before trying to access and of their properties.

hleumas:
@dominicfrost-wf I am ok with the null check, I just think the method should throw descriptive exception instead of returning `null`.

dominicfrost-wf:
Reactjs will return undefined if you try to access a ref that doesn't exist. It seems weird to change that behavior in the dart wrapper to me. But if you would rather throw an error that is fine by me, since we have solved the issue another way.

clairesarsam-wf:
@hleumas Using this library, I generally expect consistency with ReactJS behavior. I've run into this issue a few times, and would rather have a `null` return value than an exception.

hleumas:
@dominicfrost-wf @clairesarsam-wf Do you have valid use case for returning `null` instead of throwing an exception? By valid use case I mean a situation when you want to access undefined ref and it is not just bug in your program. If there are valid use cases for this, than I am open to discussion.

matystl:
fixed in https://github.com/cleandart/react-dart/pull/50

PR #45, opened 2015-05-11
return JSObject ref if appropriate

trentgrover-wf:
## Problem
- If a JS based react component is `ref`ed, the associated DOM element is returned instead of the actual JSObject.  This is useful in many cases, but prevents access to any other public methods available on the JS component.
- I realize this would be a breaking change, so if you'd prefer to expose this via some other new API call, that's a reasonable possibility as well.

## Changes
- Return the JSObject from getRef.

## Testing
- DOM node for a JS based component can still be accessed via:
```
    var whateverRef = this.ref('whateverRef');
    var domNode = whateverRef.callMethod('getDOMNode');
```
- Public API functions are also accessible on JS based components if exposed (like in some react-bootstrap components):
```
    var whateverRef = this.ref('whateverRef');
    whateverRef.callMethod('setDropdownState', [false]);
```

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -118,7 +118,7 @@ abstract class Component {
 
 }
 
-/** Syntetic event */
+/** Synthetic event */
 
 class SyntheticEvent {
 
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -8,6 +8,7 @@ import "package:react/react.dart";
 import "dart:js";
 import "dart:html";
 import "dart:async";
+import 'dart:mirrors';
 
 var _React = context['React'];
 var _Object = context['Object'];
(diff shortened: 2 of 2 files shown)


hleumas:
@trentgrover-wf Yeah, I agree that refs should always return a component and the current behavior of returning associated DOM element for js component is not good.

However, I think the proper fix should have these properties:

1. Element returned by the `ref` always implements the `Component` class.

2. It should at least support some basic operations for DOM components. Let's say by creating some `DOMComponent` class?

What do you think @trentgrover-wf ? This could also possibly solve https://github.com/cleandart/react-dart/pull/44 in a canonical way.

trentgrover-wf:
@hleumas Something like this?  This code works fine for my use case if you prefer.

The only negative is that it has to introduce mirrors to route the component API methods to the underlying `callMethod` if `noSuchMethod` is tripped.

This allows me to use this code within a component:
```
    var whateverRef = this.ref('whateverRef');
    whateverRef.setDropdownState([false]);
```

hleumas:
Ok, this is what I would propose:

After fixing issues #47 and #48, we don't need `getDomNode()` (it is deprecated in React).

We can ignore the Mirrors API and proceed with just `whateverRef['setDropDownState']([false])`;

What do you think?

matystl:
fixed in https://github.com/cleandart/react-dart/pull/50 it's also published read about findDOMNode in readme

PR #50, opened 2015-05-14
findDOMNode and upgrade to react 13.3

matystl:


README.md
```diff
@@ -128,6 +128,39 @@ void main() {
...
 }
 ```
+## Using refs and findDOMNode
+
+Using refs is little bit different that in react. You can specifi ref name in component props
+and then call ref method to get referenced element. Return value is different for dart components and for
+javascript components. For dart component you get instance of dart class for that component. 
+For js components(like dom elements) you get instance of jsObject representing that react component.
+If you want to work with dom nodes instead of component you can call top level method findDOMNode on anything that ref returns.
+
```
jurom: Proper usage of refs here is a little bit different from usage in react. You can specify a ref name in component props and then call ref method to get the referenced element. Return value for dart components and for javascript components is different. For dart component, you get an instance of dart class of the component. For js components (like DOM elements), you get instance of jsObject representing the react component. If you want to work with DOM nodes instead of a component, you can call top level method findDOMNode on anything the ref returns.

README.md
```diff
@@ -128,6 +128,39 @@ void main() {
...
+```
+var DartComponent = registerComponent(() => new _DartComponent());
+class _DartComponent extends Component {
+  var someData = 11;
+  render() => div({});
+}
+var ParrentComponent = registerComponent(() => new _ParrentComponent());
+class ParrentComponent extedns Component {
+  render() =>
```
jurom: Parent

example/test/get_dom_node_test.dart
```diff
@@ -2,22 +2,49 @@ import "package:react/react.dart" as react;
...
+
   componentWillUnmount() => print("unmount");
-  
+
   componentDidMount(root) {
-    print("getDomNode is same as mounted element ${root == getDOMNode()}");
+    customAssert("ref to span return span ", (react.findDOMNode(ref("refToSpan")) as SpanElement).text == "Test");
+    customAssert("getDomNode is same as mounted element", root == getDOMNode());
+    customAssert("findDOMNode work on this", root == react.findDOMNode(this));
+    customAssert("random ref resolute to null", react.findDOMNode(ref("someRandomRef")) == null);
```
jurom: "resolves" ?

example/test/get_dom_node_test.dart
```diff
@@ -2,22 +2,49 @@ import "package:react/react.dart" as react;
...
-  
+
   componentWillUnmount() => print("unmount");
-  
+
   componentDidMount(root) {
-    print("getDomNode is same as mounted element ${root == getDOMNode()}");
+    customAssert("ref to span return span ", (react.findDOMNode(ref("refToSpan")) as SpanElement).text == "Test");
+    customAssert("getDomNode is same as mounted element", root == getDOMNode());
+    customAssert("findDOMNode work on this", root == react.findDOMNode(this));
```
jurom: "works"

lib/react_client.dart
```diff
@@ -511,7 +512,15 @@ bool _unmountComponentAtNode(HtmlElement element) {
   return _React.callMethod('unmountComponentAtNode', [element]);
 }
 
+dynamic _findDomNode(component) {
+  // if we have dart component class so geting mounted dom is returned from getDOMNode
```
jurom: if it's a dart component class, the mounted dom is returned from getDOMNode

lib/react_client.dart
```diff
@@ -511,7 +512,15 @@ bool _unmountComponentAtNode(HtmlElement element) {
...
+  // which has jsComponent closured inside and call on it findDOMNode
```
jurom: "calls" - if the jsComponent calls the findDOMNode.

PR #58, opened 2015-07-06
Permit variadic children (no array necessary)

trentgrover-wf:
## Issue
- react-dart erroneously displays the following console warning at all times because it forces all children to be specified in a Dart list / JS array:
```
Warning: Each child in an array or iterator should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
```

## Changes
**Source:**
- Modified `ReactComponentFactoryProxy` implementation with a `noSuchMethod` that passes an arbitrary list of child arguments on to react.js as a list of arguments, rather than a single array argument.

## Testing
- Existing implementations of components with `null` children, a single child, or a single array of children, all work as expected (no functional change)
- Components with multiple children (not specified as an array) render correctly, without the above warning:
```
react.div({},
  react.span({}, 'a'),
  react.span({}, 'b'),
  react.span({}, 'c')
)
```
- A component with a single array child emits the above warning if no keys were specified on the child elements
- A component with a single array child DOES NOT emit the above warning if keys were specified on the child elements
- When testing, I saw no performance increase or decrease after these changes, regardless of which mechanism was used to specify component children.

## Code Review
@maxwellpeterson-wf
@evanweible-wf
@greglittlefield-wf 
@lancefisher-wf


lib/react_client.dart
```diff
@@ -36,19 +36,100 @@ newJsMap(Map map) {
...
+  }
+
+  dynamic noSuchMethod(Invocation invocation) {
+    if ((invocation.memberName == #call) && (invocation.isMethod)) {
+      Map props = invocation.positionalArguments[0];
+      _convertProps(props);
+      List<dynamic> reactParams = [name, newJsMap(props)];
+      List<dynamic> children = invocation.positionalArguments.sublist(1);
+      reactParams.addAll(children);
+      return _React['createElement'].apply(reactParams);
```
thegregorator: Not sure exactly how it's handled under the hood, but [`JsObject.callMethod`](https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/dart:js.JsObject#id_callMethod) might be preferable.
```dart
    return _React.callMethod('createElement', reactParams);
```

lib/react_client.dart
```diff
@@ -36,19 +36,100 @@ newJsMap(Map map) {
...
-  ReactComponentFactoryProxy(this.reactComponentFactory, this._call);
+  ReactComponentFactoryProxy(this.reactComponentFactory);
 
   JsObject call(Map props, [dynamic children]) {
-    return this._call(props, children);
+
+    if (children == null) {
+      children = [];
+    } else if (children is! Iterable) {
+      children = [children];
```
thegregorator: If `children` is a single element/String here, does wrapping it in a `List` before passing it to React still result in a warning?

lib/react_client.dart
```diff
@@ -36,19 +36,100 @@ newJsMap(Map map) {
...
+      Map props = invocation.positionalArguments[0];
+      List<dynamic> children = invocation.positionalArguments.sublist(1);
+      List<dynamic> reactParams = [_generateExtendedJSProps(props, children)];
+      reactParams.addAll(children);
+      return reactComponentFactory.apply(reactParams);
+    }
+    return super.noSuchMethod(invocation);
+  }
+
+  _generateExtendedJSProps(Map props, List<dynamic> children) {
```
thegregorator: The return value of this method could be typed to `JsObject`, correct?

greglittlefield-wf: The return value of this method could be typed to `JsObject`, correct?

lib/react_client.dart
greglittlefield-wf: If `children` is a single element/String here, does wrapping it in a `List` before passing it to React still result in a warning?

lib/react_client.dart
greglittlefield-wf: Not sure exactly how it's handled under the hood, but [`JsObject.callMethod`](https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/dart:js.JsObject#id_callMethod) might be preferable.
```dart
    return _React.callMethod('createElement', reactParams);
```

lib/react_client.dart
```diff
@@ -286,19 +337,7 @@ ReactComponentFactory _registerComponent(ComponentFactory componentFactory, [Ite
...
-      props['style'] = new JsObject.jsify(props['style']);
-    }
-    if (children is Iterable) {
-      children = new JsArray.from(children);
-    }
-    return _React['createElement'].apply([name, newJsMap(props), children]);
-  };
-
-  return new ReactComponentFactoryProxy(_React['DOM'][name], call);
+  return new ReactDomComponentFactoryProxy(name);
```
greglittlefield-wf: Cool. We'll have to update the `react_test_utils` wrapper accordingly, though, so typing continues to work with DOM components:

`react_test_utils.dart`:
```dart
/// Returns the 'type' of a component.
///
/// For a DOM components, this with return the String corresponding to its tagName ('div', 'a', etc.).
/// For React.createClass()-based components, this with return the React class as a JsFunction.
dynamic getComponentType(ReactComponentFactory componentFactory) {
  if (componentFactory is ReactComponentFactoryProxy) {
    return componentFactory.reactComponentFactory['type'];
  }
  if (componentFactory is ReactDomComponentFactoryProxy) {
    return componentFactory.name;
  }
  return null;
}
```

greglittlefield-wf: (Just make sure `test/react_test_utils_test.dart` passes.)

lib/react_client.dart
```diff
@@ -36,19 +36,100 @@ newJsMap(Map map) {
...
+      _convertProps(props);
+      List<dynamic> reactParams = [name, newJsMap(props)];
+      List<dynamic> children = invocation.positionalArguments.sublist(1);
+      reactParams.addAll(children);
+      return _React['createElement'].apply(reactParams);
+    }
+    return super.noSuchMethod(invocation);
   }
+
+  _convertProps(Map props) {
```
greglittlefield-wf: This method can be typed to `void`.

lib/react_client.dart
trentgrover-wf: good call.  based on some other code that changed, this section isn't even necessary

trentgrover-wf:
@greglittlefield-wf comments addressed

lib/react_client.dart
```diff
@@ -36,19 +36,92 @@ newJsMap(Map map) {
...
 class ReactComponentFactoryProxy implements Function {
-  final ReactComponentFactory _call;
   final JsFunction reactComponentFactory;
-  ReactComponentFactoryProxy(this.reactComponentFactory, this._call);
+  ReactComponentFactoryProxy(this.reactComponentFactory);
 
   JsObject call(Map props, [dynamic children]) {
-    return this._call(props, children);
+    List<dynamic> reactParams = [_generateExtendedJSProps(props, children)];
+    reactParams.add(children);
```
greglittlefield-wf: Does this need to be conditionally converted to a `JsArray`, like it is in the DOM component `call` method? https://github.com/cleandart/react-dart/pull/58/files#diff-12ac054bb8b76f35c1035afef1d94396R99

greglittlefield-wf:
FYI @trentgrover-wf: Created a test harness for measuring the performance of DOM element creation, using Lists and variadic noSuchMethod calls: https://github.com/greglittlefield-wf/react_varidic_perf

Here are the results I'm seeing:
screen shot 2015-07-06 at 12 44 54 pm

lib/react_client.dart
trentgrover-wf: no, it doesn't seem to (though I'm honestly not entirely sure why)

greglittlefield-wf: Looks like it does. When `children` of type `List` are passed to React as a non-`JsArray`, React's internals only sees them as a `DartObject`, so it cannot validate them properly. It may also have other negative effects on how React handles `children` internally.

I added test coverage and a fix in this PR: https://github.com/trentgrover-wf/react-dart/pull/1

Revert the changes in `react_client.dart` and rerun `child_key_warning_test`, and you'll see that validation doesn't work properly when you pass children as a `List` to a Dart component.

lib/react_client.dart
```diff
@@ -36,19 +36,92 @@ newJsMap(Map map) {
...
+      reactParams.addAll(children);
+      return reactComponentFactory.apply(reactParams);
+    }
+    return super.noSuchMethod(invocation);
+  }
+
+  JsObject _generateExtendedJSProps(Map props, dynamic children) {
+
+    Map extendedProps = new Map.from(props);
+    extendedProps['children'] = children;
```
greglittlefield-wf: Don't forget that `children` in Dart component props has always been converted to a `List` beforehand, and changing that will be a definite breaking change.

lib/react_client.dart
```diff
@@ -245,40 +318,10 @@ ReactComponentFactory _registerComponent(ComponentFactory componentFactory, [Ite
     )])
   ]);
```
greglittlefield-wf: Note that we _could_ skip the `createFactory` call and just deal with the React component class (return result of `React.createClass`), passing it in as the `type` argument to `createElement` in `ReactComponentFactoryProxy` methods.

This still works fine, just an idea that may or may not be cleaner.

trentgrover-wf: I like the idea of this, but I tried it out and there was a pretty big performance drop (30% slower)

greglittlefield-wf: Huh, wouldn't have expected that. Oh well.

trentgrover-wf: we probably get some sort of JS interop caching advantage out of explicitly calling `createFactory` (I didn't dive down into the `createElement` implementation to find a root cause)

lib/react_client.dart
greglittlefield-wf: Is everyone okay with that?

I also think that this kind of change should be accompanied by the addition of `React.children` utilities in Dart.

I could go either way on with changing `children` to be an opaque data structure like it is in JS, but perhaps we should continue converting it to a `List`, and address that in a separate PR (which may or may not be performance related).

lancefisher-wf:
Thanks for this PR @trentgrover-wf. I would love to get rid of those warnings.

@greglittlefield-wf Thanks for putting together that perf chart. So it looks like this approach is about 7% slower on average, and a couple of cases are even faster (and we're dealing with microseconds).

To get my head around the impact of this, I'm going to break down the case of the custom component with 4 children. This takes 540 microseconds to construct vs 570 microseconds with the new approach.

To keep 60fps, we have about a 16ms (or 16,000 microsecond) budget per frame. This means we can construct ~30 of these "4 children" components in a frame. (As a side note, this seems a little slow)

With the new approach taking 570 microseconds, we could construct 28 of these "4 children" components in a frame.

Do you have any idea what the re-render timings are in comparison to the original creation times?

hleumas:
Very interesting work guys! Thanks! I am gonna review this next week:)

trentgrover-wf:
@hleumas any update?

hleumas:
@trentgrover-wf Sorry for taking me so long to reply. Hard three weeks...

I really like all the work and refactoring and improvements both you and @greglittlefield-wf did. Kudos to you folks!

I am ready to merge this, before doing so I would like you to create new pull request which separates code cleaning, refactoring and new features into distinct commits.

Commit history could for example be:

- instead of merging master branch, please rebase
- factoring out ` _generateExtendedJSProps(Map props, List<dynamic> children)` and `_convertProps`
- creating `ReactDomComponentFactoryProxy` together with updating test utils
- removing unnecessary `children` to `Iterable` conversion

And please fix my comment: https://github.com/trentgrover-wf/react-dart/commit/4ea46afa89481772a980677a8b71236b0c4b17c6#commitcomment-12362676

I promise I will merge as soon as you fix these few issues:)

Thanks again for incredible work and sorry for taking it too long to reply.

greglittlefield-wf:
@hleumas @trentgrover-wf I think we should also address my comment (https://github.com/cleandart/react-dart/pull/58#discussion-diff-33992427) regarding `children` representation before merging these changes.

lib/react_client.dart
trentgrover-wf: I'd vote to keep it like this and accept the breaking change.

What sort of `React.children` utilities do you see as useful?  If so, I'd say to handle that in a separate PR.

trentgrover-wf:
@greglittlefield-wf do you want to handle the cleanup on this 1?

lib/react_client.dart
greglittlefield-wf: _For reference: These changes make `props.children` an opaque data structure, much like [in React JS](https://facebook.github.io/react/tips/children-props-type.html)._

---

Well, having `props['children']` be a List or single item without `React.Children` helpers is doable, just a little more tedious (and requires assumptions in the implementation that don't really leave the representation opaque, but that's a different topic).

##### Iterating over children
These use-case is probably a bit more rare, but the `React.Children.map`/`React.Children.forEach` utility functions  would be easy to implement and make the process a bit less tedious:

```dart
// Before variadic children changes
render() {
  return div({},
      props['children'].map(renderChild)
  );
}
```

```dart
// After variadic children changes
render() {
  dynamic children = props['children'];
  List childrenList = (children is List) ? children : [children];

  return div({},
      childrenList.map(renderChild)
  );
}
```

```dart
// After variadic children changes, with React.children utils
render() {
  return div({},
      react_children.map(props['children'], renderChild)
  );
}
```


##### Counting children
Especially not the end of the world to not have a utility for this, but it'd also be easy to implement:

```dart
// Before variadic children changes
render() {
  List children = props['children'];
  return div({
    'role': children.length > 1 ? 'group' : null
  }, children);
}
```

```dart
// After variadic children changes
render() {
  dynamic children = props['children'];
  return div({
    'role': children is List && children.length > 1 ? 'group' : null
  }, children);
}
```

---

For me, the main concern is making children not always a `List`, given that `react-dart` really only supports flat lists of children (see https://github.com/cleandart/react-dart/issues/60).

Only being able to specify flat lists of children in react-dart makes it simple to concatenate List children, but tedious to do so with children that aren't always Lists. I see this being a pain to refactor.

Before this PR's current changes, adding new rendered items as siblings to children was easy:
```dart
// Assuming props['children'] is always a List (react <0.8)
render() {
  var augmentedChildren = []
    ..add(react.span({}, 'Sibling to children'))
    ..addAll(props['children']);

  return div({}, augmentedChildren);
}
```

After the changes, it becomes a bit more tedious.
```dart
// Assuming props['children'] can be a single child or a List.
render() {
  var augmentedChildren = []
    ..add(react.span({}, 'Sibling to children'));

  dynamic children = props['children'];
  if (children is List) {
    augmentedChildren.addAll(children);
  } else {
    augmentedChildren.add(children);
  }

  return div({}, augmentedChildren);
}
```

Ideally, we'd want to allow for nested children (i.e., fix https://github.com/cleandart/react-dart/issues/60). In that case, we'd especially need the `React.Children` utilities to be able to deal with nested children. For example:
```dart
// Assuming props['children'] can be a single child or a List,
// and assuming children can be nested in Lists: https://github.com/cleandart/react-dart/issues/60
render() {
  return div({},
      react.span({}, 'Sibling to children'),
      props['children']
  );
}

// Workaround until #60 is fixed
render() {
  return div({},
      react.span({}, 'Sibling to children'),
      props['children'] is List ? new JsArray(props['children']) : props['children']
  );
}
```

greglittlefield-wf:
@trentgrover-wf Sure thing. May be a day or two before I have some bandwidth to do so, though.

hleumas:
@greglittlefield-wf @trentgrover-wf hi guys, are you gonna handle the cleanup? Cheers:)

hleumas:
@trentgrover-wf @greglittlefield-wf Btw. guys, doing react conf. http://reactive2015.com, would anyone of you like to come speak about using React with Dart?

greglittlefield-wf:
@hleumas Wow, so sorry, this slipped my mind. I'll have some time today/tonight to clean this up.

greglittlefield-wf:
Regarding the conference, I'd have to pass, but let me know if I could provide anything useful for any react-dart related talks.

greglittlefield-wf:
Refactored PR is here: https://github.com/cleandart/react-dart/pull/65

This PR can be closed.

Issue #61, opened 2015-07-28
Drag n drop bugs, FireFox and IE


grumpy:
In FireFox there is an error in syntheticDataTransferFactory, because the arrays files and types can be null instead of zero length.
In IE there is an error when dragging items from outside the browser, because IE forbids access to effectAllowed in this case.

I have a quick PR that prevents both of these errors.  

I also wanted to discussion the possibility of changing the events, by changing all of the SyntheticEvent classes to each be a simple proxy that wraps an instance if JsObject, instead of copying all of the values from the object during construction.  This was mentioned in #49 as a possible performance improvement, and I think it could improve the code as well.  I am curious if anyone else would see any drawbacks or problems with this idea.

grumpy:
See #62

PR #62, opened 2015-07-28
Fixing drag n drop in FireFox and outside the browser in IE

grumpy:
In FireFox, Dragging would sometimes leave files and types null instead of a 0 length array.
  In IE, Dragging from outside the browser would cause an exception when the property dataTransfer.effectAllowed was accessed.

See #61


hleumas:
Thanks!

PR #63, opened 2015-07-29
Add SVG clipPath element

danschultz:
This adds the missing SVG `clipPath` element.

I also added a `.pub/` to the .gitignore. Seems this file is created when running `pub get`, and probably shouldn't be committed to source control.


danschultz:
Bump :)

danschultz:
thanks!

PR #64, opened 2015-07-30
Add support for renderToStaticMarkup

danschultz:
This adds support for `React.renderToStaticMarkup` which is useful if you want to generate a DOM string without React's ID attributes. Use cases include static site generation, or generating HTML/SVG on the client to send to the server for processing.

I also added support for `renderToString` in the browser. The use-case for this method is probably small in client environments, but it's supported in ReactJS, so thought we should support it here too.

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -311,6 +311,14 @@ var render;
  */
 var renderToString;
 
+/**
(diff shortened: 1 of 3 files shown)


danschultz:
Bump

PR #65, opened 2015-08-13
Permit variadic children (no array necessary) - refactor

greglittlefield-wf:
_This PR is a refactor of changes in https://github.com/cleandart/react-dart/pull/58_

## Issue
react-dart erroneously displays the following console warning at all times because it forces all children to be specified in a Dart list / JS array:
```
Warning: Each child in an array or iterator should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
```

## Changes

* Make ReactComponentFactoryProxy an abstract superclass, with implementations ReactDomComponentFactoryProxy and ReactDartComponentFactoryProxy. (This API consistency is useful in react_test_util and external react-dart-related tools.)
  * Remove unnecessary `new JsArray.from()` conversion for DOM components.

* Implement variadic support for the DOM/Dart component factories via noSuchMethod, in which children may be specified as additional arguments. These call into a React JS in a similar fashion, allowing consumers to avoid key/children warnings.
  * For example:

    ```dart
    // Old usage that required List; logs warnings
    react.div({}, [
      react.span({}, 'a'),
      react.span({}, 'b'),
      react.span({}, 'c')
    ])

    // Variadic usage; no warnings
    react.div({},
      react.span({}, 'a'),
      react.span({}, 'b'),
      react.span({}, 'c')
    )
  ```

* Add tests that verify that key/children warnings are logged only when expected, and not when using the new variadic calling method.

* Pin `unittest` Pub dependency to the version required by all tests currently in the repo.

## Testing
* Verify that existing implementations of components with null children, a single child, or a single array of children, all work as expected (no functional change)
* Verify that unit tests regarding child/key warnings in `child_key_warning_test.html` all pass.
* Verify that components with multiple children (not specified as an array) render correctly, without the above warning

---

@trentgrover-wf 
@maxwellpeterson-wf
@evanweible-wf
@lancefisher-wf

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -59,6 +59,20 @@ class ReactComponentFactoryProxy implements Function {
     return reactComponentFactory.apply(reactParams);
   }
 
+  dynamic noSuchMethod(Invocation invocation) {
+    if (invocation.memberName == #call && invocation.isMethod) {
+      Map props = invocation.positionalArguments[0];
+      List children = invocation.positionalArguments.sublist(1);
+
+      List reactParams = [_generateExtendedJsProps(props, children)];
+      reactParams.addAll(children);
+
+      return reactComponentFactory.apply(reactParams);
+    }
+
+    return super.noSuchMethod(invocation);
+  }
+
   /**
    * Returns a JsObject version of the specified props, preprocessed for consumption by React JS
    * and prepared for consumption by the react-dart wrapper internals.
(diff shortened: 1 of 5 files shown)


lib/react_client.dart
```diff
@@ -41,13 +41,95 @@ newJsMap(Map map) {
...
+  /**
+   * Returns a JsObject version of the specified props, preprocessed for consumption by React JS
+   * and prepared for consumption by the react-dart wrapper internals.
+   */
+  static JsObject generateExtendedJsProps(Map props, dynamic children) {
+    if (children == null) {
+      children = [];
+    } else if (children is! Iterable) {
+      children = [children];
+    }
```
greglittlefield-wf: @trentgrover-wf I noticed after the fact that I had kept this logic of normalizing `children` to a `List` intact, due to the order in which I refactored things.

Are you still of the opinion that it should be removed in this changeset?

It's not hurting things in here, and it would be nice for consumers to be able to switch over without having the functionality of `this.props['children']` being broken.

I'm not against removing it eventually, but I think it'd be better to do it in a separate PR.

trentgrover-wf:
+1

hleumas:
Good job both of you!

PR #66, opened 2015-08-15
Change react_test_util.Simulate method parameters to accept DOM nodes

greglittlefield-wf:
### Ultimate Problem
ReactTestUtils JS Simulate.* and SimulateNative.* supports simulating events on [either React component instances or DOM nodes corresponding to component instances](https://github.com/facebook/react/blob/cd15dc60b8551d72107f068d9e6cebb5c660105d/src/test/ReactTestUtils.js#L430-L442).

The wrappers in react_test_utils.dart, however, type the target of the simulated events to `JsObject`, so it's not possible to pass in an `Element` (DOM node).

### Changes
* Change parameters on react_test_utils Simulate.* and SimulateNative.* events to `dynamic` so that both `JsObject` component instances and `Element` component DOM nodes can be passed in.
* Update unit tests to verify that the simulated events also work when DOM nodes are passed in.

### Testing
Run `pub serve`, navigate to http://localhost:8080/react_test_utils_test.html, and verify that all tests are passing.

### DIFF
--- a/lib/react_test_utils.dart
+++ b/lib/react_test_utils.dart
@@ -49,110 +49,110 @@ typedef bool ComponentTestFunction(JsObject componentInstance);
 /// Provides methods for each type of event that can be handled by a React
 /// component.  All methods are used in the same way:
 ///
-///   Simulate.{eventName}([JsObject] element, [Map] eventData)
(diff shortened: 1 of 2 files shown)


hleumas:
Good job! Thanks!

PR #67, opened 2015-08-15
Pass along the component instance returned by React.render(...)

greglittlefield-wf:
`React.render` in JS returns the rendered component, but react_client doesn't pass along the wrapped result.


hleumas:
Thanks!

PR #69, opened 2015-10-07
Add null-check for DOM event handler props

greglittlefield-wf:
Previously, if a `null` event callback prop was passed into a DOM component, it would throw an error when that event is fired.

Now, `null` callbacks are skipped, preventing the issue.

@hleumas


maxwellpeterson-wf:
+1

trentgrover-wf:
@hleumas can we get a review / merge of this?

PR #70, opened 2015-10-13
jsify children when needed so that Iterables are handled gracefully

greglittlefield-wf:
As part of the variadic children changes, Iterables are no longer automatically proxied when passed to React JS. This fixes that.

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -36,6 +36,17 @@ newJsMap(Map map) {
 }
 
 /**
+ * Returns [children] converted to be passed to React JS.
+ */
+convertChildren(children) {
+  if (children is Iterable || children is Map) {
+    return new JsObject.jsify(children);
+  }
+
+  return children;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


maxwellpeterson-wf:
+1 thank you for fixing this!

lib/react_client.dart
```diff
@@ -77,7 +88,7 @@ class ReactDartComponentFactoryProxy extends ReactComponentFactoryProxy {
   JsObject call(Map props, [dynamic children]) {
     List reactParams = [
       generateExtendedJsProps(props, children),
```
greglittlefield-wf: Actually, `children` should be converted beforehand and passed in here as well so the same result of any `Iterable`s is passed to both React JS and the Dart component. 

I'll update this.

greglittlefield-wf:
`new JsObject.jsify()` is a little drastic, since it does a deep conversion.

So we don't have to worry about the performance/regression implications of that, I've made a new CR with changes that are less major: https://github.com/cleandart/react-dart/pull/71.

Closing this PR...

PR #81, opened 2016-01-05
Callback refs: return the Dart component, not the JsObject instance 

greglittlefield-wf:
## Ultimate Problem
Refs specified as callbacks were called with the `JsObject` instance, not the Dart component.

This behavior is different than that of `Component.ref(...)`, and makes it impossible to use Dart components with callback refs without knowledge of react-dart internals.

## Solution
When a ref is specified as a callback to a Dart component, wrap it in a callback that calls the original with the Dart component.

## Testing 
* Open `example/test/ref_test.html`
* Verify that the examples for both string and callback refs work as expected.

---

@hleumas @trentgrover-wf @maxwellpeterson-wf @aaronlademann-wf

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -42,6 +42,13 @@ typedef JsObject ReactComponentFactory(Map props, [dynamic children]);
 typedef Component ComponentFactory();
 
 /**
+ * The type of a ref specified as a callback.
+ *
+ * See <https://facebook.github.io/react/docs/more-about-refs.html#the-ref-callback-attribute>.
+ */
+typedef _CallbackRef(componentOrDomNode);
+
+/**
  * A Dart function that creates React JS component instances.
  */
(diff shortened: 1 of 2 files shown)


lib/react_client.dart
```diff
@@ -128,7 +135,15 @@ class ReactDartComponentFactoryProxy extends ReactComponentFactoryProxy {
     }
 
     if (extendedProps.containsKey('ref')) {
-      jsProps['ref'] = extendedProps['ref'];
+      var ref = extendedProps['ref'];
+
+      // If the ref is a callback, pass React a function that will call it the
```
trentgrover-wf: typo in the description here somewhere...

greglittlefield-wf: Addressed.

maxwellpeterson-wf:
+1

trentgrover-wf:
+1 - merging

PR #83, opened 2016-01-18
Add null-check to Dart component callback refs

greglittlefield-wf:
When a Dart component is no longer rendered, its callback ref is called with `null`, which caused null errors when passed into `_getComponent`.

A null-check has been added to fix this case.


trentgrover-wf:
+1
merging and publishing the bugfix release

Issue #79, opened 2016-01-04
Upgrade to v0.14.x of React JS


greglittlefield-wf:

clairesarsam-wf:
One of the major changes to [ReactJS 0.14](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#major-changes) is the separation of the API into two packages, `react` and `react-dom`. In upgrading, `react-dart` will either need to mirror that division or keep the same API it has currently.

In the first scenario, `react.dart` would split into `react.dart`, and `react_dom.dart` (containing `render`, `unmountComponentAtNode`, `findDOMNode`, `renderToString`, and `renderToStaticMarkup`). We could deprecate those methods in `react.dart` for a time, to give consumers a period to update their usages. For that period, we would still use `ReactDOM` under the hood for those methods in `react_client.dart` - consumers would receive only Dart deprecation warnings, not the ReactJS deprecation warnings.

In the second scenario, we would simply update `react_client.dart` to use `ReactDOM` appropriately. The top-level API for `react-dart` would not change.

The first option has the advantage of remaining closer to the ReactJS api. This could be useful for future upgrades and for not confusing consumers who are familiar with/referencing documentation for ReactJS.

The second option does not require consumers to update their API usage, and would keep the upgrade in `react-dart` simple.

trentgrover-wf:
I would vote for the first scenario with an interim deprecation period.

If it's possible to introduce the new API while also maintaining a deprecated, but still backward compatible version of the API, we'd get the best of both worlds

maxwellpeterson-wf:
I agree with Trent, the first option sounds good.

greglittlefield-wf:
First option sounds good to me as well—keeping a deprecated version of the old API will make the upgrade go a lot smoother.

greglittlefield-wf:
Note that `renderToString`/`renderToStaticMarkup` should probably go into a separate library, like `react_dom_server.dart`. (See [ReactDOMServer](https://facebook.github.io/react/docs/top-level-api.html#reactdomserver).)

travissanderson-wf:
+1

clairesarsam-wf:
Thanks for the input everybody, splitting into separate libraries overlapped with a deprecation period it is. :+1:

aaronlademann-wf:
@trentgrover-wf this issue can be closed.

PR #86, opened 2016-02-23
Upgrade to ReactJS v0.14.7

clairesarsam-wf:
Summary
---

The primary purpose of this PR is to upgrade `react-dart` to use `ReactJS` v0.14.7. You can view the provided upgrade notes for going from v0.13 to v0.14 [here](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#notable-enhancements).

As an additional improvement, this PR also includes commit 54316f5, which adds exceptions when `React` or `ReactDOM` is not loaded or when `setClientConfiguration` is not called. I thought this would make the life easier for consumers of this library looking to upgrade, since they'll need to load a new resource - `react_dom.js`

Changes Made
---

*  All `.js` files have been upgraded to `React` 0.14.7.
* Two new libraries, `react_dom` and `react_dom_server` have been added. This is to keep API parity with React as discussed in issue #79.
* The top-level API members that are now in `react_dom` and `react_dom_server` have been deprecated in `react`.
* The tests in the `test` directory have been updated.
* The code in `examples` and `examples/test` has been updated, although some of that had pre-existing bugs.
* `README.md` has been updated with the new requirements for loading files and the new ref typing.
* A new SVG element, `image`, has been added since `React` now supports it.
* `initializeTouchEvents` is no longer called in `setClientConfiguration`, since it has been removed from `React` and is no longer necessary.
* Since React no longer uses the same type of object for primitive components as composite components, Dart converts the React objects for primitive components to `Element`s automatically. This is problematic in some cases - primarily the test utility methods that return `JsObject`s, and `render`, which also needs to return a `JsObject`. I'm handling this by converting the `Element` back to a `JsObject` - it still has the keys you'd expect on a `ReactElement` `JsObject` (`_reactInternalComponent`, `props`, `tagName`). I'm open to other ideas about handling this change, though.

Breaking Changes
---

* Primitive component refs now return `Element`s instead of `JsObject`s. This applied to both string and callback refs.
* `ReactDOM` must now be loaded.
* `findAllInRenderedTree` and related testing methods may no longer be passed primitive components. `React` now throws an error explaining they must be provided with a composite component. See [release notes here](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#breaking-changes).

Outstanding Questions
---

`React` is deprecating `this.getDOMNode()`. Should we deprecate `Component.getDOMNode`? We implement it using `ReactDOM.findDOMNode`, so the `React` deprecation produces no warnings. This is more a question of API consistency.

RESOLUTION: All in favor of `getDOMNode` deprecation. implemented in commit 47dfd93.

How to Test
---

Run the tests in the `test` directory. They should pass.

---

FYA:  @greglittlefield-wf @trentgrover-wf @hleumas @travissanderson-wf @evanweible-wf


README.md
```diff
@@ -8,12 +8,13 @@ If you are not familiar with React library read [react tutorial](http://facebook
...
 To integrate in Dart project add dependency [react](https://pub.dartlang.org/packages/react) to pubspec.yaml.
 
-Include native react library (provided with this library for compatibility reasons) to index.html and create element where you'll mount the react component you'll create.
+Include the native javascript `react` and `react_dom` libraries (provided with this library for compatibility reasons) in `index.html` and create element where you'll mount the react component you'll create.
 
 ```html
 <html>
   <head>
     <script async src="packages/react/react.js"></script>
+    <script async src="packages/react/react_dom.js"></script>
```
maxwellpeterson-wf: Perhaps this Dart package should provide a JS file that includes both `react.js` and `react_dom.js` so that consumers of react-dart are not forced to either a). load two separate JS files or b). perform some sort of concatenation build step themselves. In a JS project this would likely be handled with code bundling like webpack or jspm, but generally speaking a Dart consumer would not use/need those tools.

lib/react_test_utils.dart
```diff
@@ -339,21 +355,41 @@ JsObject scryRenderedComponentsWithType(
...
+List scryRenderedDOMComponentsWithTag(JsObject tree, String tagName) {
+  var components = _TestUtils.callMethod(
       'scryRenderedDOMComponentsWithTag', [tree, tagName]);
+
+  var jsObjectComponents = [];
+
+  components.forEach((component) {
+    if (component is JsObject) jsObjectComponents.add(component);
+    else jsObjectComponents.add(new JsObject.fromBrowserObject(component));
+  });
```
evanweible-wf: Could add a util method for this conversion since you use it several times.

evanweible-wf:
Just the one note, otherwise this lgtm! As for your one outstanding question - I'd vote to deprecate `getDOMNode()`. Might as well keep the API as close as possible.

lib/react_dom.dart
```diff
@@ -0,0 +1,27 @@
+library react_dom;
+
+/// Renders a ReactElement into the DOM in the supplied container and return a reference to the component
+/// (or returns null for stateless components).
+///
+/// If the ReactElement was previously rendered into the container, this will perform an update on it and only
+/// mutate the DOM as necessary to reflect the latest React component.
+var render = (component, element) {
```
aaronlademann-wf: Should the `element` param be renamed `container`? Would be nice if the dartdoc comment could reference these params by `[name]`

lib/react_dom.dart
```diff
@@ -0,0 +1,27 @@
...
+  throw new Exception('setClientConfiguration must be called before render.');
+};
+
+/// Removes a mounted React component from the DOM and cleans up its event handlers and state. If no component
+/// was mounted in the container, calling this function does nothing. Returns true if a component was unmounted
+/// and false if there was no component to unmount.
+var unmountComponentAtNode;
+
+/// If this component has been mounted into the DOM, this returns the corresponding native browser DOM element.
```
aaronlademann-wf: s/`element`/`[Element]`

aaronlademann-wf:
Couple really small nits... NICE WORK!!!

lib/react_test_utils.dart
greglittlefield-wf: Yeah, I'd imagine that such a util could be useful for external use as well.

Something like:
```dart
JsObject normalizeReactComponent(dynamic component) {
  if (component is JsObject) return component;
  if (component is Element) return new JsObject.fromBrowserObject(component);
  if (element component == null) return null;

  throw new ArgumentError(component, 'component', 'not a valid ReactComponent');
}
```
Then, you can do:
```dart
components = components.map(normalizeReactComponent).toList();
```

README.md
```diff
@@ -135,11 +137,9 @@ void main() {
 
 Proper usage of refs here is a little bit different from usage in react. You can specify
 a ref name in component props and then call ref method to get the referenced element.
-Return value for dart components and for javascript components is different.
-For dart component, you get an instance of dart class of the component.
-For js components (like DOM elements), you get instance of jsObject representing the
```
greglittlefield-wf: Shouldn't this line talking about non-DOM JS components stay?

greglittlefield-wf:
Besides the outstanding comments, looks perfect!

+1 to deprecating `Component.getDOMNode`.

README.md
clairesarsam-wf: @maxwellpeterson-wf Good idea, do you think that should be provided for both the development and minified versions, or just the minified version?

lib/react_test_utils.dart
clairesarsam-wf: @greglittlefield-wf :+1: Yep, that actually will be useful in our repository.

README.md
clairesarsam-wf: @greglittlefield-wf Where someone is using a non-Dart composite component? I suppose so, I'll add it back.

clairesarsam-wf: @greglittlefield-wf Added in commit cc4c399

clairesarsam-wf:
@greglittlefield-wf @trentgrover-wf @hleumas @travissanderson-wf @evanweible-wf @aaronlademann-wf @maxwellpeterson-wf All feedback has been addressed.

maxwellpeterson-wf:
+1 nice!

lib/react_test_utils.dart
```diff
@@ -31,6 +31,17 @@ _getNestedJsObject(
   return object;
 }
 
+/// Returns [component] if it is already a [JsObject], and converts [component] if
+/// it is an [Element]. If [component] is neither a [JsObject] or [Element], throws an
+/// [ArgumentError].
```
greglittlefield-wf: I think this should include some sort of explanation as to why you would want to do this.

A shortened version what you have in your PR description would be perfect IMO:
> Since React no longer uses the same type of object for primitive components as composite components, Dart converts the React objects for primitive components to Elements automatically. This is problematic in some cases - primarily the test utility methods that return JsObjects, and render, which also needs to return a JsObject. I'm handling this by converting the Element back to a JsObject - it still has the keys you'd expect on a ReactElement JsObject (_reactInternalComponent, props, tagName). I'm open to other ideas about handling this change, though.

clairesarsam-wf: @greglittlefield-wf :+1: Added in commit 3979db6

evanweible-wf:
+1

maxwellpeterson-wf:
+1

greglittlefield-wf:
+1

maxwellpeterson-wf:
Just a note that this should be released in a minor version bump (0.9.0)

hleumas:
+1

trentgrover-wf:
+1
will merge and release

PR #89, opened 2016-04-10
Enable Strong Mode

jacehensley-wf:
## Ultimate problem:
This package was not strong mode compliant (see: [here](https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md) for more explanation).

And it looks like Dart SDK 1.16.0 will include the `dev_complier` so it would be good to get react-dart in compliance sooner rather than later.

## How it was fixed:
- Added `.analysis_options` file (see: [here](https://github.com/dart-lang/sdk/tree/master/pkg/analyzer#configuring-the-analyzer) for more explanation).
- Got rid of strong mode warnings.
  - Most fixes were just making explicit conversions rather than implicit.

## Testing suggestions:
- Pull this down in Atom or WebStrom there should be only one hint which is the use of getDomNode which is deprecated.
  - If not using Atom or WebStrom run `dartanalyzer --strong lib/**.dart`.

## Areas of regression:
N/A

---
FYA: @hleumas @trentgrover-wf
FYI: @aaronlademann-wf @greglittlefield-wf @clairesarsam-wf @joelleibow-wf @dianachen-wf

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -384,7 +384,7 @@ class SyntheticFormEvent extends SyntheticEvent {
 class SyntheticDataTransfer {
   final String dropEffect;
   final String effectAllowed;
-  final List/*<File>*/ files;
+  final List files;
   final List<String> types;
 
   SyntheticDataTransfer(this.dropEffect, this.effectAllowed, this.files, this.types);
--- a/.analysis_options
+++ b/.analysis_options
@@ -1,0 +1,2 @@
+analyzer:
+  strong-mode: true
--- a/example/test/react_test_components.dart
+++ b/example/test/react_test_components.dart
@@ -59,7 +59,7 @@ class _ClockComponent extends react.Component {
 
   getInitialState() => {'secondsElapsed': 0};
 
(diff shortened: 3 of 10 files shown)


lib/react.dart
```diff
@@ -384,7 +384,7 @@ class SyntheticFormEvent extends SyntheticEvent {
 class SyntheticDataTransfer {
   final String dropEffect;
   final String effectAllowed;
-  final List/*<File>*/ files;
+  final List files;
```
jacehensley-wf: In Strong mode you can add parameterized types via comments. And this was causing other issues if left typed.

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

evanweible-wf:
+1

trentgrover-wf:
merging

trentgrover-wf:
@jacehensley-wf I merged this too quickly - it looks like there are 2 failing unit tests here.  can you fix them in another PR and then I'll tag 1.1.0

screen shot 2016-04-11 at 11 55 47 pm

PR #90, opened 2016-04-10
Define _get/_setProperty in JS files vs eval to enable CSP support

greglittlefield-wf:
## Ultimate problem
Using CSP (Content-Security Policy), specifically a directive/value of `script-src 'self'`, disallows the use of `eval`, which causes calls to `getProperty`/`setProperty`, used in client-side react-dart, to fail.

## Solution
Define `_getProperty`/`_setProperty` in included JS files instead of via `eval`.

We may eventually want to use something like [the `js_util` package](https://pub.dartlang.org/packages/js_util), but for now, this is an easy and non-invasive workaround.

## Testing
Verify that basic client-side rendering (e.g., `example/test/react_test.html`) works for all supported combinations of JS files:

```html
<script src="packages/react/react.js"></script>
<script src="packages/react/react_dom.js"></script>
```

```html
<script src="packages/react/react_with_addons.js"></script>
<script src="packages/react/react_dom.js"></script>
```

```html
<script src="packages/react/react_prod.js"></script>
<script src="packages/react/react_dom_prod.js"></script>
```

```html
<script src="packages/react/react_with_react_dom_prod.js"></script>
```

## Areas of regression
Basic client-side rendering

---

@hleumas @trentgrover-wf @maxwellpeterson-wf @evanweible-wf @jacehensley-wf


trentgrover-wf:
+1

maxwellpeterson-wf:
+1

lib/react_client/js_interop_helpers.dart
```diff
@@ -36,12 +53,20 @@ final _GetPropertyFn getProperty = (() {
...
-  ''');
+  try {
+    // If this throws, then the JS function isn't available.
+    _getProperty(new EmptyObject(), null);
+  } catch(_) {
+    throw new _MissingJsMemberError('_setProperty',
+        'Be sure to include React JS files included in this package '
+        '(which has this and other JS interop helper functions included) '
+        'or, alternatively, define the function yourself:\n'
+        '    function _setProperty(obj, key, value) { return obj[key] = value; }'
```
evanweible-wf: This is almost identical to the one above, could you make a util method to perform this check?

evanweible-wf:
@greglittlefield-wf could you add a couple of simple tests just to verify that `_getProperty()` and `_setProperty()` are included in the JS? When upgrading from one version of react to another, it could be easy to accidentally paste over them without realizing it

lib/react_client/js_interop_helpers.dart
greglittlefield-wf: Not sure that I follow how this could be checked using a util method; could you provide an example of what you're thinking?

greglittlefield-wf:
@evanweible-wf Added tests in https://github.com/cleandart/react-dart/pull/90/commits/aac6d2608295f7e2ee78bf89ba95327904da05b5

lib/react_client/js_interop_helpers.dart
evanweible-wf: I just meant you're doing this in two places:
```dart
try {
  // If this throws, then the JS function isn't available.
  _getProperty(new EmptyObject(), null);
} catch(_) {
```

Don't worry about it though, it's pretty small

evanweible-wf:
Thanks for adding those tests @greglittlefield-wf!

+1

maxwellpeterson-wf:
+1

lib/react_client/js_interop_helpers.dart
greglittlefield-wf: Oh! This one is supposed to use `_setProperty`... Copy/paste typo; will fix,

greglittlefield-wf: Fixed in https://github.com/cleandart/react-dart/pull/90/commits/004982773768b035a14427b55397fecd36ba672d

evanweible-wf: 👍

evanweible-wf:
+1

trentgrover-wf:
+1 - merging

PR #91, opened 2016-04-12
Untype the componentClass param of _isElementOfType.

jacehensley-wf:
## Ultimate problem:
When enabling Strong mode in https://github.com/cleandart/react-dart/pull/89 a return type was cast as a specific type while it actually has two valid return types.

## How it was fixed:
- Remove explicit cast
- Remove typing of the parameter.

## Testing suggestions:
- Verify all tests pass

## Areas of regression:
N/A

---
FYA: @hleumas @trentgrover-wf @greglittlefield-wf

### DIFF
--- a/lib/react_test_utils.dart
+++ b/lib/react_test_utils.dart
@@ -190,12 +190,12 @@ external bool isDOMComponent(/* [1] */ instance);
 external bool isElement(dynamic object);
 
 @JS('React.addons.TestUtils.isElementOfType')
-external bool _isElementOfType(dynamic element, ReactComponentFactory componentClass);
+external bool _isElementOfType(dynamic element, dynamic componentClass);
 
 /// Returns true if [element] is a ReactElement whose type is of a
 /// React componentClass.
 bool isElementOfType(dynamic element, ReactComponentFactory componentClass) {
-  return _isElementOfType(element, getComponentType(componentClass) as ReactComponentFactory);
+  return _isElementOfType(element, getComponentType(componentClass));
 }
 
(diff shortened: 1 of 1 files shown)


trentgrover-wf:
+1

maxwellpeterson-wf:
+1

lib/react_test_utils.dart
```diff
@@ -190,12 +190,12 @@ external bool isDOMComponent(/* [1] */ instance);
 external bool isElement(dynamic object);
 
 @JS('React.addons.TestUtils.isElementOfType')
-external bool _isElementOfType(dynamic element, ReactComponentFactory componentClass);
+external bool _isElementOfType(dynamic element, dynamic componentClass);
 
 /// Returns true if [element] is a ReactElement whose type is of a
 /// React componentClass.
 bool isElementOfType(dynamic element, ReactComponentFactory componentClass) {
```
greglittlefield-wf: Should `componentClass` be `dynamic` here as well?

jacehensley-wf: But that `componentClass` is always a `ReactComponentFactory` right? Maybe it should be called `componentFactory`?

greglittlefield-wf: Ah, yes, you're right. Yeah, I think a naming change would be good.

jacehensley-wf: Cool I'll do that

evanweible-wf:
+1

trentgrover-wf:
+1 merging

PR #97, opened 2016-06-06
Update README

jacehensley-wf:
## Ultimate Problem
The README was using removed methods and incorrect imports.

## Solution
Update to use correct methods and imports.

--
FYA: @hleumas @trentgrover-wf @greglittlefield-wf


evanweible-wf:
+1

trentgrover-wf:
+1
merging docs only change

PR #100, opened 2016-07-14
Support setState Variations

jacehensley-wf:
## Ultimate Problem
ReactJs supports `setState` callbacks as well as a transactional `setState`. 

For more information on what those are see [here](https://facebook.github.io/react/docs/component-api.html#setstate).

## Solution
- Allow `setState`, `replaceState`, and `redraw` to accept an optional parameter that will be called after `componentDidUpdate` is called.
- Untype the `newState` param in `setState` so that is can accept and function that accepts two params.
  - This will be called before `shouldComponentUpdate`.
- Add tests that ensure the ordering of the callbacks and component methods match the JS order.

## Testing Suggestions
- Verify all tests pass.

## Possible Areas of Regression
- None expected, all new logic.

--
FYA: @trentgrover-wf @hleumas @greglittlefield-wf @clairesarsam-wf

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -18,6 +18,16 @@ abstract class Component {
 
   dynamic _jsThis;
 
+  List _setStateCallbacks = [];
+
+  List _transactionalSetStateCallbacks = [];
+
+  /// The List of callbacks to be called after the component has been updated from a call to [setState].
+  List get setStateCallbacks => _setStateCallbacks;
+
+  /// The List of transactional `setState` callbacks to be called before the component updates.
+  List get transactionalSetStateCallbacks => _transactionalSetStateCallbacks;
+
   /// The JavaScript [`ReactComponent`](https://facebook.github.io/react/docs/top-level-api.html#reactdom.render)
   /// instance of this `Component` returned by [render].
   dynamic get jsThis => _jsThis;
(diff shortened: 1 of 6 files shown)


lib/react.dart
```diff
@@ -86,32 +96,59 @@ abstract class Component {
...
   }
 
+  void _addTransactionalSetStateCallback(Map callback(Map prevState, Map currentProps)) {
+    if (_transactionalSetStateCallbacks == null) _transactionalSetStateCallbacks = [];
+
+    _transactionalSetStateCallbacks.add(callback);
+  }
+
+  void _addSetStateCallback(callback()) {
+    if (_setStateCallbacks == null) _setStateCallbacks = [];
```
greglittlefield-wf: Why are these null-checks necessary? I don't see `_setStateCallbacks` or `_transactionalSetStateCallbacks` ever being set to null.

lib/react_client.dart
```diff
@@ -217,6 +218,19 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
     component.transferComponentState();
   }
 
+  _callSetStateCallbacks(Component component) {
+    component.setStateCallbacks.forEach((callback()) { callback(); });
+    component.setStateCallbacks.clear();
+  }
+
+  _callSetStateTransactionalCallbacks(Component component) {
```
greglittlefield-wf: #nit These could be typed to `void`

lib/react.dart
jacehensley-wf: Really "just to be safe" but I can remove them.

test/lifecycle_test.dart
```diff
@@ -374,9 +378,118 @@ void main() {
...
+
+  @override
+  shouldComponentUpdate(_, __) {
+    recordLifecyleCall('shouldComponentUpdate');
+    return props['shouldUpdate'] as bool;
+  }
+
+  Map outerTransactionalSetStateCallback(previousState, __) {
+    recordLifecyleCall('outerTransactionalSetStateCallback');
+    return {'counter': ++previousState['counter']};
```
greglittlefield-wf: This probably shouldn't mutate the previous state; `previousState['counter'] + 1` would be better

test/lifecycle_test.dart
```diff
@@ -374,9 +378,118 @@ void main() {
...
+  }
+
+  Map innerTransactionalSetStateCallback(previousState, __) {
+    recordLifecyleCall('innerTransactionalSetStateCallback');
+    return {'counter': ++previousState['counter']};
```
greglittlefield-wf: This probably shouldn't mutate the previous state; `previousState['counter'] + 1` would be better

greglittlefield-wf:
Couple small things, but otherwise looks great!! Really excited to have these features.

test/lifecycle_test.dart
```diff
@@ -459,12 +459,12 @@ class _SetStateTest extends react.Component {
...
   Map outerTransactionalSetStateCallback(previousState, __) {
     recordLifecyleCall('outerTransactionalSetStateCallback');
-    return {'counter': ++previousState['counter']};
+    return {'counter': previousState['counter'] + 1};
   }
 
   Map innerTransactionalSetStateCallback(previousState, __) {
     recordLifecyleCall('innerTransactionalSetStateCallback');
-    return {'counter': ++previousState['counter']};
+    return {'counter': ++previousState['counter'] + 1};
```
greglittlefield-wf: Forgot to remove the `++` here

lib/react_client.dart
```diff
@@ -217,6 +218,19 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
     component.transferComponentState();
   }
 
+  void _callSetStateCallbacks(Component component) {
+    component.setStateCallbacks.forEach((callback()) { callback(); });
```
trentgrover-wf: is there a reason for the parens to be here?  `forEach((callback())` vs `forEach((callback)`

jacehensley-wf: Not really, that just types it to a function that take no params and returns dynamic.

test/ReactSetStateTestComponent.js
```diff
@@ -0,0 +1,88 @@
+function getUpdatingSetStateLifeCycleCalls() {
```
trentgrover-wf: Is the testing strategy here to implement the same mock-like implementation of a component in both JS and Dart and then verify that they produce the same order of lifecycle event dispatches?

Is there a reason not to just eliminate the JS implementation and compare the Dart output to a static list of known expectations?

jacehensley-wf: Yeah this is pretty much to make sure that they are the same. We could do a static list but then if ReactJs ever changes we would not know about it.

trentgrover-wf:
+1
@greglittlefield-wf can I get a +1 from you too?
@jacehensley-wf this looks to me like a 2.2.0 release - is there any reason that I've overlooked to consider this a breaking change?

greglittlefield-wf:
+1

Sorry, I could've sworn I'd done so already...

jacehensley-wf:
@trentgrover-wf Yeah 2.2.0 sounds good 👍.

jacehensley-wf:
@trentgrover-wf I think this is ready for a merge.

PR #101, opened 2016-07-29
Revert "Support setState Variations"

jacehensley-wf:
## Ultimate Problem
The changes made in https://github.com/cleandart/react-dart/pull/100 were breaking to anyone overriding the `setState`, `replaceState`, or `redraw` because their signatures changed.

This will need to be released under `2.2.1` ASAP. There will another PR shortly to add this functionality back in a major release, `3.0.0`.

## Solution
- Revert the merge into master

## Testing Suggestions
- Verify that all changes have been reverted.

## Possible Areas of Regression
None expected

--
FYA: @trentgrover-wf @hleumas @greglittlefield-wf @clairesarsam-wf

PR #102, opened 2016-07-29
Re-add setState callback support

jacehensley-wf:
## Ultimate Problem
We still want the setState callback variations

## Notes
- This will need to be breaking change release because anyone overriding `setState`, `replaceState`, or `redraw` will have to change the signatures of the overridden methods.

## Solution
- Re add the logic that was originally added in #100 and then removed in #101.

## Testing Suggestions
- Ensure that the changes made here are the same made in #100 
- Same testing suggestions that are in #100 

## Possible Areas of Regression
- None expected

---
FYA: @trentgrover-wf @hleumas @greglittlefield-wf @clairesarsam-wf

### DIFF
--- a/example/test/speed_test.dart
+++ b/example/test/speed_test.dart
@@ -49,10 +49,6 @@ class _Hello extends react.Component {
     });
   }
 
-  void redraw(){
-    setState({});
-  }
-
   render() {
     timeprint("rendering start");
     List<List<String>> data = (props['data'] as List<List<String>>);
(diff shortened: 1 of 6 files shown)


trentgrover-wf:
+1
verified that this change set matches the previously released version:
https://github.com/cleandart/react-dart/compare/89ce5025b88c62e3052d56251c55d374b8e41e91...jacehensley-wf:add_setstate_callbacks/dev

PR #106, opened 2016-10-28
Make adjustments necessary to support strong-mode compliance for over_react

aaronlademann-wf:
In order for [`over_react` to become strong mode compliant](https://github.com/Workiva/over_react/issues/14), a few changes are necessary in the wrapper:
1. In `react.Component`: 
   1. Remove public `props` and `state` fields in-favor of private fields with public getters and setters.
      1. This is so that `over_react` can add strong typing to `props` and `state` within `UiComponent` / `UiStatefulComponent`, respectively without generating Dart analyzer errors.
   2. Type the props / state arguments of lifecycle methods as `Map`

I also cleaned up a few analyzer errors showing in this repo as well.

---

**FYA** @greglittlefield-wf @jacehensley-wf @clairesarsam-wf @trentgrover-wf

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -10,49 +10,40 @@ import 'package:react/src/typedefs.dart';
 /// Top-level ReactJS [Component class](https://facebook.github.io/react/docs/react-component.html)
 /// which provides the [ReactJS Component API](https://facebook.github.io/react/docs/react-component.html#reference)
 abstract class Component {
-  /// ReactJS [Component] props.
-  ///
   /// Currently private since the `@virtual` annotation within the meta package
   /// [doesn't work for overriding fields](https://github.com/dart-lang/sdk/issues/27452).
   ///
   /// TODO: Make public once SDK issue / meta package allows field overriding in strong mode.
-  ///
-  /// See: [props]
-  ///
-  /// Related: [_state]
   Map _props;
+
+  /// Currently private since the `@virtual` annotation within the meta package
+  /// [doesn't work for overriding fields](https://github.com/dart-lang/sdk/issues/27452).
+  ///
+  /// TODO: Make public once SDK issue / meta package allows field overriding in strong mode.
+  Map _state = {};
+
+  /// Currently private since the `@virtual` annotation within the meta package
+  /// [doesn't work for overriding fields](https://github.com/dart-lang/sdk/issues/27452).
+  ///
+  /// TODO: Make public once SDK issue / meta package allows field overriding in strong mode.
+  Ref _ref;
+
+  /// ReactJS [Component] props.
+  ///
+  /// Related: [state]
+  Map get props => _props;
+  set props(Map value) => _props = value;
 
   /// ReactJS [Component] state.
   ///
-  /// Currently private since the `@virtual` annotation within the meta package
-  /// [doesn't work for overriding fields](https://github.com/dart-lang/sdk/issues/27452).
-  ///
-  /// TODO: Make public once SDK issue / meta package allows field overriding in strong mode.
-  ///
-  /// See: [state]
-  ///
-  /// Related: [_props]
-  Map _state = {};
+  /// Related: [props]
+  Map get state => _state;
+  set state(Map value) => _state = value;
 
   /// Returns a component reference.
   ///
   /// * [Component] if it is a Dart component.
   /// * `Element` _(DOM node)_ if it is a React DOM component.
-  ///
-  /// Currently private since the `@virtual` annotation within the meta package
-  /// [doesn't work for overriding fields](https://github.com/dart-lang/sdk/issues/27452).
-  ///
-  /// TODO: Make public once SDK issue / meta package allows field overriding in strong mode.
-  ///
-  /// See: [ref]
-  Ref _ref;
-
-  Map get props => _props;
-  set props(Map value) => _props = value;
-
-  Map get state => _state;
-  set state(Map value) => _state = value;
-
   Ref get ref => _ref;
   set ref(Ref value) => _ref = value;
 
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -5,7 +5,7 @@
 /// A Dart library for building UI using ReactJS.
 library react;
 
-import 'dart:html';
+import 'dart:html' show Element;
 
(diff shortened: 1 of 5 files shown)


lib/react.dart
```diff
@@ -5,14 +5,28 @@
 /// A Dart library for building UI using ReactJS.
 library react;
 
+import 'dart:html' show Element;
```
greglittlefield-wf: `lib/react.dart` is also used for server-side rendering, and should not import `dart:html`

lib/react.dart
```diff
@@ -5,14 +5,28 @@
...
 /// which provides the [ReactJS Component API](https://facebook.github.io/react/docs/component-api.html)
 abstract class Component {
+  Map _props;
+  Map _state = {};
+  dynamic _ref;
+
   /// ReactJS `Component` props.
-  Map props;
+  Map get props => _props;
+  set props(Map value) => _props = value;
```
greglittlefield-wf: As opposed to making these getters/setters, can we just use the `@virtual` annotation instead? (See [meta package](https://pub.dartlang.org/packages/meta) 1.0.4 changelog)

test/child_key_warning_test.dart
```diff
@@ -13,7 +13,8 @@ int _nextFactoryId = 0;
 ///
 /// This prevents React JS from not printing key warnings it deems as "duplicates".
 void renderWithUniqueOwnerName(ReactElement render()) {
-  ReactDartComponentFactoryProxy factory = react.registerComponent(() => new _OwnerHelperComponent());
+  ReactDartComponentFactoryProxy factory =
```
greglittlefield-wf: #nit I think in strong mode the LHS typing is unnecessary with this cast. Also, unnecessary parentheses? Could do:

```dart
final factory =
    react.registerComponent(() => new _OwnerHelperComponent()) as ReactDartComponentFactoryProxy.
```

lib/react.dart
aaronlademann-wf: @greglittlefield-wf I just tried that, and I still get the following errors in over_react:

```
ERROR: Field declaration Component.ref cannot be overridden in UiComponent. ([over_react] lib/src/component_declaration/component_base.dart:106)
ERROR: Field declaration Component.props cannot be overridden in UiComponent. ([over_react] lib/src/component_declaration/component_base.dart:179)
ERROR: Field declaration Component.state cannot be overridden in UiStatefulComponent. ([over_react] lib/src/component_declaration/component_base.dart:230)
```

aaronlademann-wf:
@greglittlefield-wf all feedback addressed except for using the `@virtual` annotation - which for some reason, doesn't seem to work.

lib/react.dart
jacehensley-wf: I believe you will need to be using Dart 1.19.1.

jacehensley-wf: Actually just tried this, still does not work. I found this issue https://github.com/dart-lang/sdk/issues/27452, which makes it seem like strong mode does not allow you to override fields, even with the `@virtual` annotation. I think that may just be for overriding fields with other fields.

greglittlefield-wf: That's too bad. Can we add a code comment, then, explaining that this is for strong mode compliance and adding a TODO linked to that SDK issue?

lib/react.dart
```diff
@@ -8,11 +8,23 @@ library react;
...
+
+  /// ReactJS `Component` state.
+  Map get state => _state;
+  set state(Map value) => _state = value;
 
-  /// Provides access to the underlying DOM representation of the [render]ed `Component`.
-  dynamic ref;
+  /// Returns the component of the specified [ref].
+  ///
+  /// Returns a [Component] if it is a Dart component, otherwise returns an "Dom component" (Dart `Element`).
```
greglittlefield-wf: #nit Doc comments for getters should be noun phrases: https://www.dartlang.org/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases

Not sure why the comment is referencing the getter itself—doesn't really make sense.

Also, this is missing the case for JS components.

What about:
```dart
/// A function that, when called with a String ref, returns the associated component.
///
/// This function will return either:
///
/// * [Component] for Dart components
/// * `ReactElement` for JS composite components
/// * `Element` DOM components
```

lib/react.dart
```diff
@@ -8,11 +8,23 @@ library react;
...
+  dynamic get ref => _ref;
```
greglittlefield-wf: While we're in here, can we make the type more specific, as we did in over_react?
```dart
typedef dynamic _RefTypedef(String ref);

_RefTypedef _ref;

_RefTypedef get ref => _ref;
set ref(_RefTypedef  value) => _ref = value;
```

lib/react.dart
aaronlademann-wf: @greglittlefield-wf good idea.

jacehensley-wf:
+1

aaronlademann-wf:
@greglittlefield-wf ready for a final pass.

lib/react.dart
```diff
@@ -5,14 +5,56 @@
...
+  /// See: [ref]
+  Ref _ref;
+
+  Map get props => _props;
+  set props(Map value) => _props = value;
+
+  Map get state => _state;
+  set state(Map value) => _state = value;
+
+  Ref get ref => _ref;
```
greglittlefield-wf: These getters should have doc comments, since consumers can't see the doc comments of private fields.

What about this:
```dart
/// A private field that backs [ref], which is exposed via getter/setter so
/// it can be overridden in strong mode.
///
/// Necessary since the `@virtual` annotation within the meta package
/// [doesn't work for overriding fields](https://github.com/dart-lang/sdk/issues/27452).
///
/// TODO: Switch back to a plain field once this issue is fixed.
Ref _ref;

...

/// A function that returns a component reference:
///
/// * [Component] if it is a Dart component.
/// * `Element` _(DOM node)_ if it is a React DOM component.
Ref get ref => _ref;
set ref(Ref value) => _ref = value;
```

aaronlademann-wf:
@greglittlefield-wf feedback addressed

greglittlefield-wf:
__+1__, awesome stuff!

aaronlademann-wf:
@trentgrover-wf this is ready for a merge + tag.

PR #107, opened 2016-11-16
Handle IE DataTransfer.dropEffect access failures gracefully

clairesarsam-wf:
# Ultimate Problem

In both Internet Explorer 10 and 11, accessing `DataTransfer.dropEffect` throws an exception for some drag event types.

# Solution

Handle `dropEffect` access failures the same way `effectAllowed` failures are handled.

---

Please review @greglittlefield-wf @jacehensley-wf @aaronlademann-wf

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -551,6 +551,8 @@ SyntheticDataTransfer syntheticDataTransferFactory(events.SyntheticDataTransfer 
     }
   }
   var effectAllowed;
+  var dropEffect;
+
   try {
     // Works around a bug in IE where dragging from outside the browser fails.
     // Trying to access this property throws the error "Unexpected call to method or property access.".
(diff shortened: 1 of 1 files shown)


lib/react_client.dart
```diff
@@ -551,14 +551,25 @@ SyntheticDataTransfer syntheticDataTransferFactory(events.SyntheticDataTransfer
...
     effectAllowed = 'uninitialized';
   }
-  return new SyntheticDataTransfer(dt.dropEffect, effectAllowed, files, types);
+
+  try {
+    // For certain types of drag events in IE (anything but ondragenter, ondragover, and ondrop), this fails.
+    // Trying to access this property throws the error "Unexpected call to method or property access.".
+    dropEffect = dt.dropEffect;
+  } catch (exception) {
+    dropEffect = 'uninitialized';
```
greglittlefield-wf: "uninitialized" is a valid [`effectAllowed` value](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/effectAllowed#Values), but not a [`dropEffect` value](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect#Values).

Should this be `'none'`?

greglittlefield-wf:
+1

PR #111, opened 2016-12-30
Use helpers in package:js/js_util.dart

greglittlefield-wf:
Deprecate the helpers in `js_interop_helpers.dart` and use the equivalent helpers in `package:js/js_util.dart` instead, now that they're available.

Keep `jsify` around (for now) since it doesn't behave the same as the `package:js` version in regards to nested lists and functions.

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -10,8 +10,9 @@ import "dart:collection";
 import "dart:html";
 
 import "package:js/js.dart";
+import 'package:js/js_util.dart' hide jsify;
 import "package:react/react.dart";
-import "package:react/react_client/js_interop_helpers.dart";
+import "package:react/react_client/js_interop_helpers.dart" show jsify;
 import 'package:react/react_client/react_interop.dart';
 import "package:react/react_dom.dart";
 import "package:react/react_dom_server.dart";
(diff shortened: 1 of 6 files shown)


lib/react_client.dart
```diff
@@ -20,7 +21,7 @@ import 'package:react/src/typedefs.dart';
 
 export 'package:react/react_client/react_interop.dart' show ReactElement, ReactJsComponentFactory;
 
-final EmptyObject emptyJsMap = new EmptyObject();
+final dynamic emptyJsMap = newObject();
```
greglittlefield-wf: I was initially thinking that this should be okay due to how loosey-goosey the typing of JS interop objects are.

However, I'm not sure this will work with strong mode... Hmm. Maybe it'd be best to just deprecate this field.

jacehensley-wf:
Can you add a warning from within the JS `_getProperty`/`_setProperty` methods incase, for whatever reason, someone is using that directly? Add also add a TODO to remove them in the next major release?

maxwellpeterson-wf:
+1

greglittlefield-wf:
Holding off on this until the following `getProperty` bug is fixed, or a suitable workaround is found: https://github.com/dart-lang/sdk/issues/28462

PR #110, opened 2016-12-30
Upgrade from `unittest` to `test` package

greglittlefield-wf:
### Ultimate Problem
#78: Tests used the deprecated `unittest` package, and had to be run manually.

### Solution
- Use the `test` package, which allows running tests via command-line and brings other features.
- Split js_interop_helpers tests into separate `_test.dart` files so they can be picked up automatically by `pub run test`

### Testing
* Run `pub run test -p content-shell,chrome`
* Verify that all unit tests run and pass.

Issue #109, opened 2016-12-15
Question - Material UI example


jtrunick:
Is it possible to use a component from the Material UI components? If so, is there an example?

(didn't know where to post this question)

greglittlefield-wf:
Yes, it is possible to use JavaScript React components, though it's not very straightforward at this time... We can use this issue to track the documentation/support around that.

There was some code written to use JS components in an older version of react-dart, but that needs to be updated. I'll see if I can get some code posted early this week.

greglittlefield-wf:
Sorry for the delay! I blame the holidays 😄.

Got together a PR to add support and an example for using JS components with react-dart: https://github.com/cleandart/react-dart/pull/112. A couple tests to add still, but it should be about done

Also, I went ahead and ported one of the Material-UI examples to use react-dart! You can see that here: https://github.com/greglittlefield-wf/material-ui-webpack-example 

Let me know if you have any questions about that code or any feedback regarding this work!

jestan:
Sorry to comment on old issue.
Is there any plan to support this feature?
It will be very useful to reuse existing reactjs components in react-dart/over_react

aaronlademann-wf:
@greglittlefield-wf ^

aaronlademann-wf:
@jestan @GiovanniL19 @Sach97 @ayushin @jtrunick I believe this will be supported / documented in the upcoming react-dart version 5.0.0 _(support for ReactJS 16)_ via #160.

PR #112, opened 2017-01-02
Support component classes declared in JS

greglittlefield-wf:
Not a clean diff: Depends on PRs https://github.com/cleandart/react-dart/pull/110 and https://github.com/cleandart/react-dart/pull/111

---

## Problem
#109 There wasn't a path to use React components declared in JS within react-dart.

## Solution
- Add `ReactJsComponentFactoryProxy` to enable wrapping of JS components
- Add example of using JS components from within Dart
- Add index to `example` directory
- Add common factory proxy tests
- Deprecate `jsify`, add new `jsifyAndAllowInterop`
    - More closely matches `jsify` from `package:js`, and doesn't name-collide
- TODO:
    - tests around:
        - [ ] conditional event handler wrapping
        - [ ] wrapEventHandler
        - [ ] ReactJsComponentFactoryProxy construction

## Testing
- Verify unit tests pass
- Verify example makes sense and works as expected


example/js_components/js_components.dart
```diff
@@ -0,0 +1,81 @@
...
+/// the same way props are converted for DOM components.
+final Foo = new ReactJsComponentFactoryProxy(_Foo);
+
+/// JS interop wrapper class for the component,
+/// allowing us to interact with component instances
+/// made available via refs or [react_dom.render] calls.
+///
+/// This is optional, as you won't always need to access the component's API.
+@JS()
+class FooComponent {
```
jacehensley-wf: Just for my own understanding, this is only necessary if you want a _typed_ interface into your JS component correct? You could still call API methods on the component you just wouldn't get autocompletion or static analysis, right?

greglittlefield-wf: Good question; I'll clarify this in the comment.

For `package:js`, you actually have to declare the interop class to be able to call those methods on it. The typing, however, doesn't matter, so long as it is a `@JS`.

greglittlefield-wf:
This will be added in some form as part of the upcoming react-dart 5.0.0 release

PR #113, opened 2017-01-06
Remove quiver dependency to slim down dependency tree

greglittlefield-wf:
## Problem
quiver is <1.0.0 and breaking changes come more frequently, which requires upkeep of the upper version bound in pubspec.yaml.

Only one utility from quiver is used, so there isn't much benefit to keeping it as a dependency.

## Solution
* Remove quiver dependency
* Replace usage of `enumerate` with equivalent iteration

## Testing
* Verify that `data-reactid` values of siblings increment as just like they do in master in `react_test_server.dart` output:
    ```bash
    pub run example/test/react_test_server.dart | tidy -imq
    ```

    Examples:

    *  `<span data-reactid="foo.bar.0" /> <span data-reactid="foo.bar.1" />` 👍 
    *  `<span data-reactid="foo.baz.$0" /> <span data-reactid="foo.baz.$1" />` 👍

### DIFF
--- a/lib/react_server.dart
+++ b/lib/react_server.dart
@@ -9,7 +9,6 @@ import 'dart:typed_data';
 
 import 'package:react/react.dart';
 import 'package:react/react_dom_server.dart';
-import 'package:quiver/iterables.dart';
 
 // ReactJS constants needed to create correct checksum
 String _SEPARATOR = '.';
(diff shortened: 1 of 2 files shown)


maxwellpeterson-wf:
(nice) +1

evanweible-wf:
+1

jacehensley-wf:
+1

PR #114, opened 2017-01-20
Add support for capturing event handlers

greglittlefield-wf:
From https://facebook.github.io/react/docs/events.html#supported-events
> To register an event handler for the capture phase, append Capture to the event name; for example, instead of using onClick, you would use onClickCapture to handle the click event in the capture phase.

Added support for this by adding -`Capture` keys to the event handling logic.

Diff best viewed ignoring whitespace: https://github.com/cleandart/react-dart/pull/114/files?w=1


greglittlefield-wf:
@jacehensley-wf

jacehensley-wf:
+1

PR #117, opened 2017-03-28
Fix memory leak caused by retained `internal` from first render

greglittlefield-wf:
## Ultimate Problem

That `ReactDartComponentInternal ` instance from the first render *should* be GC'd in subsequent renders as it's replaced with a new instance.

However, this wasn't happening since it was retained by the `Component._jsRedraw` closure, which hangs around for the lifetime of the component.

This resulted in the props from the initial render of any given component being retained until the component unmounts.

Big thanks to @bencampbell-wf for discovering the presence of a memory leak and creating a reduced test case!

## Solution
- Don't reference `internal` inside that closure

## Testing
Verify via the attached test case that this memory leak has been fixed: [leak_test_case.zip](https://github.com/cleandart/react-dart/files/877003/leak_test_case.zip)

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -162,11 +162,9 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
 
   /// Wrapper for [Component.getInitialState].
   void initComponent(ReactComponent jsThis, ReactDartComponentInternal internal, ComponentStatics componentStatics) => zone.run(() {
-    var redraw = () {
-      if (internal.isMounted) {
-        jsThis.setState(emptyJsMap);
-      }
-    };
+    void jsRedraw() {
+      jsThis.setState(emptyJsMap);
+    }
 
     Ref getRef = (name) {
       var ref = getProperty(jsThis.refs, name);
@@ -177,7 +175,7 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
     };
 
     Component component = componentStatics.componentFactory()
-        ..initComponentInternal(internal.props, redraw, getRef, jsThis);
+        ..initComponentInternal(internal.props, jsRedraw, getRef, jsThis);
 
     internal.component = component;
     internal.isMounted = false;


lib/react_client.dart
```diff
@@ -162,11 +162,9 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
 
   /// Wrapper for [Component.getInitialState].
   void initComponent(ReactComponent jsThis, ReactDartComponentInternal internal, ComponentStatics componentStatics) => zone.run(() {
-    var redraw = () {
-      if (internal.isMounted) {
```
greglittlefield-wf: This check always evaluated to true after mounting since this `internal` was never updated, so this can be safely removed.

lib/react_client.dart
```diff
@@ -162,11 +162,9 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
...
-        jsThis.setState(emptyJsMap);
-      }
-    };
+    void jsRedraw() {
```
greglittlefield-wf: Used a local function instead of a function expression for better stack traces.

bencampbell-wf:
+1

trentgrover-wf:
+1

dustinlessard-wf:
+1

lib/react_client.dart
dustinlessard-wf: Can you give an example of the before and after stacktraces?

evanweible-wf:
+1

lib/react_client.dart
greglittlefield-wf: Yeah! Output from https://dartpad.dartlang.org/a995d5478ae22d1c042922a3a40ee75c:

Local function:
```
 at main_localFunction.call$0 (<anonymous>:1454:19)
```
Function variable:
```
 at main_closure.call$0 (<anonymous>:1460:19)
```

maxwellpeterson-wf:
+1

bencampbell-wf:
+10

Things Tested
-------------
- [x] Added `feb7f9a6c32ae3fef769f32adc584d12ec42cfc2` as react dependency in  minimal example
- [x] Confirmed `Bar` instance was in memory in initial render
- [x] Waited for subsequent re-render with updated props
- [x] Confirmed `Bar` instance was no longer in memory and was successfully GC'd

PR #118, opened 2017-05-25
Fix strong mode warning

greglittlefield-wf:
Fix for strong mode warning in js_interop_helpers.

Testing: pull down and verify there are no strong mode warnings.

### DIFF
--- a/lib/react_client/js_interop_helpers.dart
+++ b/lib/react_client/js_interop_helpers.dart
@@ -90,13 +90,16 @@ EmptyObject jsify(Map map) {
   var jsMap = new EmptyObject();
 
   map.forEach((key, value) {
+    var jsValue;
     if (value is Map) {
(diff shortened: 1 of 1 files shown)


robbecker-wf:
I verified that the strong mode error in master is no longer present in this PR.

The below error is gone:
```
file: 'file:///Users/robbecker/code/react-dart/lib/react_client/js_interop_helpers.dart'
severity: 'Error'
message: 'Couldn't infer type parameter 'F'.

Tried to infer 'dynamic' for 'F' which doesn't work:
  Type parameter 'F' declared to extend 'Function'.
The type 'dynamic' was inferred from:
  Parameter 'f' declared as     'F'
                but argument is 'dynamic'.

Consider passing explicit type argument(s) to the generic.

'
at: '96,15'
source: 'dart'
```

PR #120, opened 2017-06-15
UIP-2397 Add memory related lints / misc. strong mode fixes

aaronlademann-wf:
* Added memory related lints
* Addressed strong mode lints introduced in Dart SDK `1.24.0`
* Misc. typo fixes / doc comment improvements

---

FYA @greglittlefield-wf @jacehensley-wf @clairesarsam-wf @trentgrover-wf


lib/react_client/react_interop.dart
```diff
@@ -113,7 +113,7 @@ class ReactElementStore {
 @JS()
 @anonymous
 class ReactElement {
-  external ReactElementStore get _store;
+  external ReactElementStore get _store; // ignore: unused_element
```
jacehensley-wf: If this is private and unused do we even need to keep it? If so we should probably add a code comment explaining why it's still around.

aaronlademann-wf: I don't really know what it is @jacehensley-wf ... let's follow up on that at a later date.

jacehensley-wf:
+1

PR #124, opened 2017-06-30
Add null-aware operator when calling componentWillUnmount

jacehensley-wf:
`internal.component` can be null within `handleComponentWillUnmount`. Use a the null aware operator so it doesn't throw

@aaronlademann-wf @greglittlefield-wf @clairesarsam-wf @joelleibow-wf

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -312,7 +312,7 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
   /// Wrapper for [Component.componentWillUnmount].
   void handleComponentWillUnmount(ReactDartComponentInternal internal) => zone.run(() {
     internal.isMounted = false;
-    internal.component.componentWillUnmount();
+    internal.component?.componentWillUnmount();
     // Clear the reference to the component. It won't be used by it anymore,
     // and we don't want it leaking.
     internal.component = null;


aaronlademann-wf:
introduced by https://github.com/cleandart/react-dart/pull/123

aaronlademann-wf:
+1

greglittlefield-wf:
@jacehensley-wf Can you elaborate as to when this happens, and why it being null at that point is acceptable?

At first glance, it seems like this is caused by a bug in the component being nulled when it shouldn't have been.

jacehensley-wf:
@greglittlefield-wf Yeah I'm not entirely sure. I think it's happening when the prev component is nulled and then getting unmounted.

greglittlefield-wf:
Yeah, I'd prefer it if we knew the root cause of this issue before putting this food in place.

Yeah, I'll pay those up when I get a chance.

You my also want to try just removing this line, and seeing if that fixes it: github.com/cleandart/react-dart/pull/123/files#diff-12ac054bb8b76f35c1035afef1d94396R308

greglittlefield-wf:
@jacehensley-wf Here are those changes: https://github.com/cleandart/react-dart/compare/internal_instance_refactor

After thinking about it more, those changes shouldn't exhibit the issue you're seeing, since the component instance is never nulled out.

jacehensley-wf:
@greglittlefield-wf What do you want to do about this then? Maybe just put this fix in place till we can release a breaking change with those changes in the branch?

greglittlefield-wf:
Like I said, I think it'd be a good idea to get a better understanding of this issue before putting this change in place.

If the component being null at that point is expected, and we do need a null-check, then I'm totally fine putting one in.

I just want to make sure that the problem described here isn't a symptom of some other underlying issue that should be fixed instead.

greglittlefield-wf:
Also, see my previous message; that line I mentioned is probably not necessary, and may be causing issues..

aaronlademann-wf:
@greglittlefield-wf @jacehensley-wf looks like the issue is present anywhere that `internal.component` is referenced...

```bash
VM290:1 Uncaught Unhandled exception:
NoSuchMethodError: The setter 'nextProps=' was called on null.
Receiver: null
Tried calling: nextProps=_LinkedHashMap len:9
#0      Object._noSuchMethod (dart:core-patch/object_patch.dart:43)
#1      Object.noSuchMethod (dart:core-patch/object_patch.dart:47)
#2      _dartInteropStatics.<anonymous closure>.handleComponentWillReceiveProps.<anonymous closure> (package:react/react_client.dart:264:9)
#3      _RootZone.run (dart:async/zone.dart:1366)
#4      _dartInteropStatics.<anonymous closure>.handleComponentWillReceiveProps (package:react/react_client.dart:261:126)
```

greglittlefield-wf:
That makes the line I referenced earlier even more suspect. I wonder if
that's getting cleared during a rerender, and not updated due to
`shouldComponentUpdate` or something.

jacehensley-wf:
@greglittlefield-wf Just tried removing this [line](https://github.com/cleandart/react-dart/pull/123/files#diff-12ac054bb8b76f35c1035afef1d94396R308) and am still seeing the error.

aaronlademann-wf:
@greglittlefield-wf we're going to revert #123 for now so that consumers aren't broken while we look into it more.

jacehensley-wf:
Gonna close this while we figure out a good solution

PR #130, opened 2017-09-18
Release 3.6.0: Allow retrieving the original converted event handlers

greglittlefield-wf:
## Ultimate Problem
When event handlers are added to DOM components, those handlers are converted to functions that proxy the JS events to the Dart event handlers.

Those converted functions are what end up on the `ReactElement` props so that they work when passed to the JS side of React.

However, this causes problems when chaining event handlers that exist on DOM `ReactElement`s, since you need to chain the original Dart function and not the converted JS function.

## Solution
Allow retrieving the original function from converted event handlers via a helper function backed by an Expando.

## Testing
Verify that tests pass in:
- the Dart VM: `pub run test -p content-shell`
- dart2js: `pub run test -p chrome`
- ddc: 
    - In one terminal: `pub serve test --web-compiler=dartdevc`
    - In another: `pub run test -p chrome --pub-serve=8080 -x ddcFailure`

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -448,6 +448,24 @@ _convertBoundValues(Map args) {
   }
 }
 
+/// A mapping from converted/wrapped JS handler functions (the result of [_convertEventHandlers])
+/// to the original Dart functions (the input of [_convertEventHandlers]).
+final Expando<Function> _originalEventHandlers = new Expando();
+
+/// Returns the original Dart handler function that, within [_convertEventHandlers],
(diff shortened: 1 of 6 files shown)


clairesarsam-wf:
+10

* Code looks good
* Tests pass in the VM, dart2js, and the DDC

aaronlademann-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10
- [x] Unit tests created/updated
- [x] All unit tests pass
- [x] Rosie ran/Rosie comment displays expected info
- [x] Dependency Scan Clean

PR #135, opened 2017-11-01
Context support

johnbland-wf:
## Feature

Support context.

## Solution

- Provided support for context without a breaking change by adding extra `*WithContext` methods. All of these methods replace the normal lifecycle methods so both are not needed.
  - `componentDidUpdateWithContext`
  - `componentWillUpdateWithContext`
  - `shouldComponentUpdateWithContext`
  - `componentWillReceivePropsWithContext`
- `shouldComponentUpdateWithContext` returns `null` by default so `shouldComponentUpdate` is still used. If `shouldComponentUpdateWithContext` returns a valid `bool`, `shouldComponentUpdate` will not get checked
- Updated tests to verify all of the lifecycle methods are called

## Testing

- Verify unit tests pass
- Verify context example makes sense and works as expected

### DIFF
--- a/js_src/dart_helpers.js
+++ b/js_src/dart_helpers.js
@@ -49,7 +49,7 @@ function _createReactDartComponentClassConfig(dartInteropStatics, componentStati
       config.childContextTypes[childContextKeys[i]] = React.PropTypes.object;
     }
 
-    // Only declare this when `hasChildContext` is true to avoid unnecessarily
+    // Only declare this when `childContextKeys` is non-empty to avoid unnecessarily
     // creating interop context objects for components that won't use it.
     config.getChildContext = function() {
       return dartInteropStatics.handleGetChildContext(this.dartComponent);
--- a/js_src/dart_helpers.js
+++ b/js_src/dart_helpers.js
@@ -6,7 +6,32 @@ function _getProperty(obj, key) { return obj[key]; }
 function _setProperty(obj, key, value) { return obj[key] = value; }
 
 function _createReactDartComponentClassConfig(dartInteropStatics, componentStatics) {
+  var childContextTypes;
+  var contextTypes;
+
+  var childContextKeys = jsConfig && jsConfig.childContextKeys;
+  var contextKeys = jsConfig && jsConfig.contextKeys;
+
+  if (childContextKeys && childContextKeys.length !== 0) {
+    childContextTypes = {};
+    for (var i = 0; i < childContextKeys.length; i++) {
+      childContextTypes[childContextKeys[i]] = React.PropTypes.any;
+    }
+  }
+
+  if (contextKeys && contextKeys.length !== 0) {
+    contextTypes = {};
+    for (var i = 0; i < contextKeys.length; i++) {
+      contextTypes[contextKeys[i]] = React.PropTypes.any;
+    }
+  }
+
   return {
+    getChildContext: function() {
+      return dartInteropStatics.getChildContext(this.props.internal);
+    },
+    childContextTypes: childContextTypes,
+    contextTypes: contextTypes,
     getInitialState: function() {
       this.dartComponent = dartInteropStatics.initComponent(this, this.props.internal, componentStatics);
       return {};
@@ -17,17 +42,17 @@ function _createReactDartComponentClassConfig(dartInteropStatics, componentStati
     componentDidMount: function() {
       dartInteropStatics.handleComponentDidMount(this.dartComponent);
     },
-    componentWillReceiveProps: function(nextProps) {
-      dartInteropStatics.handleComponentWillReceiveProps(this.dartComponent, nextProps.internal);
+    componentWillReceiveProps: function(nextProps, nextContext) {
+      dartInteropStatics.handleComponentWillReceiveProps(this.dartComponent, nextProps.internal, nextContext);
     },
-    shouldComponentUpdate: function(nextProps, nextState) {
-      return dartInteropStatics.handleShouldComponentUpdate(this.dartComponent);
+    shouldComponentUpdate: function(nextProps, nextState, nextContext) {
+      return dartInteropStatics.handleShouldComponentUpdate(this.dartComponent, nextContext);
     },
-    componentWillUpdate: function(nextProps, nextState) {
-      dartInteropStatics.handleComponentWillUpdate(this.dartComponent);
+    componentWillUpdate: function(nextProps, nextState, nextContext) {
+      dartInteropStatics.handleComponentWillUpdate(this.dartComponent, nextContext);
     },
-    componentDidUpdate: function(prevProps, prevState) {
-      dartInteropStatics.handleComponentDidUpdate(this.dartComponent, prevProps.internal);
+    componentDidUpdate: function(prevProps, prevState, prevContext) {
+      dartInteropStatics.handleComponentDidUpdate(this.dartComponent, prevProps.internal, prevContext);
     },
     componentWillUnmount: function() {
       dartInteropStatics.handleComponentWillUnmount(this.dartComponent);
--- a/js_src/dart_helpers.js
+++ b/js_src/dart_helpers.js
@@ -26,8 +26,8 @@ function _createReactDartComponentClassConfig(dartInteropStatics, componentStati
     componentWillUpdate: function(nextProps, nextState, nextContext) {
       dartInteropStatics.handleComponentWillUpdate(this.dartComponent, nextContext);
     },
-    componentDidUpdate: function(prevProps, prevState, prevContext) {
-      dartInteropStatics.handleComponentDidUpdate(this.dartComponent, prevProps.internal, prevContext);
+    componentDidUpdate: function(prevProps, prevState) {
+      dartInteropStatics.handleComponentDidUpdate(this.dartComponent, prevProps.internal);
     },
     componentWillUnmount: function() {
       dartInteropStatics.handleComponentWillUnmount(this.dartComponent);
--- a/js_src/dart_helpers.js
+++ b/js_src/dart_helpers.js
@@ -33,7 +33,7 @@ function _createReactDartComponentClassConfig(dartInteropStatics, componentStati
     childContextTypes: childContextTypes,
     contextTypes: contextTypes,
     getInitialState: function() {
-      this.dartComponent = dartInteropStatics.initComponent(this, this.props.internal, componentStatics);
+      this.dartComponent = dartInteropStatics.initComponent(this, this.props.internal, this.context, componentStatics);
       return {};
     },
     componentWillMount: function() {
--- a/js_src/dart_helpers.js
+++ b/js_src/dart_helpers.js
@@ -52,7 +52,7 @@ function _createReactDartComponentClassConfig(dartInteropStatics, componentStati
     // Only declare this when `hasChildContext` is true to avoid unnecessarily
     // creating interop context objects for components that won't use it.
     config.getChildContext = function() {
-      return dartInteropStatics.getChildContext(this.dartComponent);
+      return dartInteropStatics.handleGetChildContext(this.dartComponent);
     };
   }
 
--- a/js_src/dart_helpers.js
+++ b/js_src/dart_helpers.js
@@ -5,33 +5,8 @@
 function _getProperty(obj, key) { return obj[key]; }
 function _setProperty(obj, key, value) { return obj[key] = value; }
 
-function _createReactDartComponentClassConfig(dartInteropStatics, componentStatics) {
-  var childContextTypes;
-  var contextTypes;
-
-  var childContextKeys = jsConfig && jsConfig.childContextKeys;
-  var contextKeys = jsConfig && jsConfig.contextKeys;
-
-  if (childContextKeys && childContextKeys.length !== 0) {
-    childContextTypes = {};
-    for (var i = 0; i < childContextKeys.length; i++) {
-      childContextTypes[childContextKeys[i]] = React.PropTypes.any;
-    }
-  }
-
-  if (contextKeys && contextKeys.length !== 0) {
-    contextTypes = {};
-    for (var i = 0; i < contextKeys.length; i++) {
-      contextTypes[contextKeys[i]] = React.PropTypes.any;
-    }
-  }
-
-  return {
-    getChildContext: function() {
-      return dartInteropStatics.getChildContext(this.props.internal);
-    },
-    childContextTypes: childContextTypes,
-    contextTypes: contextTypes,
+function _createReactDartComponentClassConfig(dartInteropStatics, componentStatics, jsConfig) {
+  var config = {
     getInitialState: function() {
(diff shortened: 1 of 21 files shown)


js_src/dart_helpers.js
```diff
@@ -36,6 +36,34 @@ function _createReactDartComponentClassConfig(dartInteropStatics, componentStati
...
+  var childContextKeys = jsConfig && jsConfig.childContextKeys;
+  var contextKeys = jsConfig && jsConfig.contextKeys;
+
+  if (childContextKeys && childContextKeys.length !== 0) {
+    config.childContextTypes = {};
+    for (var i = 0; i < childContextKeys.length; i++) {
+      config.childContextTypes[childContextKeys[i]] = React.PropTypes.object;
+    }
+
+    // Only declare this when `hasChildContext` is true to avoid unnecessarily
```
jeroencranendonk-wf: Referencing an old variable?

johnbland-wf: Possibly. I'll dig into it.

greglittlefield-wf: I think so, sorry about that. Should probably read `when childContextKeys is non-empty`

johnbland-wf: Fixed

greglittlefield-wf:
@greglittlefield-wf @aaronlademann-wf @jacehensley-wf @clairesarsam-wf

lib/react.dart
```diff
@@ -81,29 +89,53 @@ abstract class Component {
...
     this._jsThis = _jsThis;
+    _initContext(context);
     _initProps(props);
   }
 
+  /// Initializes context
+  _initContext(context) {
+    this.context = new Map.from(context ?? {});
+
+    /// Call `transferComponentContext` to get state also to `_prevContext`
```
greglittlefield-wf: #nit This should not be a doc comment

lib/react.dart
```diff
@@ -81,29 +89,53 @@ abstract class Component {
...
   _initProps(props) {
     this.props = new Map.from(props);
     this.nextProps = this.props;
   }
 
   initStateInternal() {
     this.state = new Map.from(getInitialState());
-    // Call `transferComponentState` to get state also to `_prevState`
+
+    /// Call `transferComponentState` to get state also to `_prevState`
```
greglittlefield-wf: #nit This should not be a doc comment

lib/react.dart
```diff
@@ -114,16 +146,31 @@ abstract class Component {
 
   /// Public getter for [_nextState].
```
greglittlefield-wf: s/`_nextState`/`_nextContext`/

lib/react_client.dart
```diff
@@ -225,9 +250,15 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
   /// 1. Update [Component.props] using the value stored to [Component.nextProps]
   ///    in `componentWillReceiveProps`.
   /// 2. Update [Component.state] by calling [Component.transferComponentState]
-  void _afterPropsChange(Component component) {
+  /// 3. Update [Component.context] by calling [Component.transferComponentContext]
+  void _afterPropsChange(Component component, InteropContextValue nextContext) {
     component.props = component.nextProps; // [1]
     component.transferComponentState();    // [2]
+    component.transferComponentContext();  // [3]
```
greglittlefield-wf: From this code, and from testing the example, context doesn't seem to ever get updated. 

`nextState` gets updated in `setState` and then transferred to `state` in `transferComponentContext`, but there doesn't seem to be an analagous updating of `nextContext` here.

I would recommend updating how context is managed to mirror `props` as opposed to `state`, since they both are updated externally.

This updating should also have test coverage (see later comment in tests).

lib/react.dart
```diff
@@ -204,37 +251,102 @@ abstract class Component {
...
+  /// This method is not called for the initial [render].
+  ///
+  /// Use this as an opportunity to operate on the [rootNode] (DOM) when the `Component` has been updated as a result
+  /// of the values of [prevProps] / [prevState] /[prevContext].
+  ///
+  /// __Note__: Choose either this method or [componentDidUpdate]. They are both called at the same time so using both
+  /// provides no added benefit.
+  ///
+  /// See: <https://facebook.github.io/react/docs/react-component.html#updating-componentdidupdate>
+  void componentDidUpdateWithContext(Map prevProps, Map prevState, Map prevContext) {}
```
greglittlefield-wf: Since React 16 doesn't include `prevContext` in `componentDidUpdate`, can we just omit this lifecycle method?

test/lifecycle_test.dart
```diff
@@ -255,17 +264,22 @@ void main() {
...
+        matchCall('componentWillUpdateWithContext',       args: [newPropsWithDefaults, newState, expectedContext], props: initialPropsWithDefaults, state: initialState),
         matchCall('render',                                                                    props: newPropsWithDefaults, state: newState),
         matchCall('componentDidUpdate',        args: [initialPropsWithDefaults, initialState], props: newPropsWithDefaults, state: newState),
+        matchCall('componentDidUpdateWithContext',        args: [initialPropsWithDefaults, initialState, expectedContext], props: newPropsWithDefaults, state: newState),
       ]));
     });
 
-    group('when shouldComponentUpdate returns false:', () {
+    void testShouldUpdates({bool shouldComponentUpdateWithContext, bool shouldComponentUpdate}) {
       test('recieves updated props with correct lifecycle calls and does not rerender', () {
```
greglittlefield-wf: There should be tests that context maps are updated as expected and with the correct values in the correct lifecycle methods.

An easy way to do that would be to parallel this props update test as well as the other "recieves updated props" test.

lib/react_client.dart
johnbland-wf: I believe that is a react-js issue: https://github.com/facebook/react/issues/2517.

greglittlefield-wf: I believe that issue is around how to handle context not getting updated when `shouldComponentUpdate` returns `false`. 

Otherwise, context should update.

johnbland-wf: I agree. I'm validating that. The initial code had this issue and I found several references of it not updating.

I'll ping back once I confirm it.

johnbland-wf: Built a react-js app and it does indeed update. Investigating it.

johnbland-wf: Ok...it is refactored to match the props integration.

johnbland-wf:
Ok, this now updates context as expected. It isn't yet unit tested, but the big hurdle is done. Gotta jet. I'll tighten this up tomorrow.

johnbland-wf:
Ok, @greglittlefield-wf...final review time.

That test found some issues w/ the wrong context being passed to lifecycle methods. Let me know how it sits.

lib/react.dart
```diff
@@ -139,6 +172,8 @@ abstract class Component {
   ///
   /// [A.k.a "forceUpdate"](https://facebook.github.io/react/docs/react-component.html#forceupdate)
   void redraw([callback()]) {
+    context = getChildContext();
```
jacehensley-wf: Why does this happen here but not in `setState`?

johnbland-wf: Hrmm...guess I mistook JS for Dart here w/ the `_jsRedraw` vs `redraw`.

Should it go in `replaceState` as well?

jacehensley-wf: Hmm, what's the purpose of doing this anyways? Doesn't this https://github.com/cleandart/react-dart/pull/135/files#diff-12ac054bb8b76f35c1035afef1d94396R257 update the context?

johnbland-wf: Great find. I removed it and all tests pass + examples work. See 723caa5.

jacehensley-wf:
+1

lib/react_client/react_interop.dart
```diff
@@ -178,8 +192,7 @@ class InteropProps {
   external factory InteropProps({ReactDartComponentInternal internal, String key, dynamic ref});
 }
 
-/// Internal react-dart information used to proxy React JS lifecycle to Dart
-/// [Component] instances.
+/// A Dart object that stores .
```
greglittlefield-wf: Looks like this was leftover from my one of my commits, sorry 😬. Can you revert this doc comment change, please?

test/ReactSetStateTestComponent.js
```diff
@@ -27,15 +27,28 @@ var ReactSetStateTestComponent = React.createClass({
     this.recordLifecyleCall("componentWillReceiveProps");
   },
 
+  componentWillReceivePropsWithContext: function(_) {
```
greglittlefield-wf: These `-WithContext` methods on the JS component won't ever get called, right? Why do they exist?

test/lifecycle_test.dart
```diff
@@ -371,11 +476,28 @@ void main() {
...
       });
+    }
+
+    group('when shouldComponentUpdate returns false:', () {
+      testShouldUpdates(shouldComponentUpdateWithContext: null, shouldComponentUpdate: false);
+    });
+
+    group('when shouldComponentUpdateWithContext returns false:', () {
+      testShouldUpdates(shouldComponentUpdateWithContext: false, shouldComponentUpdate: false);
     });
```
greglittlefield-wf: Should we add one more test case that `shouldComponentUpdate` is not called when `shouldComponentUpdateWithContext` returns `true`? Or is that overkill?

test/lifecycle_test.dart
```diff
@@ -139,7 +140,78 @@ void main() {
       ]));
     });
```
greglittlefield-wf: There should be a lifecycle tests that asserts that `getChildContext` is not called when `childContextKeys` is empty.

test/lifecycle_test.dart
```diff
@@ -139,7 +140,78 @@ void main() {
...
+        matchCall('componentWillReceivePropsWithContext', args: [newPropsWithDefaults, expectedContext],                props: initialPropsWithDefaults, context: initialContext),
+        matchCall('shouldComponentUpdateWithContext',     args: [newPropsWithDefaults, expectedState, expectedContext], props: initialPropsWithDefaults, context: initialContext),
+        matchCall('componentWillUpdate',                  args: [newPropsWithDefaults, expectedState],                  props: initialPropsWithDefaults, context: initialContext),
+        matchCall('componentWillUpdateWithContext',       args: [newPropsWithDefaults, expectedState, expectedContext], props: initialPropsWithDefaults, context: initialContext),
+        matchCall('render',                                                                                             props: newPropsWithDefaults, context: expectedContext),
+        matchCall('componentDidUpdate',                   args: [initialPropsWithDefaults, expectedState],              props: newPropsWithDefaults, context: expectedContext),
+      ]));
+    });
```
greglittlefield-wf: Awesome test! Could we add one more that verifies that only the keys present in `childContextKeys` are passed down, though?

lib/react.dart
```diff
@@ -112,14 +139,20 @@ abstract class Component {
   /// __DO NOT set__ from anywhere outside react-dart lifecycle internals.
   Map prevState;
 
+  /// Public getter for [_nextContext].
+  ///
+  /// If `null`, then [_nextContext] is equal to [context] - which is the value that will be returned.
+  // Map get nextContext => _nextContext == null ? context : _nextContext;
```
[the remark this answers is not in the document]
johnbland-wf: Nope, that was dead code from a previous build. Removing...

test/ReactSetStateTestComponent.js
johnbland-wf: I believe it was only testing so `recordLifecyleCall` would record the call and could be validated.

lib/react_client.dart
```diff
@@ -224,10 +248,14 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
 
   /// 1. Update [Component.props] using the value stored to [Component.nextProps]
   ///    in `componentWillReceiveProps`.
-  /// 2. Update [Component.state] by calling [Component.transferComponentState]
-  void _afterPropsChange(Component component) {
-    component.props = component.nextProps; // [1]
-    component.transferComponentState();    // [2]
+  /// 2. Update [Component.context] using the value stored to [Component.nextContext]
```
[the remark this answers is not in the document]
johnbland-wf: Ah, man...too many twists and turns in here. :-D Great find. I found where the issue was w/ `nextContext` being outdated.

test/ReactSetStateTestComponent.js
greglittlefield-wf: Not sure I'm following. `recordLifecycleCall` won't be called since nothing is calling these methods on the JS component, right?

test/lifecycle_test.dart
johnbland-wf: Setting an extra key to pass but not having the value in the child should be enough. That's incoming.

test/lifecycle_test.dart
johnbland-wf: Tests incoming. Good call here.

test/lifecycle_test.dart
johnbland-wf: That's pretty much what this is testing:
```
        List calls = [
          matchCall('componentWillReceiveProps', args: [newPropsWithDefaults],                props: initialPropsWithDefaults),
          matchCall('componentWillReceivePropsWithContext', args: [newPropsWithDefaults, expectedContext],                props: initialPropsWithDefaults),
          matchCall('shouldComponentUpdateWithContext',     args: [newPropsWithDefaults, expectedState, expectedContext], props: initialPropsWithDefaults),
        ];

        if (shouldComponentUpdateWithContext == null) {
          calls.add(
            matchCall('shouldComponentUpdate', args: [newPropsWithDefaults, expectedState], props: initialPropsWithDefaults),
          );
        }
```

When `shouldComponentUpdateWithContext` is `null` we validate the method is called. If it was called when `shouldComponentUpdateWithContext` had a value, this test would fail.

johnbland-wf:
Ok, @greglittlefield-wf, I think it is ready.

Big props to you and @jacehensley-wf for catching those issues.

test/lifecycle_test.dart
johnbland-wf: See 
282914b

test/lifecycle_test.dart
johnbland-wf: See 282914b.

lib/react_client.dart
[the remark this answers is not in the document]
johnbland-wf: See f0d8b3a.

test/lifecycle_test.dart
greglittlefield-wf: Right, I just meant testing `true` as a value in addition to `false`.

greglittlefield-wf:
+1 looks great!

Going to pull this down and test, and also pull into some other repos for testing.

greglittlefield-wf:
- Context works as expected in example
- All tests pass in content-shell, dart2js, and DDC
- All tests pass when pulled into OverReact and WSD

+10

jacehensley-wf:
+1 refresh

PR #138, opened 2018-02-26
Expose “dev mode” flag

aaronlademann-wf:
Creates a top-level getter that can be used to conditionally do stuff based on whether the "dev" build of ReactJS is being used on the page.

@greglittlefield-wf @jacehensley-wf @clairesarsam-wf @trentgrover-wf


js_src/dart_env_dev.js
```diff
@@ -0,0 +1 @@
+__isDevelopment = true;
```
greglittlefield-wf: Can we make this `React.__isDevelopment` instead of a global?

clairesarsam-wf:
+10

* All tests pass locally in content-shell and chrome and DDC

greglittlefield-wf:
Merging this and releasing 4.3.0

PR #141, opened 2018-05-07
Prepare for Dart SDK 2 Support / Misc. Housekeeping

aaronlademann-wf:
### Problem(s):

1. The next major version of `react` will support version 2 of the Dart SDK.  In order to do that, we have to make a few breaking changes.
2. Some general housekeeping / tech debt needs to be addressed.

### Solution(s):

1. __Deprecate `ReactComponentFactory`__, and all public test utilities that use it as a type signature.
    1. __`ReactComponentFactory` will be removed in the `react 5.0.0` release.__
    2. __Use `ReactComponentFactoryProxy` instead, except for the return value of `registerComponent`.__
        1. The return value of `registerComponent` should be changed to `final` or `var`.
    3. This deprecation stems from a breaking change in the upcoming Dart SDK 2.x release:
        > We (the Dart team) are preparing to land a breaking change for Dart 2 related to callable classes. Below is a summary of the changes and how they might impact you. Feel free to follow up here or with me directly with questions.
        >
        > In Dart 2, we will continue to allow objects with call methods to be called, and to be assigned to variables of function type. 
        >
        > However, the assignment of an object with a call method to a function type will now be treated as an implicit closurization of the .call method of the object.
2. Address housekeeping / tech debt items:
    1. https://github.com/cleandart/react-dart/pull/141/commits/0e84b8e043622106596400a113c5c68f4e28b606 Exclude the `.dart_tool` directory _(created by Dart 2's new `pub get` impl)_ from git.
    2. https://github.com/cleandart/react-dart/pull/141/commits/8fb414f223c2176198eaf0fb53316f500b617a47 Update authors / home page in `pubspec.yaml`
    3. https://github.com/cleandart/react-dart/pull/141/commits/7136328ab4f894e9be375d4a018d80ebb846b536 Address some lints that will show up when using the Dart 2 analyzer
    4. https://github.com/cleandart/react-dart/pull/141/commits/7e48ce6a951313b2fd9174cc3232fdea1fb586f6 Improve function / return signatures for `setState` callbacks
    5. https://github.com/cleandart/react-dart/pull/141/commits/d8b269e0ad939d0a2ef284a7f47b4bd4f952ff14 Improve instructions for running library tests
    6. https://github.com/cleandart/react-dart/pull/141/commits/d018a3b2119b46929abe0d552c2c3ef07422aebf Get all examples running in the DDC
        1. Also added a landing page for the examples, added a link to it from the README, and added some basic styling.

### Testing Instructions:

Follow the [new testing instructions](https://github.com/aaronlademann-wf/react-dart/blob/d8b269e0ad939d0a2ef284a7f47b4bd4f952ff14/README.md#running-tests). Verify that all tests pass.

---

__FYA:__ @greglittlefield-wf @clairesarsam-wf


lib/src/typedefs.dart
```diff
@@ -1,7 +1,19 @@
...
 /// * `react.Component` if it is a Dart component.
 /// * `Element` _(DOM node)_ if it is a React DOM component.
 typedef dynamic Ref(String ref);
+
+/// Typedef of a transactional [Component.setState] callback.
+///
+/// See: <https://facebook.github.io/react/docs/react-component.html#setstate>
+typedef Map TransactionalSetStateCallback(
+    Map prevState,
+    UnmodifiableMapView props,
```
greglittlefield-wf: I don't think this should be updated from `Map`, since it limits us in the future from passing in a Map that isn't an `UnmodifiableMapView` (e.g., one that is non-modifiable by some other mechanism that doesn't implement `UnmodifiableMapView`).

lib/react_client.dart
```diff
@@ -263,7 +277,7 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
   }
 
   void _callSetStateCallbacks(Component component) {
-    component.setStateCallbacks.forEach((callback()) { callback(); });
+    component.setStateCallbacks.forEach((callback) { callback(); });
```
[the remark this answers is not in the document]
aaronlademann-wf: Fixed via https://github.com/cleandart/react-dart/pull/141/commits/be9727139c7fa7bcd0a8fd7c8ebf20568279d049

lib/react.dart
```diff
@@ -177,16 +177,21 @@ abstract class Component {
...
-      throw new ArgumentError('setState expects its first parameter to either be a Map or a Function that accepts two parameters.');
+      throw new ArgumentError('setState expects its first parameter to either be a Map or a `TransactionalSetStateCallback`.');
     }
 
-    if (callback != null) _setStateCallbacks.add(callback);
+    if (callback != null) {
+      // Store locally to prevent concurrent modification during iteration
+      var localSetStateCallbacks = new List<SetStateCallback>.from(_setStateCallbacks);
+      localSetStateCallbacks.add(callback);
+      _setStateCallbacks = localSetStateCallbacks;
```
greglittlefield-wf: Looks like this is necessary for `replaceState` as well. You could also prevent the concurrent modification when the callbacks are iterated, which would require no changes to either of these:
```dart
  void _callSetStateCallbacks(Component component) {
    var callbacks = component.setStateCallbacks.toList();
    component.setStateCallbacks.clear();
    callbacks.forEach((callback()) { callback(); });
  }
```

test/lifecycle_test.dart
```diff
@@ -309,6 +309,50 @@ void main() {
...
+    });
+
+    test('properly handles a call to setState within another setState callback', () {
+      _LifecycleTest component = getDartComponent(render(LifecycleTest({})));
+
+      expect(() {
+        component.setState({'foo': 'bar'}, () {
+          component.lifecycleCalls.clear();
+
+          component.setState({'baz': 'foobar'}, () {});
```
greglittlefield-wf: I know this test was mentioned earlier, but now I'm a little worried that this isn't fully testing what could go wrong with the state transition and the callbacks... (for example, a callback added inside a setState callback getting called before that transition occurs)

Something like this would provide more coverage around the inner callback being called, and the state of `state` in response to each call:
```dart
var expectedState1 = {}..addAll(initialState)..addAll({'foo', 'bar'};
var expectedState2 = {}..addAll(expectedState2)..addAll({'baz': 'foobar'});

component.setState({'foo': 'bar'}, expectAsync0(() {
  expect(component.state, expectedState1);
  component.setState({'baz': 'foobar'}, expectAsync0(() {
    expect(component.state, expectedState2);
  }));
}));

expect(component.lifecycleCalls, containsAllInOrder([
  matchCall('componentWillUpdate', args: [expectedProps, expectedState1]),
  matchCall('componentWillUpdate', args: [expectedProps, expectedState2]),
]));
```
Thoughts?

clairesarsam-wf:
QA +1

Merging.

lib/react_test_utils.dart
```diff
@@ -190,18 +219,37 @@ external bool isElement(dynamic object);
 @JS('React.addons.TestUtils.isElementOfType')
 external bool _isElementOfType(dynamic element, dynamic componentClass);
 
-/// Returns true if [element] is a ReactElement whose type is of a
-/// React componentClass.
+/// __Deprecated. Will be removed in the `5.0.0` release.__ Use [isElementOfTypeV2] instead.
+///
+/// Returns `true` if [element] is a [ReactElement]
+/// that is of the `type` of the provided [componentFactory].
+@Deprecated('4.4.0')
```
greglittlefield-wf: Just saw this... These should be the version they're to be removed in, not the version they were deprecated in.

PR #142, opened 2018-05-08
[5.0.0-dev - dart2] AF-222 Fix Dart 2 Callable Classes Issues

aaronlademann-wf:
> __NOTE: THIS IS NOT A CLEAN DIFF.__ 
>
> It will not be clean until #141 merges, and the `5.0.0-dev` branch is updated from `master`.
>
> [__VIEW A CLEAN DIFF HERE__](https://github.com/aaronlademann-wf/react-dart/pull/1)

### Problem(s):
1. In the Dart 2 SDK, the assignment of an object with a call method to a function type will now be treated as an implicit closurization of the `.call` method of the object.  This is not compatible with how `ReactComponentFactory` works.
    > See: [Deprecation of `ReactComponentFactory`](https://github.com/cleandart/react-dart/pull/141) _(slated for the `4.4.0` release)_
2. Misc. housekeeping items need to be addressed to get the `5.0.0-dev` branch ready to support further Dart 2 SDK testing / `5.0.0-dev.<n>` releases as the Dart team moves closer and closer to a production release of the Dart 2 SDK.

### Solution(s):

1. https://github.com/cleandart/react-dart/commit/7350fef17febb2818216453f064d73c208cbc184 __Remove `ReactComponentFactory`__
    1. Change the type signature of `registerComponent` to `ComponentRegistrar`, which returns a new instance of `ReactComponentFactoryProxy`.
    2. _NOTE:_ the existing `ReactComponentFactoryProxy` class and `ComponentFactory` typedef were moved to `react.dart`.  They are still exported from `react_client.dart`.
1. Address housekeeping items
    1. Update Dart SDK dependency range to `>=2.0.0-dev.0 <=2.0.0-dev.49.0`

### Testing Instructions:

1. Switch your Dart SDK to `2.0.0-dev.49.0`
2. Verify that there are no analysis errors / warnings
3. Serve the examples using the DDC:
    ```bash
    pub serve example --web-compiler=dartdevc
    ```
4. Load `http://localhost:8080` and navigate to some examples _(the Geocodes app exercises a good portion of the APIs)_
5. Verify that the examples work as expected without producing run time errors

### Known Issues:

There are some issues present that, upon some investigation, appear to be unrelated to these changes:

1. There are numerous test failures centered around type casting.  I am unsure if this is a `test` package issue, or if we need to do away with some of our "shared" test suites.
2. A lot of the test errors seems to be centered around the conversion of synthetic event handlers.

---

__FYA:__ @greglittlefield-wf @clairesarsam-wf 

__FYI:__ @dustinlessard-wf @sebastianmalysa-wf @robbecker-wf


lib/react.dart
```diff
@@ -359,6 +362,35 @@ abstract class Component {
   dynamic render();
 }
 
+/// Creates ReactJS [ReactElement] instances.
```
greglittlefield-wf: `ReactElement` isn't valid here, since those are only used for non-server runtimes.

lib/react_client.dart
```diff
@@ -429,9 +400,10 @@ class ReactDomComponentFactoryProxy extends ReactComponentFactoryProxy {
     var children = _convertArgsToChildren(childrenArgs);
     children = listifyChildren(children);
 
-    convertProps(props);
+    var convertibleProps = {}..addAll(props);
```
greglittlefield-wf: This could use a comment: 
```dart
// We can't mutate the original since we can't be certain that the value of the
// the converted event handler will be compatible with the Map's type parameters.
```

clairesarsam-wf:
+10

- [x] No analysis errors
- [x] Smoke testing of the geocodes example app produces no errors

PR #146, opened 2018-07-12
AF-364: Support null return from render()

clairesarsam-wf:
# Ultimate problem

See <https://github.com/dart-lang/sdk/issues/27485>

# Solution

Normalize 'undefined' to 'null' in the js layer of interop.

# How to +10/QA

* Load the new test example (<http://localhost:8080/test/false_and_null_test.html>) in chrome and in dartium. It should execute without throwing exceptions.

---

Please review @greglittlefield-wf @aaronlademann-wf

### DIFF
--- a/example/test/false_and_null_test.dart
+++ b/example/test/false_and_null_test.dart
@@ -6,6 +6,10 @@ import 'package:react/react_client.dart';
 
 class _Component extends react.Component {
   render() {
+    if (props['hardCodedNullReturn'] == true) {
+      return null;
+    }
+
     return props['returnValue'];
   }
 }
@@ -16,10 +20,13 @@ void main() {
   setClientConfiguration();
 
   var content = react.div({}, [
-    react.p({}, 'Testing a return value of "null"...'),
+    react.p({}, 'Testing a dynamic return value of "null"...'),
     component({'returnValue': null, 'key': 0}),
+    react.p({}, 'Testing a hard-coded return value of "null"...'),
+    component({'hardCodedNullReturn': true, 'key': 1}),
     react.p({}, 'Testing a return value of "false"...'),
-    component({'returnValue': false, 'key': 1}),
+    component({'returnValue': false, 'key': 2}),
   ]);
+
   react_dom.render(content, querySelector('#content'));
 }
--- a/example/test/false_and_null_test.dart
+++ b/example/test/false_and_null_test.dart
@@ -1,0 +1,25 @@
+import 'dart:html';
+
+import 'package:react/react.dart' as react;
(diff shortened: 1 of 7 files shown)


greglittlefield-wf:
This fixes issue shown here: https://github.com/dart-lang/sdk/issues/27485

example/test/false_and_null_test.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:html';
+
+import 'package:react/react.dart' as react;
+import 'package:react/react_dom.dart' as react_dom;
+import 'package:react/react_client.dart';
+
+class _Component extends react.Component {
+  render() {
+    return props['returnValue'];
```
greglittlefield-wf: FYI this doesn't hit the case shown here: https://github.com/dart-lang/sdk/issues/27485

The issue is that Dart's `return null;` sometimes gets compiled to `return;` in dart2js. Having a dynamic return value won't result in the same output:

You'd need something like the following to fully exercise this:
```diff
diff --git a/example/test/false_and_null_test.dart b/example/test/false_and_null_test.dart
index d6b58e4..d8bec3f 100644
--- a/example/test/false_and_null_test.dart
+++ b/example/test/false_and_null_test.dart
@@ -6,6 +6,10 @@ import 'package:react/react_client.dart';
 
 class _Component extends react.Component {
   render() {
+    if (props['hardCodedNullReturn'] == true) {
+      return null;
+    }
+
     return props['returnValue'];
   }
 }
@@ -16,10 +20,12 @@ void main() {
   setClientConfiguration();
 
   var content = react.div({}, [
-    react.p({}, 'Testing a return value of "null"...'),
+    react.p({}, 'Testing a dynamic return value of "null"...'),
     component({'returnValue': null, 'key': 0}),
+    react.p({}, 'Testing a hard-coded return value of "null"...'),
+    component({'hardCodedNullReturn': true, 'key': 2}),
     react.p({}, 'Testing a return value of "false"...'),
-    component({'returnValue': false, 'key': 1}),
+    component({'returnValue': false, 'key': 3}),
   ]);
   react_dom.render(content, querySelector('#content'));
 }
```

dart2js:
```js
render$0: function(_) {
  if (J.$eq$(J.$index$asx(this._props, "hardCodedNullReturn"), true))
    return;
  return J.$index$asx(this._props, "returnValue");
}
```

greglittlefield-wf:
+10 on the fix, happy to have this in!

Not sure if we want to update the test case before merging, though

clairesarsam-wf:
@greglittlefield-wf Fixed!

greglittlefield-wf:
+10

Merging

PR #153, opened 2019-02-06
CPLAT-3968 CPLAT-3969 Add unconvertJsProps() util to react_client.dart

evanweible-wf:
## Description

The `getJsProps()` method in over_react makes more sense in this library, as it is intended to be used only with JS components and not Dart components. over_react has its own `getProps()` that should be sufficient (and uses `getJsProps()` internally).

To avoid naming conflicts and ambiguous imports in over_react, I've added the same functionality under a different name: `unconvertJsProps()`. This should also help clarify its intended usage in the future when some additional JS map work is taken on.

Additionally, logic was added to handle the unconverting of the JS event handlers (they are converted to their synthetic counterparts when added), which is necessary for Dart 2 compatibility.

## Testing
- [ ] CI passes
- [x] Verify this branch with over_react (on dart 1 and dart 2) as well as any other internal consumer large enough to instill adequate confidence

## Code Review
@greglittlefield-wf @aaronlademann-wf @kealjones-wk @corwinsheahan-wf

PR #154, opened 2019-02-07
Revert "Remove binding for nonexistent SyntheticDataTransfer class"

evanweible-wf:
Reverts cleandart/react-dart#151 - it results in some typing errors like so:

```
Unhandled exception:
type 'JSObjectImpl' is not a subtype of type 'DataTransfer' of 'dt' where
JSObjectImpl is from dart:js
DataTransfer is from dart:html
#0      syntheticDataTransferFactory (package:react/react_client.dart:746:65)
#1      syntheticMouseEventFactory (package:react/react_client.dart:784:30)
#2      _convertEventHandlers.<anonymous closure>.<anonymous closure>.<anonymous closure> (package:react/react_client.dart:561:35)
```

This commit had not yet been released, so for the time being (in order to get #153 through) we are going to revert. We can circle back to fix this by finding a way to construct a real DataTransfer object using JS.

@greglittlefield-wf @aaronlademann-wf @kealjones-wk @corwinsheahan-wf

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -712,7 +712,7 @@ SyntheticFormEvent syntheticFormEventFactory(events.SyntheticFormEvent e) {
 }
 
 /// Wrapper for [SyntheticDataTransfer].
-SyntheticDataTransfer syntheticDataTransferFactory(DataTransfer dt) {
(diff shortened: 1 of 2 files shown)


greglittlefield-wf:
+1

PR #158, opened 2019-02-22
testEvent logic could create false positives

kealjones-wk:
This fixes faulty logic in `testEvent` Tests. Previously the `eventName` supplied to the test was used in `eventData` and would create a false positive because the event name supplied was, in essence, compared to itself, not the actual `eventData`. If the event that was supplied was previously setup to call `onEvent`, would show whatever `eventName` that was provided to the test as `eventData`. Meaning, that the `eventName` comparison would match even if the actual even that fired it was not that event name. This fixes that, using the actual `event.type` as the comparison to the supplied `eventName`.

### Testing
- [ ] CI Passes

### DIFF
--- a/test/react_test_utils_test.dart
+++ b/test/react_test_utils_test.dart
@@ -74,12 +74,12 @@ void main() {
 
       test('with React instance as arg', () {
         event(findRenderedDOMComponentWithTag(component, 'div'), eventData);
-        expect(domNode.text, contains('$eventName $fakeTimeStamp'));
+        expect(domNode.text, equals('$eventName $fakeTimeStamp'));
       });
 
       test('with DOM Element as arg', () {
         event(domNode, eventData);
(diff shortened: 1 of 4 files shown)


test/react_test_utils_test.dart
```diff
@@ -59,27 +59,17 @@ void main() {
...
-    void testEvent(
-        void event(dynamic instanceOrNode, Map eventData), String eventName) {
-      Map eventData;
-      int fakeTimeStamp;
-
-      setUp(() {
-        fakeTimeStamp = eventName.hashCode;
-        eventData = {
-          'type': eventName,
-          'timeStamp': fakeTimeStamp,
```
greglittlefield-wf: Can we keep `eventData` and `timeStamp` in? I believe they're needed to test coverage of the eventData arg.

PR #159, opened 2019-02-26
CPLAT-4447: Type style map as Map<String, dynamic>

corwinsheahan-wf:
The typing for the unconverted js map for `style` should be `Map<String, dynamic>`.


greglittlefield-wf:
+10

PR #160, opened 2019-02-28
CPLAT-4303: React 16 New Context

kealjones-wk:
We may delay this PR until we start work on `ComponentV2` and then implement it in `ComponentV2` and put Legacy Context back in `Component` but this pr is ready for an implementation CR with this in mind. 

Merge conflicts are to be expected in the `lib` folders .js and .map files these are generated from the `js_src` files and running `npm run build`. They will be resolved with a merge of the `js_src` folders and a new build on the wip branch.

## Description:
React 16 implements a better New Context API.

## Changes:
- Added `createContext` to react_client.dart
- Added `contextType` to `react.Component2`
- Added `ReactDartContext` that utilize interop `ReactContext` for use in ReactJS
- Added `ReactJsComponentFactoryProxy` that was originally created by @greglittlefield-wf about 2 years ago, then added a `ReactJsContextComponentFactoryProxy` for use with react16 and context components.
    - Resurrects #112 to support #109
- Add tests for `ReactJsComponentFactoryProxy`
- Changes type checking test that was used for JsBackedMaps to a shared suite, which was then also used for context. 
- Updated tests that previously skipped thought to be deprecated lifecycle method `shouldComponentUpdateWithContext`
- Add tests for `calculateChangedBits` argument of context.
- Update `getSnapshotBeforeUpdate` tests to use new shared type tests
- Removes js_util deprecations and todos from https://github.com/cleandart/react-dart/pull/171

## Testing suggestions:
- [ ] CI Passes
- [ ] run `webdev serve example` and open http://localhost:8080
  - [ ] check http://localhost:8080/test/context_test.html and make sure it works as expected
  - [ ] check http://localhost:8080/js_components/index.html and make sure it works as expected


kealjones-wk:
FINALLLYYYY! darn you dart1... i was building in dart2! 
1_9smelzjfnrokxcc4qxi_gw

lib/react.dart
```diff
@@ -15,7 +16,23 @@ typedef ReactComponentFactoryProxy ComponentRegistrar(
 /// Top-level ReactJS [Component class](https://facebook.github.io/react/docs/react-component.html)
 /// which provides the [ReactJS Component API](https://facebook.github.io/react/docs/react-component.html#reference)
 abstract class Component {
-  Map _context;
```
aaronlademann-wf: @kealjones-wk @greglittlefield-wf isn't the plan was to implement all new features in the `ComponentV2` class?

kealjones-wk: The old context api exposed in react-dart wasn't all that great to being with on the reactjs side it was bad and the exposed api on the dart side didn't match reactjs's either. Also I think @greglittlefield-wf and I looked into it and found no one using it. I feel that the advantages of the new context API far outweigh the old and think that it will have much better real world application in both `Component` and `ComponentV2` that i dont belive keeping the old context in `Component` helps anyone.

aaronlademann-wf: We aren’t going to release any of this until 5.0.0 anyway - so there is no point putting it in a class that will be deprecated when that happens (`Component`).

kealjones-wk: Good point, i added a note to the PR description to that effect. I still would like an implementation review on this by anyone willing so that when we do decide to implement it in `ComponentV2` we can know this approach will work.

greglittlefield-wf: 👍 👍

lib/react_client.dart
```diff
@@ -844,6 +895,25 @@ dynamic _findDomNode(component) {
...
+class ReactDartContext {
+  dynamic _jsThis;
+  ReactDartContext(this.Provider, this.Consumer, this._jsThis);
+  ReactJsComponentFactoryProxy Provider;
+  ReactJsComponentFactoryProxy Consumer;
+  dynamic get jsThis => _jsThis;
+}
+
+ReactDartContext createContext(
+    [dynamic defaultValue, Function calculateChangedBits]) {
```
greglittlefield-wf: Same here, can we make this typed?

webpack.config.js
```diff
@@ -17,16 +17,11 @@ var babelPlugin = new webpack.DefinePlugin({
...
 var devPlugins = [
-  babelPlugin
+  babelPlugin,
 ];
 
 var prodPlugins = [
   babelPlugin,
-  new webpack.DefinePlugin({
-    optimization: {
-      minimizer: [new UglifyJsPlugin()]
```
greglittlefield-wf: Why was this removed? Don't we want minified code in prod?

test/factory/common_factory_tests.dart
```diff
@@ -21,7 +21,7 @@ void commonFactoryTests(ReactComponentFactoryProxy factory) {
 
   group('passes children to the component when specified as', () {
     dynamic getJsChildren(ReactElement instance) =>
-        getProperty(instance.props, 'children');
+        getProperty(instance.props as dynamic, 'children');
```
greglittlefield-wf: Why is this needed?

lib/react.dart
```diff
@@ -44,11 +61,15 @@ abstract class Component {
...
-  /// The React context map of this component, passed down from its ancestors' [getChildContext] value.
-  ///
-  /// Only keys declared in this component's [contextKeys] will be present.
-  Map get context => _context;
-  set context(Map value) => _context = value;
+  /// The React context value of this component, passed down from its Provider.
+  dynamic get context => _context;
+  set context(dynamic value) => _context = value;
+
+  /// The contextType property on a class can be assigned a Context object created by createContext().
```
greglittlefield-wf: This should have a note that it's only accessed once per component class (just like getDefaultProps).

lib/react.dart
```diff
@@ -44,11 +61,15 @@ abstract class Component {
...
+  /// This lets you consume the nearest current value of that Context type using this.context.
+  /// You can reference this in any of the lifecycle methods including the render function.
+  ReactDartContext get contextType => _contextType;
```
greglittlefield-wf: Since this is only accessed once per component class, there shouldn't be a need for a setter or this local variable.

webpack.config.js
kealjones-wk: So actually using webpacks `mode` config when marked as production does all this for us, turned out this was just redundant: https://webpack.js.org/concepts/mode/#mode-production

test/factory/common_factory_tests.dart
kealjones-wk: dart1

lib/react_client/react_interop.dart
```diff
@@ -19,12 +19,12 @@ typedef ReactElement ReactJsComponentFactory(props, children);
 @JS()
 abstract class React {
   external static String get version;
+  external static ReactContext createContext(
+      [dynamic defaultValue, Function calculateChangedBits]);
```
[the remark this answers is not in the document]
kealjones-wk: something like: 
`typedef int calculateChangedBits(dynamic currentValue, dynamic nextValue);`?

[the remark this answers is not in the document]
aaronlademann-wf: Since Component2 will be Dart 2 only, we should be able to [use the new inline function types instead of typedefs!](https://www.dartlang.org/guides/language/effective-dart/design#prefer-inline-function-types-over-typedefs)

```dart
external static ReactContext createContext([
    dynamic defaultValue, 
    int Function(dynamic, dynamic) calculateChangedBits,
]);
```

Can `external`s be parameterized? If so, doing something like this could make the interface safer:

```dart
external static ReactContext createContext<T>([
    T defaultValue,
    int Function(T, T) calculateChangedBits,
]);
```

lib/react_client.dart
aaronlademann-wf:
```dart
ReactDartContext createContext<T>([
    T defaultValue,
    int Function(T, T) calculateChangedBits,
]);
```

@greglittlefield-wf @kealjones-wk do you agree it makes sense to parameterize `createContext` to ensure that `defaultValue` and the type of the args passed to the `calculateChangedBits` function are the same?

test/factory/common_factory_tests.dart
aaronlademann-wf: @kealjones-wk since this will be going into the `5.0.0-wip` branch _(which will be Dart 2 only)_, this can be removed.

lib/react.dart
```diff
@@ -126,7 +147,7 @@ abstract class Component {
   /// Private reference to the value of [context] for the upcoming render cycle.
   ///
   /// Useful for ReactJS lifecycle methods [shouldComponentUpdateWithContext] and [componentWillUpdateWithContext].
-  Map nextContext;
+  dynamic nextContext;
```
aaronlademann-wf: For my own knowledge - what are the possible expected types?

lib/react_client.dart
```diff
@@ -27,6 +29,8 @@ export 'package:react/react_client/react_interop.dart'
 export 'package:react/react.dart'
     show ReactComponentFactoryProxy, ComponentFactory;
 
+@Deprecated('5.0.0')
```
```suggestion
@Deprecated('6.0.0')
```

The version within the annotation should be the version in which the deprecated thing will be removed.

lib/react_client.dart
```diff
@@ -192,11 +193,11 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
   Component initComponent(
           ReactComponent jsThis,
           ReactDartComponentInternal internal,
-          InteropContextValue context,
+          dynamic context,
```
aaronlademann-wf: This should be reverted since we only want to expose the new context stuff from `ReactDartInteropStatics2` class being created in #161.

lib/react_client.dart
```diff
@@ -276,10 +272,8 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
   }
 
   /// Wrapper for [Component.componentWillReceiveProps].
-  void handleComponentWillReceiveProps(
-          Component component,
-          ReactDartComponentInternal nextInternal,
-          InteropContextValue nextContext) =>
+  void handleComponentWillReceiveProps(Component component,
+          ReactDartComponentInternal nextInternal, dynamic nextContext) =>
```
aaronlademann-wf: I believe all of these changes within `ReactDartInteropStatics` will need to be [moved to `ReactDartInteropStatics2`](https://github.com/cleandart/react-dart/pull/161/files#diff-12ac054bb8b76f35c1035afef1d94396R497) once #161 merges.

The existing API should be left alone so that new context is completely opt-in.

lib/react_client.dart
```diff
@@ -217,11 +218,6 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
         return component;
       });
 
-  InteropContextValue handleGetChildContext(Component component) =>
```
aaronlademann-wf: If this is public API, it should be restored since a `ReactDartInteropStatics2` class is being created in #161.

Once all of the new context stuff is implemented in there, `ReactDartInteropStatics2.handleGetChildContext` should be deprecated with a comment explaining what to use instead.

lib/react_client.dart
```diff
@@ -167,23 +171,20 @@ dynamic _convertArgsToChildren(List childrenArgs) {
...
-  return new Map.fromIterable(_objectKeys(interopContext), value: (key) {
-    ReactDartContextInternal internal = getProperty(interopContext, key);
-    return internal?.value;
-  });
+dynamic _unjsifyContext(dynamic interopContext) {
+  return getProperty(interopContext, '___internal_dart_context_value___')
+      ?.value;
 }
 
+final _emptyJsMap = newObject();
```
aaronlademann-wf: Can we move this directly underneath the deprecated public `emptyJsMap` - just so its easier to see what happened / the reason for the deprecation of the public one?

lib/react_client.dart
aaronlademann-wf: Same goes for `handleShouldComponentUpdate`, `handeComponentWillUpdate`, etc.

lib/react_client.dart
```diff
@@ -364,7 +356,6 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
 
   return new ReactDartInteropStatics(
       initComponent: allowInterop(initComponent),
-      handleGetChildContext: allowInterop(handleGetChildContext),
```
aaronlademann-wf: This should be restored since the new stuff will be getting added to `ReactDartInteropStatics2`, and the existing `ReactDartInteropStatics` impl will be deprecated with no functional changes.

lib/react_client/js_interop_helpers.dart
```diff
@@ -26,17 +27,20 @@ class _MissingJsMemberError extends Error {
...
+/// __Deprecated: use [js_util.getProperty] instead.
+///
 /// Returns the property at the given [_GetPropertyFn.key] on the
 /// specified JavaScript object [_GetPropertyFn.jsObj]
 ///
 /// Necessary because `external operator[]` isn't allowed on JS interop classes
 /// (see: https://github.com/dart-lang/sdk/issues/25053).
 ///
 /// __Defined in this package's React JS files.__
+@Deprecated('4.0.0')
```
aaronlademann-wf:
```suggestion
@Deprecated('6.0.0')
```

lib/react_client/js_interop_helpers.dart
```diff
@@ -49,17 +53,20 @@ final _GetPropertyFn getProperty = (() {
...
+/// __Deprecated: use [js_util.setProperty] instead.
+///
 /// Sets the property at the given [_SetPropertyFn.key] to [_SetPropertyFn.value]
 /// on the specified JavaScript object [_SetPropertyFn.jsObj]
 ///
 /// Necessary because `external operator[]=` isn't allowed on JS interop classes
 /// (see: https://github.com/dart-lang/sdk/issues/25053).
 ///
 /// __Defined in this package's React JS files.__
+@Deprecated('4.0.0')
```
aaronlademann-wf:
```suggestion
@Deprecated('6.0.0')
```

lib/react_client/js_interop_helpers.dart
```diff
@@ -72,21 +79,27 @@ final _SetPropertyFn setProperty = (() {
   return _setProperty;
 })();
 
+/// __Deprecated: use [js_util.newObject] instead.
```
aaronlademann-wf:
```suggestion
/// __Deprecated:__ Use [js_util.newObject] instead.
```

lib/react_client/js_interop_helpers.dart
```diff
@@ -72,21 +79,27 @@ final _SetPropertyFn setProperty = (() {
...
+///
 /// An interop class for an anonymous JavaScript object, with no properties.
 ///
 /// For use when dealing with dynamic properties via [getProperty]/[setProperty].
 @JS()
 @anonymous
+@Deprecated('5.0.0')
```
aaronlademann-wf:
```suggestion
@Deprecated('6.0.0')
```

lib/react_client/js_interop_helpers.dart
```diff
@@ -72,21 +79,27 @@ final _SetPropertyFn setProperty = (() {
...
 class EmptyObject {
   external factory EmptyObject();
 }
 
+/// **Deprecated. use [js_util.jsify]/[jsifyAndAllowInterop] based on your needs.**
```
aaronlademann-wf:
```suggestion
/// __Deprecated.__ Use [js_util.jsify]/[jsifyAndAllowInterop] based on your needs.
```

lib/react_client/js_interop_helpers.dart
```diff
@@ -72,21 +79,27 @@ final _SetPropertyFn setProperty = (() {
...
+///
 /// Returns [map] converted to a JavaScript object, similar to
 /// `new JsObject.jsify` in `dart:js`.
 ///
 /// Recursively converts nested [Map]s, and wraps [Function]s with [allowInterop].
-EmptyObject jsify(Map map) {
-  var jsMap = new EmptyObject();
+@Deprecated('4.0.0')
```
aaronlademann-wf:
```suggestion
@Deprecated('6.0.0')
```

lib/react_client/react_interop.dart
```diff
@@ -152,8 +152,6 @@ class ReactComponent {
   external get refs;
   external void setState(state, [callback]);
   external void forceUpdate([callback]);
-
-  external bool isMounted();
```
aaronlademann-wf: Reminder: we need to deprecate this in the current master line using a `@Deprecated('5.0.0')` annotation.

lib/react_client/react_interop.dart
```diff
@@ -172,6 +170,18 @@ class InteropContextValue {
...
+/// A JavaScript interop class representing the return value of `createContext`.
+///
+/// Used for accessing Dart [Context.Provider] & [Context.Consumer] components.
+///
+/// __For internal/advanced use only.__
+@JS()
+@anonymous
+class ReactContext {
+  external get Provider;
+  external get Consumer;
```
aaronlademann-wf: Do we not know what either of these types are?

lib/react_client/react_interop.dart
```diff
@@ -314,7 +323,6 @@ class ComponentStatics {
 @anonymous
 class JsComponentConfig {
   external factory JsComponentConfig({
-    Iterable<String> childContextKeys,
-    Iterable<String> contextKeys,
+    contextType,
```
aaronlademann-wf: If `JsComponentConfig` is public API, it should merely be deprecated with a `@Deprecated('6.0.0')` annotation, and these changes should be rolled into a new `JsComponentConfig2` class.

test/factory/common_factory_tests.dart
```diff
@@ -126,7 +126,7 @@ void commonFactoryTests(ReactComponentFactoryProxy factory) {
   });
 }
 
-void domEventHandlerWrappingTests(ReactDomComponentFactoryProxy factory) {
+void domEventHandlerWrappingTests(ReactComponentFactoryProxy factory) {
```
aaronlademann-wf: ?

test/js_interop_helpers_test/shared_tests.dart
```diff
@@ -4,6 +4,8 @@ library js_function_test;
 import 'dart:html';
 
 import 'package:js/js.dart';
+import 'package:js/js_util.dart' show newObject;
+// TODO: remove in 5.0.0
```
aaronlademann-wf: What needs to be removed?

test/js_interop_helpers_test/shared_tests.dart
```diff
@@ -19,28 +21,34 @@ void verifyJsFileLoaded(String filename) {
 
 void sharedJsFunctionTests() {
   group('JS functions:', () {
+    // TODO: remove in 5.0.0
```
aaronlademann-wf: Are we able to deprecate `getProperty` in the current master line _(is there an alternative for consumers)_? Or do we need to deprecate in 5.0.0, and remove in 6.0.0?

test/js_interop_helpers_test/shared_tests.dart
```diff
@@ -50,7 +58,7 @@ void sharedJsFunctionTests() {
 
     group('markChildValidated', () {
       test('is function that does not throw when called', () {
-        expect(() => markChildValidated(new EmptyObject()), returnsNormally);
+        expect(() => markChildValidated(newObject()), returnsNormally);
```
aaronlademann-wf: This test should likely remain unchanged unless `markChildValidated` is backwards-compatible to still work with `EmptyObject` - in which case a test case for both should be maintained.

lib/react_client/react_interop.dart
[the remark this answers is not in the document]
greglittlefield-wf: I believe they can be, but I'm not sure if there's any runtime type-checking. +1 to generics.

Can we also add the parameter names?
```dart
int Function(T currentValue, T nextValue) calculateChangedBits,
```

lib/react_client.dart
kealjones-wk: The javascript in my blood doesn't agree, but i understand why you guys want to make it this way... :'(

test/factory/common_factory_tests.dart
kealjones-wk: Yeah this was before we removed the travis config for dart1 :P

lib/react.dart
kealjones-wk: literally anything... hence dynamic. lol but more often than not probably like `String`, `List`, `Map`, `int`, `bool`

lib/react_client/js_interop_helpers.dart
```diff
@@ -30,7 +30,7 @@ class _MissingJsMemberError extends Error {
 /// (see: https://github.com/dart-lang/sdk/issues/25053).
 ///
 /// __Defined in this package's React JS files.__
-final dynamic Function(dynamic jsObj, String key) getProperty = (() {
+final dynamic Function(dynamic jsObj, dynamic key) getProperty = (() {
```
kealjones-wk: These could be moved back to string, they are left over from when i was attempting the jsbackedmap indicated by a symbol path...

test/factory/common_factory_tests.dart
kealjones-wk: loosened to allow testing with the new ReactJsComponentFactoryProxy

test/js_interop_helpers_test/shared_tests.dart
kealjones-wk: It appears that `dart:js_util` has `getProperty`,`setProperty`, `newObject`, `hasProperty` and a few other fancy helpers we can use instead. Maybe i should just map all ours to those instead and then add a deprecation message to use that instead?

test/js_interop_helpers_test/shared_tests.dart
kealjones-wk: sure

lib/react_client.dart
```diff
@@ -404,6 +394,66 @@ ReactDartComponentFactoryProxy _registerComponent(
...
+          'Ensure that the JS component class you\'re referencing is available and being accessed correctly.');
+    }
+  }
+
+  @override
+  ReactElement build(Map<dynamic, dynamic> props, [List childrenArgs]) {
+    dynamic potentiallyConvertedProps;
+    dynamic children = _convertArgsToChildren(childrenArgs);
+
+    if (isContextComponent) {
```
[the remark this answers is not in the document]
kealjones-wk: done :)

lib/react_client/js_interop_helpers.dart
```diff
@@ -98,8 +111,91 @@ EmptyObject jsify(Map map) {
...
 
-    setProperty(jsMap, key, jsValue);
+    js_util.setProperty(jsMap, key, jsValue);
   });
 
   return jsMap;
 }
+
+// The following code is adapted from `package:js` in the dart-lang/sdk repo:
+// https://github.com/dart-lang/sdk/blob/1.21.0/sdk/lib/js_util/dart2js/js_util_dart2js.dart#L25
```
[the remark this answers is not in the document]
kealjones-wk: none that i could see

lib/react_client/react_interop.dart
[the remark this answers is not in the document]
kealjones-wk: when i tried to parameterize it i got strange issues with the values passed to the actual js layer, i ended up with as the best i could do, I believe the dart version of the `createContext` function does have the parameterization though. 
```
external static ReactContext createContext([
    dynamic defaultValue,
    int Function(dynamic currentValue, dynamic nextValue) calculateChangedBits,
  ]);
```

lib/react_client/js_interop_helpers.dart
kealjones-wk: These deprecations have been lumped into joe's ticket for master line un-deprecating of `shouldComponentUpdateWithContext` to be marked for 5.0.0 deprecation with alternative usage suggestions, this PR now completely removes them.

lib/react_client/js_interop_helpers.dart
kealjones-wk: see https://github.com/cleandart/react-dart/pull/160#discussion_r262942145

lib/react_client/js_interop_helpers.dart
kealjones-wk: see https://github.com/cleandart/react-dart/pull/160#discussion_r262942145

lib/react_client/js_interop_helpers.dart
kealjones-wk: see https://github.com/cleandart/react-dart/pull/160#discussion_r262942145

lib/react_client/js_interop_helpers.dart
kealjones-wk: see https://github.com/cleandart/react-dart/pull/160#discussion_r262942145

lib/react_client/js_interop_helpers.dart
kealjones-wk: see https://github.com/cleandart/react-dart/pull/160#discussion_r262942145

lib/react_client/react_interop.dart
kealjones-wk: added

test/js_interop_helpers_test/shared_tests.dart
kealjones-wk: we no longer use the get/setProperty functions from `js_interop_helpers.dart` and `_dart_helpers.js` in favor of using `dart:js_util` instead, so we no longer need the test coverage.

test/js_interop_helpers_test/shared_tests.dart
kealjones-wk: see https://github.com/cleandart/react-dart/pull/160#discussion_r262942145

test/js_interop_helpers_test/shared_tests.dart
kealjones-wk: see https://github.com/cleandart/react-dart/pull/160#discussion_r262942145

lib/react.dart
```diff
@@ -146,7 +147,7 @@ abstract class Component {
   /// Initializes context
   _initContext(context) {
     this.context = new Map.from(context ?? const {});
-    this.nextContext = this.context;
+    this.nextContext = new Map.from(this.context ?? {});
```
greglittlefield-wf: #nit this should be a `const` to prevent an extra object allocation
```suggestion
    this.nextContext = new Map.from(this.context ?? const {});
```

lib/react.dart
```diff
@@ -339,14 +332,7 @@ abstract class Component {
...
 
-  /// > __DEPRECATED - DO NOT USE__
-  /// >
-  /// > This API was never stable in any version of ReactJS, and was replaced with a new, incompatible context API
-  /// > in ReactJS 16 that is exposed via the [Component2] class.
-  /// >
-  /// > This will be completely removed when the JS side of it is slated for removal (ReactJS 17 / react.dart 6.0.0)
-  @Deprecated('6.0.0')
-  bool shouldComponentUpdateWithContext(Map nextProps, Map nextState, Map nextContext) => null;
+  bool shouldComponentUpdateWithContext(Map nextProps, Map nextState, dynamic nextContext) => null;
```
greglittlefield-wf: Why was this changed to `dynamic`?

lib/react_client.dart
```diff
@@ -229,6 +228,24 @@ Map<String, dynamic> _unjsifyContext(InteropContextValue interopContext) {
...
+  }
+  return context;
+}
+
+@JS()
+external get JsBackedMapSymbol;
+
+dynamic _unjsifyNewContext(dynamic interopContext) {
+  if (interopContext is JsMap) {
+    return JsBackedMap.copyToDart(interopContext);
```
greglittlefield-wf: This could result in type errors when the Map is converted back to Dart, since it's always going to be a `Map<dynamic, dynamic>`.

Can we remove the special case for maps?

lib/react_client.dart
```diff
@@ -493,20 +511,33 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
         component.componentWillReceiveProps(new JsBackedMap.backedBy(jsNextProps));
       });
 
-  void _updatePropsAndStateWithJs(Component2 component, JsMap props, JsMap state) {
+  void _updatePropsAndStateWithJs(Component2 component, JsMap props, JsMap state, [dynamic context]) {
     component
       ..props = new JsBackedMap.backedBy(props)
       ..state = new JsBackedMap.backedBy(state);
+    if (context != null) {
```
greglittlefield-wf: This shouldn't be wrapped in a conditional, to accommodate context changing from a non-null value to `null`.

lib/react_client.dart
```diff
@@ -534,6 +565,7 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
       });
 
   dynamic handleRender(Component2 component) => zone.run(() {
+        _updatePropsAndStateWithJs(component, component.jsThis.props, component.jsThis.state, component.jsThis.context);
```
greglittlefield-wf: Can we pass in props/state/context as arguments directly here?

Going through `jsThis` is slower because it causes an interceptor lookup.

lib/react_client.dart
```diff
@@ -583,25 +615,130 @@ ReactDartComponentFactoryProxy _registerComponent(ComponentFactory componentFact
...
+  /// Returns a JavaScript version of the specified [props], preprocessed for consumption by ReactJS and prepared for
+  /// consumption by the [react] library internals.
+  generateExtendedJsProps(Map props) {
+    JsBackedMap propsForJs = new JsBackedMap.from(props);
+
+    if (isProvider) {
+      propsForJs['value'] = _jsifyNewContext(propsForJs['value']);
+    }
+
+    if (convertDomProps) {
```
greglittlefield-wf: Is it valid to be able to pass DOM props to a context component? If not, can we remove this? Same question below with `ref`.

lib/react_client.dart
```diff
@@ -583,25 +615,130 @@ ReactDartComponentFactoryProxy _registerComponent(ComponentFactory componentFact
...
+        });
+      }
+    }
+
+    return factory(generateExtendedJsProps(new Map.from(props)), children);
+  }
+
+  /// Returns a JavaScript version of the specified [props], preprocessed for consumption by ReactJS and prepared for
+  /// consumption by the [react] library internals.
+  generateExtendedJsProps(Map props) {
```
greglittlefield-wf: This should have a return type.

lib/react_client.dart
```diff
@@ -583,25 +615,130 @@ ReactDartComponentFactoryProxy _registerComponent(ComponentFactory componentFact
...
+    if (isConsumer) {
+      if (children is Function) {
+        Function contextCallback = children;
+        children = allowInterop((args) {
+          return contextCallback(_unjsifyNewContext(args));
+        });
+      }
+    }
+
+    return factory(generateExtendedJsProps(new Map.from(props)), children);
```
greglittlefield-wf: This copy of the map is unnecessary since it's copied below.
```suggestion
    return factory(generateExtendedJsProps(props), children);
```

lib/react_client.dart
```diff
@@ -1020,6 +1157,25 @@ dynamic _findDomNode(component) {
   return ReactDom.findDOMNode(component is Component ? component.jsThis : component);
 }
 
+class ReactDartContext {
```
greglittlefield-wf: This could use a Doc comment.

lib/react_client.dart
```diff
@@ -1020,6 +1157,25 @@ dynamic _findDomNode(component) {
...
+  ReactDartContext(this.Provider, this.Consumer, this._jsThis);
```
greglittlefield-wf: Can we make this constructor private to decrease the public API surface area?

lib/react_client.dart
```diff
@@ -1020,6 +1157,25 @@ dynamic _findDomNode(component) {
...
+  final ReactContext _jsThis;
+  final ReactJsContextComponentFactoryProxy Provider;
+  final ReactJsContextComponentFactoryProxy Consumer;
+  ReactContext get jsThis => _jsThis;
+}
+
+ReactDartContext createContext([
```
greglittlefield-wf: This needs a doc comment

lib/react_client.dart
```diff
@@ -1020,6 +1157,25 @@ dynamic _findDomNode(component) {
...
+  dynamic defaultValue,
+  int Function(dynamic currentValue, dynamic nextValue) calculateChangedBits,
```
greglittlefield-wf: Starting up this conversation again; could we make this generic? Could something like this work?
```dart
ReactDartContext createContext<TValue>([
  TValue defaultValue,
  int Function(TValue currentValue, TValue nextValue) calculateChangedBits,
]) {
```
You mentioned running into issues with generics before; if you're still getting those, could you provide more details around what's not working?

lib/react_client.dart
```diff
@@ -1020,6 +1157,25 @@ dynamic _findDomNode(component) {
...
+]) {
+  var JSContext = React.createContext(defaultValue, allowInterop((currentValue, nextValue) {
```
greglittlefield-wf: I think `defaultValue` needs to be wrapped in `jsifyContext`, right? Can we add a test for that?

lib/react_client.dart
```diff
@@ -1020,6 +1157,25 @@ dynamic _findDomNode(component) {
...
+    return calculateChangedBits(_unjsifyNewContext(currentValue), _unjsifyNewContext(nextValue));
+  }));
+  return new ReactDartContext(new ReactJsContextComponentFactoryProxy(JSContext.Provider, isProvider: true),
+      new ReactJsContextComponentFactoryProxy(JSContext.Consumer, isConsumer: true), JSContext);
```
greglittlefield-wf: #nit trailing commas would help with readability here
```dart
  return new ReactDartContext(
    new ReactJsContextComponentFactoryProxy(JSContext.Provider, isProvider: true),
    new ReactJsContextComponentFactoryProxy(JSContext.Consumer, isConsumer: true),
    JSContext,
  );
```

lib/react_client.dart
greglittlefield-wf: Could we split this closure out to a local method? That helps provide better stack traces if something goes wrong in calculateChangedBits:
```dart
void handleCalculateChangedBits(currentValue, nextValue) { ... }
var JSContext = React.createContext(defaultValue, allowInterop(handleCalculateChangedBits));
```

test/lifecycle_test.dart
```diff
@@ -367,7 +369,7 @@ void sharedLifecycleTests<T extends react.Component>({
       final Map newPropsWithDefaults = unmodifiableMap({}..addAll(defaultProps)..addAll(newProps));
 
       const Map expectedState = const {};
-      const Map expectedContext = const {};
```
greglittlefield-wf: Why were these changed?

test/lifecycle_test.dart
```diff
@@ -166,6 +167,7 @@ void sharedLifecycleTests<T extends react.Component>({
   @required ReactDartComponentFactoryProxy ContextWrapper,
   @required ReactDartComponentFactoryProxy LifecycleTestWithContext,
   @required ReactDartComponentFactoryProxy LifecycleTest,
+  isComponent2: false,
```
greglittlefield-wf: This needs a type annotation

lib/react_client.dart
kealjones-wk: I dont think i can, I construct it in createContext

kealjones-wk:
Holding till https://github.com/cleandart/react-dart/pull/169 is done and merged to update its snapshot type tests to utilize the new shared type tester i added in this pr: https://github.com/cleandart/react-dart/pull/169#discussion_r267532067

example/js_components/js_components.dart
```diff
@@ -0,0 +1,79 @@
...
+
+import 'package:js/js.dart';
+import 'package:react/react.dart' as react;
+import 'package:react/react_client.dart';
+import 'package:react/react_client/react_interop.dart';
+import 'package:react/react_dom.dart' as react_dom;
+
+var fooRef;
+
+main() {
```
aaronlademann-wf: @greglittlefield-wf is this file sufficient to document / provide OSS explanation(s) for #109? Or should an example be created that shows how to "wrap" an existing JS component from a library like material-ui?

example/test/react_test_components.dart
```diff
@@ -223,23 +223,168 @@ class _ContextConsumerComponent extends react.Component {
...
+}
+
+var newContextRefComponent = react.registerComponent(() => new _NewContextRefComponent());
+
+int calculateChangedBits(currentValue, nextValue) {
+  int result = 1 << 1;
+  if (nextValue['renderCount'] % 2 == 0) {
+    result |= 1 << 2;
+  }
+  return result;
```
aaronlademann-wf: I have literally no idea what is happening, or why it is happening.  The second argument of `createContext` deserves some documentation explaining what it does.  For instance, why does it always return an integer?

example/test/react_test_components.dart
```diff
@@ -223,23 +223,168 @@ class _ContextConsumerComponent extends react.Component {
...
+  _onButtonClick(event) {
+    this.setState({'renderCount': this.state['renderCount'] + 1, 'complexMap': false});
+  }
+}
+
+var newContextProviderComponent = react.registerComponent(() => new _NewContextProviderComponent());
+
+class _NewContextConsumerComponent extends react.Component2 {
+  render() {
+    return TestNewContext.Consumer({'unstable_observedBits': props['unstable_observedBits']}, (value) {
```
aaronlademann-wf: Why are we using `unstable` ReactJS APIs in something extending from `Component2`?

example/test/react_test_components.dart
```diff
@@ -223,23 +223,168 @@ class _ContextConsumerComponent extends react.Component {
...
+      ]);
+    });
+  }
+}
+
+var newContextConsumerComponent = react.registerComponent(() => new _NewContextConsumerComponent());
+
+class _NewContextConsumerObservedBitsComponent extends react.Component2 {
+  render() {
+    return TestNewContext.Consumer({'unstable_observedBits': props['unstable_observedBits']}, (value) {
```
aaronlademann-wf: Same question here.  I don't understand why we would be utilizing an unstable context api... I thought the whole point of this was to transition to the new context _(which surely, is not unstable)_

js_src/_dart_helpers.js
```diff
@@ -1,9 +1,8 @@
 /**
  * react-dart JS interop helpers (used by react_client.dart and react_client/js_interop_helpers.dart)
  */
-function _getProperty(obj, key) { return obj[key]; }
-function _setProperty(obj, key, value) { return obj[key] = value; }
-
+var _reactDartSymbolPrefix = 'react-dart.';
+var _reactDartContextSymbol = Symbol.for(_reactDartSymbolPrefix+'context');
```
aaronlademann-wf: These could use some doc comments to explain what is happening / why it is necessary.

lib/react.dart
```diff
@@ -146,7 +147,7 @@ abstract class Component {
   /// Initializes context
   _initContext(context) {
     this.context = new Map.from(context ?? const {});
-    this.nextContext = this.context;
+    this.nextContext = new Map.from(this.context ?? const {});
```
aaronlademann-wf: It is confusing to read this since `nextContext` is `dynamic`, and yet, we can somehow safely pass it to the `Map.from` constructor, which only accepts a `Map`.

This deserves a comment (for both `context` and `nextContext`) explaining that the type was loosened from `Map` to `dynamic` for the purposes of being able to have `Component2 implements Component` - but that only "legacy" context APIs are supported in `Component` - which means it will still be a `Map` on the JS side of things.

lib/react.dart
```diff
@@ -474,6 +467,22 @@ abstract class Component2 implements Component {
   // TODO make private using expando?
   Component2Adapter adapter;
 
+  /// Accessed once and cached when instance is created. The [contextType] property on a class can be assigned
+  /// a [ReactDartContext] object created by [React.createContext]. This lets you consume the nearest current value of
+  /// that Context using [context].
+  ReactDartContext contextType;
```
aaronlademann-wf: Need an example of how to use this in the doc comment.

lib/react.dart
```diff
@@ -474,6 +467,22 @@ abstract class Component2 implements Component {
...
+  /// The context value from the [contextType] assigned to this component.
+  /// The value is passed down from the provider of the same [contextType].
+  /// You can reference [context] in any of the lifecycle methods including the render function.
+  ///
+  /// If you need multiple context values, consider using multiple consumers instead.
+  /// Read more: https://reactjs.org/docs/context.html#consuming-multiple-contexts
+  ///
+  /// This only has a value when [contextType] is set.
+  @override
+  dynamic context;
```
aaronlademann-wf: Need an example of how to use this in the doc comment.

lib/react_client.dart
```diff
@@ -587,25 +608,117 @@ ReactDartComponentFactoryProxy _registerComponent(ComponentFactory componentFact
   return new ReactDartComponentFactoryProxy(reactComponentClass);
 }
 
+class ReactJsContextComponentFactoryProxy extends ReactJsComponentFactoryProxy {
```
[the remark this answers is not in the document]
kealjones-wk: Probably could be private actually.

example/test/react_test_components.dart
kealjones-wk: The React team doesn't even have documentation around it. They act as if it doesn't exist even though it obviously does, i think this may still be a WIP/secret or experimental feature...

The idea is that you return a bit that indicates which parts/or how the context value changed, then you can use the observedBits props on a consumer and they will only re-render if the bits they are observing are returned from calculateChangedBits

example/test/react_test_components.dart
kealjones-wk: This is just an example but, the `unstable_observedBits` prop functions better than the stable version of the prop: `observedBits`... thats why its used here

example/test/react_test_components.dart
kealjones-wk: there is a "stable" version of this prop: `observedBits` but it doesn't work as well... Like i mentioned above the whole calculateChangedBits portion isn't even documented by react.

lib/react.dart
kealjones-wk: Sounds Good.

lib/react_client.dart
```diff
@@ -643,7 +756,7 @@ class ReactDomComponentFactoryProxy extends ReactComponentFactoryProxy {
     var convertibleProps = new Map.from(props);
     convertProps(convertibleProps);
 
-    return factory(jsify(convertibleProps), children);
+    return factory(jsifyAndAllowInterop(convertibleProps), children);
```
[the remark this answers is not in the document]
kealjones-wk: So that warning, is actually from the `jsify` doc comment. The only difference between `jsifyAndAllowInterop` and `jsify` is that it in addition to all the stuff `jsify` does it also wraps `allowInterop` around dart functions in the map.

lib/react_client.dart
```diff
@@ -493,20 +509,25 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
...
     component
       ..props = new JsBackedMap.backedBy(props)
-      ..state = new JsBackedMap.backedBy(state);
+      ..state = new JsBackedMap.backedBy(state)
+      ..context = _unjsifyNewContext(context);
   }
 
-  bool handleShouldComponentUpdate(Component2 component, JsMap jsNextProps, JsMap jsNextState) => zone.run(() {
-        final value = component.shouldComponentUpdate(
+  bool handleShouldComponentUpdate(Component2 component, JsMap jsNextProps, JsMap jsNextState, dynamic jsNextContext) =>
```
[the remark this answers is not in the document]
kealjones-wk: it doesn't need to be because on the JS side (where this is called in `_dart_helpers.js` it is always an argument so the value/whatever it may be will always be included, however i can make it optional just to match the dart api and just in case we ever change the JS side to not include it...

lib/react.dart
```diff
@@ -699,23 +706,23 @@ abstract class Component2 implements Component {
...
   @Deprecated('6.0.0')
-  void componentWillUpdateWithContext(Map nextProps, Map nextState, Map nextContext) {}
+  bool shouldComponentUpdateWithContext(Map nextProps, Map nextState, dynamic nextContext) => null;
 
   /// Do not use.
   ///
   /// Will be removed when [Component] is removed in the `6.0.0` release.
   @Deprecated('6.0.0')
-  void componentWillReceivePropsWithContext(Map newProps, nextContext) {}
+  void componentWillUpdateWithContext(Map nextProps, Map nextState, Map nextContext) {}
```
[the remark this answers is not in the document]
kealjones-wk: because this is in `Component`, it's still treated as a map

example/js_components/js_components.dart
kealjones-wk: Updated this example to use Material UI and A custom component

greglittlefield-wf: It looks like that is sufficient! 🏅

test/react_client/js_backed_map_test.dart
```diff
@@ -23,181 +24,17 @@ main() {
       JsBackedMap jsBackedMap;
       JsBackedMap dynamicJsBackedMap;
 
-      setUp(() {
+      testTypeValue(testValue) {
```
tainhenning-wk: The type information of testValue is no longer being propagated. Are we sure that this test is equivalent to the old one? @greglittlefield-wf

js_src/_dart_helpers.js
```diff
@@ -1,9 +1,11 @@
...
  * react-dart JS interop helpers (used by react_client.dart and react_client/js_interop_helpers.dart)
  */
-function _getProperty(obj, key) { return obj[key]; }
-function _setProperty(obj, key, value) { return obj[key] = value; }
-
+/// Prefix to namespace react-dart symbols
+var _reactDartSymbolPrefix = 'react-dart.';
+/// A global symbol to identify javascript objects owned by react-dart context,
+/// in order to jsify and unjsify context objects correctly.
+var _reactDartContextSymbol = Symbol.for(_reactDartSymbolPrefix+'context');
```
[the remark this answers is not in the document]
kealjones-wk: Nice catch for some reason I didn't fully realize i would be setting and fetching with the same instance...

lib/react.dart
```diff
@@ -474,6 +474,54 @@ abstract class Component2 implements Component {
...
+  /// Accessed once and cached when instance is created. The [contextType] property on a class can be assigned
+  /// a [ReactDartContext] object created by [React.createContext]. This lets you consume the nearest current value of
+  /// that Context using [context].
+  ///
+  /// __Example__:
+  ///
+  ///     var MyContext = createContext('test');
+  ///
+  ///     class MyClass extends react.Component2 {
+  ///       var contextType = MyContext;
```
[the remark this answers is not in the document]
kealjones-wk:
```suggestion
  ///       @override
  ///       var contextType = MyContext;
```

test/react_client/js_backed_map_test.dart
greglittlefield-wf: I was thinking about this and we may be able to use generics to propagate the static type information, and use JS inlining annotations from package:meta to force dart2js to use that type info (if it doesn't already):
```suggestion
      @tryInline
      void testTypeValue<T>(T testValue) {
```
I'll give that a shot locally.

Related: we should update the builder config for this repo to use the Dart 2 equivalent of the trust-type-annotations flag, since that's what we'll be using in prod most of the time.

test/shared_type_tester.dart
```diff
@@ -0,0 +1,148 @@
...
+@JS()
+library react_test_utils_test;
+
+import 'dart:async';
+import 'dart:html' as html;
+
+import 'package:js/js.dart';
+import 'package:test/test.dart';
+
+sharedTypeTests(Function testTypeValue,
```
greglittlefield-wf: This typing should be more specific:

```suggestion
typedef void TestTypeValue<T>(T value);
void sharedTypeTests(TestTypeValue testTypeValue,
```

test/shared_type_tester.dart
```diff
@@ -0,0 +1,148 @@
...
+    {skipNormalDartObjects: false,
```
greglittlefield-wf: Can we add typing (`bool`) to these?

test/react_context_test.dart
```diff
@@ -0,0 +1,159 @@
...
+                'contextToUse': TestCalculateChangedBitsContext,
+                'unstable_observedBits': 1 << 3,
+                'ref': (ref) {
+                  consumerOddRef = ref;
+                }
+              })
+            ]),
+            mountNode);
+      });
+      test('on first render', () {
```
greglittlefield-wf: #nit add spaces between tests
```suggestion

      test('on first render', () {
```

test/react_context_test.dart
```diff
@@ -0,0 +1,159 @@
...
+        expect(consumerEvenRef.latestValue, 1);
+        expect(consumerOddRef.latestValue, 1);
+      });
+      test('on value updates', () {
```
greglittlefield-wf: #nit add space between tests
```suggestion

      test('on value updates', () {
```

lib/react.dart
```diff
@@ -163,16 +171,9 @@ abstract class Component {
...
   /// Private reference to the value of [context] for the upcoming render cycle.
   ///
-  /// Useful for ReactJS lifecycle methods [shouldComponentUpdateWithContext] and [componentWillUpdateWithContext].
-  ///
-  /// > __DEPRECATED - DO NOT USE__
-  /// >
-  /// > This API was never stable in any version of ReactJS, and was replaced with a new, incompatible context API
-  /// > in ReactJS 16 that is exposed via the [Component2] class.
-  /// >
-  /// > This will be completely removed when the JS side of it is slated for removal (ReactJS 17 / react.dart 6.0.0)
```
greglittlefield-wf: Why was this part of the comment removed? Is it not still applicable?

lib/react.dart
```diff
@@ -563,16 +614,14 @@ abstract class Component2 implements Component {
   /// See: <https://facebook.github.io/react/docs/react-component.html#updating-componentwillreceiveprops>
```
greglittlefield-wf: Can we remove the note in the comment above about `componentWillReceivePropsWithContext`?

test/react_context_test.dart
```diff
@@ -0,0 +1,159 @@
...
+              consumerRef = ref;
+            }
+          })
+        ]),
+        mountNode);
+    expect(contextTypeRef.context, same(typeToTest));
+    expect(consumerRef.latestValue, same(typeToTest));
+  }
+
+  group('New Context API (Component2 only)', () {
```
greglittlefield-wf: Are there any tests that the correct `nextContext` arg is passed into `shouldComponentUpdate`? I can't seem to find any. If not, could you add some?

example/test/react_test_components.dart
aaronlademann-wf: Sorry for the crazy delayed reply here @kealjones-wk - a doc comment about how you perceive it to work (like what you just typed above) would be a great addition!

example/test/react_test_components.dart
aaronlademann-wf: Sorry for the delayed reply.

I get that its just an example... but it is an example intended to exercise the functionality in-lieu of a formal functional test - which is why I'm concerned about our implementation possibly not functioning well with the stable version of `observedBits`.

test/shared_type_tester.dart
aaronlademann-wf: We should avoid unnecessary `typedef`s since we're Dart 2 only now... we can just use the new inline `Function` syntax.

lib/react.dart
```diff
@@ -171,9 +171,16 @@ abstract class Component {
...
+  ///
+  /// > __DEPRECATED - DO NOT USE__
+  /// >
+  /// > This API was never stable in any version of ReactJS, and was replaced with a new, incompatible context API
+  /// > in ReactJS 16 that is exposed via the [Component2] class.
+  /// >
+  /// > This will be completely removed when the JS side of it is slated for removal (ReactJS 17 / react.dart 6.0.0)
   @Deprecated('6.0.0')
-  dynamic nextContext;
+  Map nextContext;
```
aaronlademann-wf: Doesn't this have to remain `dynamic` since `Component2` implements `Component`?

test/react_client/js_backed_map_test.dart
greglittlefield-wf: It looks manually inlining doesn't do much for the dart2js output, but we should add the generics so that both tests aren't essentially just `as dynamic`.

I'll make a PR to add those and also add the build.yaml and test changes

test/shared_type_tester.dart
greglittlefield-wf: Good point; the only reason I put in the typedef was because I wasn't sure if the semantics were the same for generics

greglittlefield-wf: Addressed in https://github.com/kealjones-wk/react-dart/pull/3

js_src/_dart_helpers.js
```diff
@@ -1,9 +1,11 @@
 /**
  * react-dart JS interop helpers (used by react_client.dart and react_client/js_interop_helpers.dart)
  */
-function _getProperty(obj, key) { return obj[key]; }
-function _setProperty(obj, key, value) { return obj[key] = value; }
-
+/// Prefix to namespace react-dart symbols
+var _reactDartSymbolPrefix = 'react-dart.';
+/// A global symbol to identify javascript objects owned by react-dart context,
```
greglittlefield-wf: #nit
```suggestion
/// A symbol to identify JavaScript objects owned by react-dart context,
```

lib/react_client.dart
```diff
@@ -493,20 +515,24 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
...
     component
       ..props = new JsBackedMap.backedBy(props)
-      ..state = new JsBackedMap.backedBy(state);
+      ..state = new JsBackedMap.backedBy(state)
+      ..context = _unjsifyNewContext(context);
   }
 
-  bool handleShouldComponentUpdate(Component2 component, JsMap jsNextProps, JsMap jsNextState) => zone.run(() {
+  bool handleShouldComponentUpdate(Component2 component, JsMap jsNextProps, JsMap jsNextState,
+          [dynamic jsNextContext]) =>
```
greglittlefield-wf: This shouldn't be optional, right?

lib/react_client.dart
```diff
@@ -587,25 +612,117 @@ ReactDartComponentFactoryProxy _registerComponent(ComponentFactory componentFact
   return new ReactDartComponentFactoryProxy(reactComponentClass);
 }
 
+class _ReactJsContextComponentFactoryProxy extends ReactJsComponentFactoryProxy {
```
greglittlefield-wf: This could use a doc comment

kealjones-wk:
> * Comment about testing the `nextContext` value passed into shouldComponentUpdate (the only test we have uses `null`; we should also have a test with another value)

@greglittlefield-wf if the only test you're referring to is this one: https://github.com/cleandart/react-dart/pull/160/files#diff-c00a02aa016b25a86fb30a2093f935a1R520

because that component it is a child of a context.consumer callback and not a consumer/contextType component itself it will actually never have a `this.context` the only time `this.context` ever has a value is when `contextType` is set, and when `contextType` is set context and prop updates don't trigger `shouldComponentUpdate`.

The only time i have seen `shouldComponentUpdate` and `nextContext` be called and have a value is on a component using `contextType` that has a state change. But this `nextContext` is no different than `this.context` at this point so its utterly useless.

greglittlefield-wf:
> But this `nextContext` is no different than `this.context` at this point so its utterly useless.

Yeah, I think we just need to verify that when it is passed in, that it's valid.

PR #161, opened 2019-03-04
JS-object-based props/state, Component2 initial impl, JsMap

greglittlefield-wf:
#### Initial implementation of JS-object-based props/state
- Implemented in new `Component2` base class 
    - This API is extremely far from final; it currently contains pieces that will be removed or completely reworked before release
- Dart component props are now stored directly on the JS `props` object, as opposed to within a Dart object stored as an `internal` prop
- Dart component state is now also stored directly on the JS `state` object, as opposed to completely within the Dart layer

Advantages: 
- Props/state are now visible/mutable in the React Dev Tools, greatly improving the Dart component debugging experience
- Complexity of the Dart/JS bindings are reduced significantly by leveraging React JS logic more directly, allowing easier maintenance and reduced risk of bugs
    - This will be especially important when implementing the new React lifecycle methods and [concurrent mode](https://reactjs.org/blog/2018/11/27/react-16-roadmap.html#react-16x-q2-2019-the-one-with-concurrent-mode)
- Performance is increased by a small amount by avoiding Dart Map overhead

#### JsBackedMap
A Dart class that adapts JS objects to the Dart `Map` interface, allowing properties to be easily read/written directly. 
- Used as the basis for props/state Maps used in `Component2`'s fields and lifecycle methods 
- Used to replace some existing logic in `ReactDomComponentFactory` that served similar functions, removing an extra copy in DOM `ReactElement` creation

#### Etc.
- Update tests to run on `Component2`
- Add basic tests for JsBackedMap (more test coverage will come in a later PR)
- Bump SDK constraint to be Dart-2-only
- Remove Dart 1 build from CI

## Testing
- CI passes


lib/react_client/react_interop.dart
```diff
@@ -149,6 +159,7 @@ class ReactElement {
 class ReactComponent {
   external Component get dartComponent;
   external InteropProps get props;
+  external InteropProps get state;
```
aaronlademann-wf: Is this a typo, or are `props` and `state` both instances of `InteropProps`?

thegregorator: Ooh good catch, this should be JsMap and maybe InteropProps should extend JsMap so we can remove some of those casts. I'll look into that

lib/react_client.dart
```diff
@@ -146,6 +150,73 @@ class ReactDartComponentFactoryProxy<TComponent extends Component>
...
+      children,
+    );
+  }
+
+  /// Returns a JavaScript version of the specified [props], preprocessed for consumption by ReactJS and prepared for
+  /// consumption by the [react] library internals.
+  static JsMap generateExtendedJsProps(Map props) {
+    final propsForJs = new JsBackedMap.from(props);
+
+    // FIXME forwarded DOM props???
```
kealjones-wk: Would DOM props not be forwarded? seems like this works without any intervention...

lib/react_client/react_interop.dart
greglittlefield-wf: Ooh good catch, this should be JsMap and maybe InteropProps should extend JsMap so we can remove some of those casts. I'll look into that

lib/react_client.dart
```diff
@@ -376,12 +448,177 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
...
+final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
+  final zone = Zone.current;
+
+  /// Wrapper for [Component.getInitialState].
+  Component2 initComponent(ReactComponent jsThis,
+          ComponentStatics<Component2> componentStatics) =>
+      zone.run(() {
+        final component = componentStatics.componentFactory();
+        component.adapter =
+            new JsComponent2Adapter(jsThis: jsThis); // TODO displayName;
```
kealjones-wk: seems done to me? am i missing something?

lib/src/react_client/js_backed_map.dart
```diff
@@ -0,0 +1,139 @@
...
+  bool _isValidKey(Object key) => key == null || key is String;
+
+  // ----------------------------------
+  // Core overrides
+  // ----------------------------------
+
+  @override
+  V operator [](Object key) {
+//    if (!_isValidKey(key)) return null;
+//    assert(_isValidKey(key));
```
kealjones-wk: Get this guff outa here.

lib/src/react_client/js_backed_map.dart
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // todo this cast seems to work in Dart 1 without much overhead in dart2js, but may break in DDC/Dartium and may also break reified types
+  @override
+  Iterable<K> get keys => _Object.keys(jsObject) as List<K>;
+
+  @override
+  V remove(Object key) {
+//    assert(_isValidKey(key));
+//    if (!_isValidKey(key)) return null;
```
kealjones-wk: more guff.

lib/src/react_client/js_backed_map.dart
```diff
@@ -0,0 +1,139 @@
...
+      _Object.assign(jsObject, (other as JsBackedMap).jsObject);
+    } else {
+      super.addAll(other);
+    }
+  }
+
+  @override
+  bool containsKey(Object key) {
+//    if (!_isValidKey(key)) return false;
+//    assert(_isValidKey(key));
```
kealjones-wk: and again, more guff

lib/src/react_client/js_backed_map.dart
```diff
@@ -0,0 +1,139 @@
...
+  @override
+  Iterable<V> get values => _Object.values(jsObject) as List<V>;
+
+// TODO not sure if overriding makes sense; check to see which is more optimal for smaller maps
+//  @override
+//  bool containsValue(Object value) => values.contains(value);
+
+// overridden for to avoid cast in case the implementation of keys changes
+//  @override
+//  int get length => _Object.keys(jsObject).length;
```
kealjones-wk: from line 99 to here all can probably go, unless the worry of implementation change is still relevant then id probably un-comment it out.

lib/react.dart
```diff
@@ -371,6 +375,312 @@ abstract class Component {
...
 }
 
+abstract class Component2Adapter {
+  void setState(dynamic newState, SetStateCallback callback);
+  void forceUpdate(SetStateCallback callback);
+}
+
+/// Top-level ReactJS [Component class](https://facebook.github.io/react/docs/react-component.html)
+/// which provides the [ReactJS Component API](https://facebook.github.io/react/docs/react-component.html#reference)
+abstract class Component2<T extends Map> implements Component {
```
kealjones-wk: Why do we implement Component?

lib/react.dart
```diff
@@ -371,6 +375,312 @@ abstract class Component {
...
+  /// See: <https://facebook.github.io/react/docs/react-component.html#render>
+  dynamic render();
+
+  // Unsupported things
+
+  /// Do not use.
+  ///
+  /// Will be removed when [Component] is removed in the `6.0.0` release.
+  @Deprecated('6.0.0')
+  void replaceState(Map newState, [SetStateCallback callback]) =>
```
kealjones-wk: Why is this deprecated in `Component2` and not in `Component`? why do we even need this in `Component2`?

lib/react_client.dart
```diff
@@ -376,12 +448,177 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
...
+        );
+
+        _updatePropsAndStateWithJs(component, jsNextProps, jsNextState);
+      });
+
+  void handleComponentDidUpdate(Component2 component, ReactComponent jsThis,
+          JsMap jsPrevProps, JsMap jsPrevState) =>
+      zone.run(() {
+//    final prevProps = component.props; // todo do this instead of creating new wrappers?
+//    final prevState = component.state; // todo do this instead of creating new wrappers?
```
kealjones-wk: Don't we need it to be a JSBackedMap?

lib/react.dart
greglittlefield-wf: For compatibility purposes with existing APIs that accept/return a `Component`. It's not ideal to have to stub in all these unused methods and things, but I think it's worth it for this class to be compatible.

lib/react.dart
greglittlefield-wf: Good point, it should probably be deprecated in Component, along with any other non-lifecycle members that will be unsupported in Component2.

aaronlademann-wf: I will add "non lifecycle members of `Component`" to our deprecation ticket.

greglittlefield-wf: I say we do that in the PR where we add all the other deprecations

PR #163, opened 2019-03-06
Fix DOM component callback ref regression

greglittlefield-wf:
### Problem
DOM callback refs weren't working right in dart2js because they weren't being `allowInterop`'d

This bug was introduced as part of cleanup in the JS map PR.

### Solution
- Revert changes that broke behavior
- Add regression tests for callback refs

### Testing
- Passing CI


kealjones-wk:
+1

PR #162, opened 2019-03-06
CPLAT-4335 Add deprecations in preparation for 5.0.0

aaronlademann-wf:
As part of preparations for ReactJS 16 support coming in version 5.0.0, we need to alert consumers to some important deprecations by annotating those deprecated members, and providing comments about what to use instead _(if applicable)_.

### Deprecations:

1. [Legacy `context` APIs](https://github.com/cleandart/react-dart/pull/162/commits/f8cd44a63bba1ba8339f0a6a740411e0accc9189)
1. [`isMounted`](https://github.com/cleandart/react-dart/pull/162/commits/33f4f7508801b1549f8a83a7a0a2b40458256451)
1. [`react_test_utils.SimulateNative`](https://github.com/cleandart/react-dart/pull/162/commits/e5398815df1c0567e18db16caae74341df265758)
1. [String `Component.ref`s](https://github.com/cleandart/react-dart/pull/162/commits/911df628b492e61ebb30b6cda94ab351fb791728)
1. [`Component.replaceState`s](https://github.com/cleandart/react-dart/pull/162/commits/9486382c2744ed9ac2bef6fefe580c985d5e3837)
1. [All aspects of server-side rendering](https://github.com/cleandart/react-dart/pull/162/commits/e259daddc769979226041acfb904af42000cf7ee)
1. [`Component.bind`](https://github.com/cleandart/react-dart/pull/162/commits/e849f4b1e9550e96ea89183c9e3bf49911cce97b)
1. [`Component.transferComponentState`](https://github.com/cleandart/react-dart/pull/162/commits/dd5db69f231df54d1d5b2d3616d2c4caab94c63e)

---

@kealjones-wk @joebingham-wk @greglittlefield-wf

PR #164, opened 2019-03-06
CPLAT-4787 Remove 5.0.0 deprecations, Add / update 6.0.0 deprecations

aaronlademann-wf:
- [ ] Adds deprecations that we couldn't make in the master line via #162
- [ ] Updates comments on existing `6.0.0` deprecations
- [ ] Adds a few new `6.0.0` deprecations
- [ ] Removes all `5.0.0` deprecations


lib/react_client/react_interop.dart
```diff
@@ -40,16 +40,6 @@ abstract class ReactDom {
       ReactDOM.unmountComponentAtNode(element);
 }
 
-/// __DEPRECATED.__
-///
-/// Server-side rendering support via APIs in the `react` package will not be maintained beyond the `5.0.0` release.
-@Deprecated('5.0.0')
-@JS('ReactDOMServer')
```
greglittlefield-wf: Actually, I'm not sure we should be removing this... This actually doesn't relate to the react_server entrypoint (we can still use dart:html with this API) and there's not much we have to do to maintain it.

kealjones-wk:
Looks good to be but can we also throw in removals of the js_src files?
```diff
diff --git a/js_src/react_dom_server.js b/js_src/react_dom_server.js
deleted file mode 100644
index e66bcba..0000000
--- a/js_src/react_dom_server.js
+++ /dev/null
@@ -1,3 +0,0 @@
-const ReactDOMServer = require('react-dom/server');
-
-module.exports = ReactDOMServer;
diff --git a/webpack.config.js b/webpack.config.js
index 3ad5212..74f4a13 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -76,10 +76,8 @@ module.exports = createExports(
       ["react.js",                "react_with_addons.js"],
       ["react.js",                "react.js"],
       ["react_dom.js",            "react_dom.js"],
-      ["react_dom_server.js",     "react_dom_server.js"],
       ["react.js",                "react_prod.js"],
       ["react_dom.js",            "react_dom_prod.js"],
-      ["react_dom_server.js",     "react_dom_server_prod.js"],
       ["react_with_react_dom.js", "react_with_react_dom_prod.js"],
     ]
   );
```

kealjones-wk:
+1
Also would like final approval from @greglittlefield-wf

lib/react_client.dart
```diff
@@ -1112,9 +1099,4 @@ void setClientConfiguration() {
   setReactConfiguration(_reactDom, _registerComponent);
   setReactDOMConfiguration(
       ReactDom.render, ReactDom.unmountComponentAtNode, _findDomNode);
-  // Accessing ReactDomServer.renderToString when it's not available breaks in DDC.
-  if (context['ReactDOMServer'] != null) {
```
greglittlefield-wf: This needs to stay in place

lib/react_client/react_interop.dart
```diff
@@ -384,16 +322,39 @@ bool get inReactDevMode => _inReactDevMode;
...
+    bool Function(
+      Component component,
+      InteropContextValue nextContext,
+    )
+        handleShouldComponentUpdate,
+    void Function(
+      Component component,
+      InteropContextValue nextContext,
+    )
+        handleComponentWillUpdate,
```
greglittlefield-wf: #nit could we remove the trailing commas on some of these so that the dartfmt output isn't so bad?

lib/react_dom_server.dart
```diff
@@ -1,31 +0,0 @@
-// Copyright (c) 2016, the Clean project authors.  Please see the AUTHORS file
```
greglittlefield-wf: This and all the `react_dom_server*` JS files needs to stay in place

greglittlefield-wf:
+10

aaronlademann-wf:
+1

PR #165, opened 2019-03-07
CPLAT-4764 React16 setState updates

aaronlademann-wf:
> ~~__NOT A CLEAN DIFF.__  Will not be a clean diff until #162 / #164 are merged.~~

### Problem 1
In ReactJS 16, calling `setState` with a null value no longer triggers an update. However, our interop layer results in an update happening regardless because we always pass an empty `Map` into the JS layer.

### Solution 1
https://github.com/cleandart/react-dart/commit/d382d716c2ab3c50a6b558e8887b2301076e72b4
* Update logic within `Component`, and within `JsComponent2Adapter` so that a `null` value passed as the first argument of `setState` does not result in a component update.
* Add tests

---

### Problem 2
In `Component`, the first argument of `setState` is `dynamic` so that either a `Map`, or a `TransactionalSetStateCallback` can be provided. Dynamic is gross.

### Solution 2
https://github.com/cleandart/react-dart/commit/cbee295abd35357e6539cef7c40ca2449fb06d29
* Tighten the type on the first argument of `Component2.setState` to be a `Map`
* Add a new `setStateTransaction` method that has a first argument value type of `TransactionalSetStateCallback`.
* Add tests

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -106,7 +106,7 @@ abstract class Component {
 
   List<SetStateCallback> _setStateCallbacks = [];
 
-  List<TransactionalSetStateCallback> _transactionalSetStateCallbacks = [];
+  List<StateUpdaterCallback> _transactionalSetStateCallbacks = [];
 
   /// The List of callbacks to be called after the component has been updated from a call to [setState].
   List get setStateCallbacks => _setStateCallbacks;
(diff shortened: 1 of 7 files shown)


lib/src/react_client/js_backed_map.dart
```diff
@@ -130,6 +130,8 @@ abstract class _Reflect {
 }
 
 JsMap jsBackingMapOrJsCopy(Map other) {
+  other ??= const {};
```
greglittlefield-wf: Is this null-aware needed? I would prefer to keep this nullable.

lib/react.dart
```diff
@@ -419,13 +502,24 @@ abstract class Component2 implements Component {
...
   }
 
+  /// Triggers a rerender with new state obtained by shallow-merging
+  /// the return value of [stateUpdater] into the current [state].
+  ///
+  /// Optionally accepts a [callback] that gets called after the component updates.
+  ///
+  /// See: <https://reactjs.org/docs/react-component.html#setstate>
+  void setStateTransaction(TransactionalSetStateCallback stateUpdater,
+      [SetStateCallback callback]) {
```
greglittlefield-wf: We don't have to land on the final name right now, but it looks like the word "transaction" isn't ever used in the react docs: https://reactjs.org/docs/react-component.html#setstate ; they use "updater" instead:
> ```
> setState(updater[, callback])
>```


Could we update the naming to match? Might also be nice to have inline function types, since typedefs are harder to make sense of in the IDE:
```dart
void setStateWithUpdater(Map updater(Map state, Map props), [void callback()]);
```

PR #166, opened 2019-03-07
CPLAT-4789: fix string refs

kealjones-wk:
## Issue
String refs were breaking due to a bundling issue in webpack when using 2 seperate files that would require('react'). Each file would use a different instance of react.

## Solution
Update webpack config to use the window.React instance for any require('react') that are not the initial provider of react.


webpack.config.js
```diff
@@ -47,19 +47,23 @@ function createExports(exportMappings) {
...
+        },
+        entry: path.resolve(inputPath, entryFilename),
+        plugins: isProduction ? prodPlugins : devPlugins,
+        mode: isProduction ? "production" : "development",
+        externals: [{ window: "window" }],
+        devtool: "source-map",
+      };
+
+      if ( entryFilename != 'react.js' && !entryFilename.includes('react_with') ) {
+        exportObject.externals[0].react = "window.React";
```
greglittlefield-wf: Could you add a comment saying why this is necessary?

webpack.config.js
```diff
@@ -47,19 +47,23 @@ function createExports(exportMappings) {
...
+          filename: outputFilename
+        },
+        entry: path.resolve(inputPath, entryFilename),
+        plugins: isProduction ? prodPlugins : devPlugins,
+        mode: isProduction ? "production" : "development",
+        externals: [{ window: "window" }],
+        devtool: "source-map",
+      };
+
+      if ( entryFilename != 'react.js' && !entryFilename.includes('react_with') ) {
```
greglittlefield-wf: #nit could we split this out to a variable?
```js
includeReact = entryFilename === 'react.js' || entryFilename.includes('react_with');

webpack.config.js
kealjones-wk: I added a comment but i am not so great with the word so feel free to suggest something better :)

greglittlefield-wf: LGTM

PR #155, opened 2019-02-11
Dart "pana" health check fixes

aaronlademann-wf:
Fixing the lints shown by [the Dart "pana" health check on pub.dartlang.org](https://pub.dartlang.org/packages/react/versions/4.6.0#-analysis-tab-)

image

@greglittlefield-wf @kealjones-wk @evanweible-wf

PR #167, opened 2019-03-08
Release React 4.6.2

aaronlademann-wf:
This release includes:

- #162: **Important Deprecations**
    > These deprecations are being put in place to prepare consumers for the upcoming `5.0.0` release which will include support for React JS version 16.x
    - **All aspects of server-side rendering**
    - Legacy `context` APIs
    - `isMounted`
    - `react_test_utils.SimulateNative`
    - String `Component.ref`s
    - `Component.replaceState`s
    - `Component.bind`
    - `Component.transferComponentState`

- #155: Clean the lint trap.

---


lib/react_client/react_interop.dart
```diff
@@ -39,8 +39,6 @@ abstract class ReactDom {
 /// __DEPRECATED.__
```
greglittlefield-wf: This notice needs to be removed, as do the deprecations within `react_dom_server.dart`

greglittlefield-wf:
+10, merging/releasing

PR #169, opened 2019-03-14
Implement getSnapshotBeforeUpdate (CPLAT-4766)

joebingham-wk:
## Description:

React16 added the lifecycle method `getSnapshotBeforeUpdate`. We needed to add this to Component2 and begin the transition of removing older lifecycle events that will no longer work.

## Changes

- Add the new lifecycle event to react.dart, react_client, react_interop, and _dart_helpers.  
- Add an optional third parameter to `componentDidUpdate`.
- Add a mixin that can be used to ensure type consistency between the return value of `getSnapshotBeforeUpdate` and the new third parameter for `componentDidUpdate`.
- Remove componentWillUpdate from react_interop. 
- Add deprecation messages to the method. 
- Add new tests.
- Adjust old tests that relied upon lifecycle methods that are no longer being called for Component2.

## Testing suggestions:

- CI Passes
- Double check that the new added tests are not false positives
- Manually test the new lifecycle methods and ensure the return value of `getSnapshotBeforeUpdate` gets passed to `componentDidUpdate` successfully.

## Potential areas of regression:

There shouldn't be any for consumers as Component2 is new, but any code reliant upon old Component2 lifecycle events would now break.


lib/react.dart
```diff
@@ -775,6 +829,44 @@ abstract class Component2 implements Component {
...
 }
 
+/// Creates a mixin used to ensure consistent typing of the `snapshot`
+/// parameter passed between Component2.getSnapshotBeforeUpdate and
+/// Component2.componentWillUpdate.
+///
+/// __EXAMPLE__:
+///
+/// class _ParentComponent extends react.Component2 implements react
+///    .TypedSnapshot<int> {
```
greglittlefield-wf: Could we put this on the previous line, or insert a newline `implements` instead?

lib/react.dart
```diff
@@ -775,6 +829,44 @@ abstract class Component2 implements Component {
...
+///       return true;
+///     }
+///     return null;
+///   }
+///
+///   // Note that the snapshot parameter is consistently typed with the return
+///   // value of getSnapshotBeforeUpdate.
+///   void componentDidUpdate(prevProps, prevState, [int snapshot]) {
+///     if (snapshot == true) {
+///       //Do something
```
greglittlefield-wf: #nit add spaces after comments
```suggestion
///       // Do something
```

lib/react.dart
```diff
@@ -775,6 +829,44 @@ abstract class Component2 implements Component {
...
+///     }
+///   }
+///
+///   /*Include other standard component logic*/
```
greglittlefield-wf: #nit add spaces before/after comments
```suggestion
///   /* Include other standard component logic */
```

lib/react_client.dart
```diff
@@ -512,20 +512,23 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
...
-          new JsBackedMap.backedBy(jsNextProps),
-          new JsBackedMap.backedBy(jsNextState),
+  dynamic handleGetSnapshotBeforeUpdate(Component2 component, JsMap jsPrevProps, JsMap jsPrevState) => zone.run(() {
+        final snapshotValue = component.getSnapshotBeforeUpdate(
+          new JsBackedMap.backedBy(jsPrevProps),
+          new JsBackedMap.backedBy(jsPrevState),
         );
 
-        _updatePropsAndStateWithJs(component, jsNextProps, jsNextState);
+        return snapshotValue;
```
greglittlefield-wf: Hmm since this is passed through the JS layer, it may need to be wrapped in an object to ensure it doesn't get jsified. I'll try this out when testing locally and let you know.

test/lifecycle_test.dart
```diff
@@ -145,11 +147,33 @@ main() {
...
+
+        expect(
+            component.lifecycleCalls,
+            containsAllInOrder([
+              matchCall('getSnapshotBeforeUpdate'),
+              matchCall('componentDidUpdate', args: [
+                anything,
+                anything,
+                'sna'
+                    'pshot'
```
greglittlefield-wf: This string should be combined

test/lifecycle_test.dart
```diff
@@ -145,11 +147,33 @@ main() {
...
-              matchCall('componentWillUpdate', args: [expectedProps, initialState], state: initialState),
               matchCall('render', state: initialState),
-              matchCall('componentDidUpdate', args: [expectedProps, initialState], state: initialState),
+              matchCall('getSnapshotBeforeUpdate', args: [expectedProps, initialState], state: initialState),
+              matchCall('componentDidUpdate', args: [expectedProps, initialState, null], state: initialState),
             ].where((matcher) => matcher != null).toList()));
       });
+
+      test(
+          'getSnapshotBeforeUpdate returns the same type being passed into '
```
greglittlefield-wf:
```suggestion
          'getSnapshotBeforeUpdate returns the same value being passed into '
```

test/lifecycle_test.dart
```diff
@@ -467,13 +519,20 @@ void sharedLifecycleTests<T extends react.Component>({
...
                 ? null
                 : matchCall('componentWillUpdateWithContext',
                     args: [expectedProps, initialState, newContext], state: initialState),
             matchCall('render', state: initialState),
-            matchCall('componentDidUpdate', args: [expectedProps, initialState], state: initialState),
+            isComponent2
+                ? matchCall('getSnapshotBeforeUpdate', args: [expectedProps, initialState], state: initialState)
+                : null,
+            isComponent2
+                ? matchCall('componentDidUpdate', args: [expectedProps, initialState, null], state: initialState)
```
greglittlefield-wf: #nit should be `snapshot`
```suggestion
                ? matchCall('componentDidUpdate', args: [expectedProps, initialState, snapshot], state: initialState)
```

test/lifecycle_test.dart
```diff
@@ -368,6 +402,7 @@ void sharedLifecycleTests<T extends react.Component>({
 
       const Map expectedState = const {};
       const Map expectedContext = const {};
+      const Null snapshot = null;
```
greglittlefield-wf: I think it'd be clearer if this were named `expectedSnapshot`

lib/react.dart
```diff
@@ -775,6 +829,44 @@ abstract class Component2 implements Component {
...
+///
+/// }
+
+abstract class TypedSnapshot<TSnapshot> implements Component2 {
```
greglittlefield-wf: Can we also add an example using this?

js_src/_dart_helpers.js
```diff
@@ -85,26 +85,17 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics) {
...
-      dartInteropStatics.handleComponentWillReceiveProps(this.dartComponent, nextProps);
-    }
     shouldComponentUpdate(nextProps, nextState) {
       return dartInteropStatics.handleShouldComponentUpdate(this.dartComponent, nextProps, nextState);
     }
-    // FIXME remove unsafe members when implementing new React 16 lifecycle methods
-    UNSAFE_componentWillUpdate(nextProps, nextState) {
-      dartInteropStatics.handleComponentWillUpdate(this.dartComponent, nextProps, nextState);
+    getSnapshotBeforeUpdate(prevProps, prevState) {
+      return dartInteropStatics.handleGetSnapshotBeforeUpdate(this.dartComponent, prevProps, prevState);
```
greglittlefield-wf: When testing a component without `getSnapshotBeforeUpdate`, I get the following error:
> Warning: _ClockComponent.getSnapshotBeforeUpdate(): A snapshot value (or null) must be returned. You have returned undefined.

I think we need the same undefined check that we have in render a couple lines down, since we can't guarantee in Dart whether its `null` will be converted to `null` or `undefined` in JS (even if you `return null;` explicitly)

test/lifecycle_test.dart
```diff
@@ -807,7 +876,10 @@ void sharedLifecycleTests<T extends react.Component>({
         // Check against the JS component to ensure no regressions.
         expect(component.state['counter'], 3);
         expect(component.state['counter'], getLatestJSCounter());
-        expect(component.lifecycleCalls, orderedEquals(getUpdatingSetStateLifeCycleCalls()));
+
+        if (!isComponent2) {
+          expect(component.lifecycleCalls, orderedEquals(getUpdatingSetStateLifeCycleCalls()));
```
greglittlefield-wf: Why isn't this run in for Component2?

lib/react_client.dart
greglittlefield-wf: Looks like we're good. Could we add an additional test case for passing a Dart List through this, to ensure it doesn't get minified?

You'll also want to use the `same` matcher to ensure it's the same reference:
```dart
arguments: [expectedProps, expectedState, same(snapshot)].
```

test/lifecycle_test.dart
joebingham-wk: I made this only for `Component` because `getUpdatingSetStateLifeCycleCalls` referenced `ReactSetStateTestComponent`, which at the moment extends from `Component`. Should I make another StateTestComponent for Component2 to perform the same test by with the newer lifecycle methods?

lib/react.dart
joebingham-wk: An example component in the example/test?

greglittlefield-wf: Yup, either there or in some file in a test directory, just to make sure this works as intended without any static typing errors

test/lifecycle_test.dart
```diff
@@ -166,12 +166,31 @@ main() {
...
+      });
+
+      test(
+          'Ensure getSnapshotBeforeId returned snapshot value is not '
+          'minified', () {
+        List<String> expectedSnapshot = [
+          'List Item 1',
+          'List Item 2',
+          'List '
+              'Item 3'
```
greglittlefield-wf: #nit this string shouldn't be split like this

test/lifecycle_test.dart
```diff
@@ -166,12 +166,31 @@ main() {
...
-                'sna'
-                    'pshot'
-              ]),
+              matchCall('componentDidUpdate', args: [anything, anything, 'snapshot']),
+            ]));
+      });
+
+      test(
+          'Ensure getSnapshotBeforeId returned snapshot value is not '
+          'minified', () {
```
greglittlefield-wf: Could we update this description to more clearly express what we're testing?

```dart
'does not jsify the return value of getSnapshotBeforeId, and '
'passes the same instance of that value to componentDidUpdate'
```

test/lifecycle_test.dart
greglittlefield-wf: Ah, gotcha. Yeah, I think we should have that, but I'd be fine splitting that out into a separate ticket as opposed to doing it in this PR. Whichever you prefer!

test/lifecycle_test.dart
```diff
@@ -145,11 +147,51 @@ main() {
...
+        expect(
+            component.lifecycleCalls,
+            containsAllInOrder([
+              matchCall('getSnapshotBeforeUpdate'),
+              matchCall('componentDidUpdate', args: [anything, anything, 'snapshot']),
+            ]));
+      });
+
+      test(
+          'does not jsify the return value of getSnapshotBeforeId, and '
```
kealjones-wk:
```suggestion
          'does not jsify the return value of getSnapshotBeforeUpdate, and '
```

test/lifecycle_test.dart
```diff
@@ -145,11 +147,51 @@ main() {
...
+
+        expect(
+            component.lifecycleCalls,
+            containsAllInOrder([
+              matchCall('getSnapshotBeforeUpdate'),
+              matchCall('componentDidUpdate', args: [anything, anything, 'snapshot']),
+            ]));
+      });
+
+      test(
```
kealjones-wk: I actually modified gregs jsBackedMaps type tests into a shared test in my pr that i use to check lots of types for context and in gregs jsBackedMaps tests that i think this would be beneficial to the use for testing the snapshot types here too.
Add this file: https://github.com/cleandart/react-dart/blob/4044e5c2e48224cd379d93c984665b78e149a179/test/shared_type_tester.dart

Then import it in this file:
`import 'shared_type_tester.dart';`

and the test would change to be like:
```
      group(
          'componentDidUpdate receives the same value created in getSnapshotBeforeUpdate',
          () {
        testSnapshotType(expectedSnapshot) {
          LifecycleTestHelper component = getDartComponent(
              render(components2.LifecycleTest({'getSnapshotBeforeUpdate': (_, __, ___) => expectedSnapshot})));

          component.lifecycleCalls.clear();
          component.setState({});

          expect(
            component.lifecycleCalls,
            containsAllInOrder([
              matchCall('getSnapshotBeforeUpdate'),
              matchCall('componentDidUpdate', args: [anything, anything, same(expectedSnapshot)])
            ]),
          );
        }

        sharedTypeTests(testSnapshotType);
      });
```

then if you do this you probably don't even need the test above.

test/ReactSetStateTestComponent.js
```diff
@@ -35,43 +35,43 @@ class ReactSetStateTestComponent extends React.Component {
     _counter = newCount;
   }
 
-  recordLifecyleCall(name) {
+  recordLifecycleCall(name) {
```
kealjones-wk: lol, wow nice fix.

test/lifecycle_test.dart
```diff
@@ -629,7 +716,9 @@ void sharedLifecycleTests<T extends react.Component>({
         react_dom.render(LifecycleTest(newProps), mountNode);
 
         List calls = [
-          matchCall('componentWillReceiveProps', args: [newPropsWithDefaults], props: initialPropsWithDefaults),
+          isComponent2
+              ? null
```
greglittlefield-wf: Why isn't this being tested for `Component2`?

test/lifecycle_test.dart
```diff
@@ -145,11 +147,51 @@ main() {
...
+            render(components2.LifecycleTest({'getSnapshotBeforeUpdate': (_, __, ___) => expectedSnapshot})));
+
+        component.lifecycleCalls.clear();
+        component.setState({});
+
+        expect(
+            component.lifecycleCalls,
+            containsAllInOrder([
+              matchCall('getSnapshotBeforeUpdate'),
+              matchCall('componentDidUpdate', args: [anything, anything, same(expectedSnapshot)]),
```
kealjones-wk: I dont think the `same` does anything here... I know @greglittlefield-wf said https://github.com/cleandart/react-dart/pull/169/files#r266086539 but @greglittlefield-wf doesnt the `containsAllInOrder` check this?

test/lifecycle_test.dart
```diff
@@ -368,6 +420,7 @@ void sharedLifecycleTests<T extends react.Component>({
 
       const Map expectedState = const {};
       const Map expectedContext = const {};
+      const Null expectedSnapshot = null;
```
kealjones-wk: well these are pretty hardcore for a null check lol @greglittlefield-wf is this needed or could he just do: `var expectedSnapshot;`

test/lifecycle_test.dart
greglittlefield-wf: This test ensures that the same instance is passed back, which we need `same` for.

test/lifecycle_test.dart
joebingham-wk: I believe the thought process was that `componentWillReceiveProps` is no longer getting called because of getSnapshotBeforeUpdate, so just exclude it from the test. Is that logic sound or is there a test that would be good to put there in its place?

test/lifecycle_test.dart
greglittlefield-wf: This looks good to me; it should definitely be a `const` as opposed to a var

test/lifecycle_test.dart
joebingham-wk: I believe the thought process was that componentWillReceiveProps is no longer getting called because of getSnapshotBeforeUpdate, so just exclude it from the test. Is that logic sound or is there a test that would be good to put there in its place?

test/lifecycle_test.dart
kealjones-wk: You can ignore this, ill add it to my pr

test/ReactSetStateTestComponent.js
joebingham-wk: Haha thank you! I was looking for typos in my own code and ended up discovering that one.

test/lifecycle_test.dart
kealjones-wk: `componentWillReceiveProps` was removed from `Component2`: https://github.com/cleandart/react-dart/pull/169/files#diff-ec47402eaffd4414fd19bddea36f05d0L96

kealjones-wk:
+10

lib/react.dart
```diff
@@ -775,6 +843,43 @@ abstract class Component2 implements Component {
   List<StateUpdaterCallback> get transactionalSetStateCallbacks => throw new UnimplementedError();
 }
 
+/// Mixin that enforces consistent typing of the `snapshot` parameter
+/// returned by [Component2.getSnapshotBeforeUpdate]
+/// and passed into [Component2.componentDidUpdate].
+///
+/// __EXAMPLE__:
+///
+/// class _ParentComponent extends react.Component2
```
aaronlademann-wf: Examples will only render as a `<pre>` (code block) if it indented 4 spaces.

lib/react.dart
```diff
@@ -775,6 +843,43 @@ abstract class Component2 implements Component {
...
+///   void componentDidUpdate(prevProps, prevState, [bool snapshot]) {
+///     if (snapshot == true) {
+///       // Do something
+///     }
+///   }
+///
+///   /* Include other standard component logic */
+///
+/// }
+abstract class TypedSnapshot<TSnapshot> implements Component2 {
```
aaronlademann-wf: This is a perfect situation to utilize Dart 2's `mixin`s.  It will allow use to keep the methods abstract, and is cleaner / less confusing than `implements`, IMO.

```dart
mixin TypedSnapshot<TSnapshot> on Component2 {
  @override
  TSnapshot getSnapshotBeforeUpdate(Map prevProps, Map prevState);

  @override
  void componentDidUpdate(Map prevProps, Map prevState, [covariant TSnapshot snapshot]);
}
```

aaronlademann-wf: Also - note the use of the `@override` annotation.  Its a good idea to get in the habit of annotating overridden members.

aaronlademann-wf:
QA +1

PR #170, opened 2019-03-18
Reformat with line length of 120

greglittlefield-wf:
> Mirror of: https://github.com/cleandart/react-dart/pull/168

80 was too short for a lot of the code that has already been written, causing excessive wrapping and code that's hard to read.

While we'd like to adhere to the recommended line length of 80, we're going to go with 120 to preserve our sanity.


aaronlademann-wf:
+1

PR #174, opened 2019-03-26
CPLAT-4765 Implement getDerivedStateFromProps lifecycle method

tainhenning-wk:
Description:
Deprecated componentWillReceiveProps for the new getDerivedStateFromProps introduced in react 16
Testing:
- [ ] CI Passes
- [ ] Try using getDerivedStateFromProps

**IMPORTANT**
For a clean diff check, use this:
https://github.com/cleandart/react-dart/pull/174/commits/4c1c62bf3b6277b9c70c86e740132fde9a5639cb this will have all changes that I did


lib/react.dart
```diff
@@ -565,12 +571,39 @@ abstract class Component2 implements Component {
...
+  /// ReactJS lifecycle method that is invoked before rendering on the intial mount and on subsequent updates
+  /// It should return an object to update the state, or null to update nothing (this is different from componentWillRecieveProps).
   ///
-  /// __Note__: Choose either this method or [componentWillReceivePropsWithContext]. They are both called at the same
-  /// time so using both provides no added benefit.
+  /// [prevState] will be null when this lifcecyle method is called before first mount.
   ///
-  /// See: <https://facebook.github.io/react/docs/react-component.html#updating-componentwillreceiveprops>
-  void componentWillReceiveProps(Map newProps) {}
+  /// This method is also static, so features like [this] or [ref] cannot be used in this method
```
greglittlefield-wf: #nit This wording could use a little tweaking. We also shouldn't mention ref, since it's deprecated.
```
 /// This method is also effectively static, so using instance members like [props] or ref variables will not work.
```

lib/react.dart
```diff
@@ -565,12 +571,39 @@ abstract class Component2 implements Component {
...
+  ///
+  /// See: <https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops>
+  ///
+  /// __Example__:
+  ///
+  ///    getDerivedStateFromProps(nextProps, prevState) {
```
greglittlefield-wf: This should have type annotations and an `@override`. Looks like the body is also still in JS, and needs to be converted to Dart.

lib/react.dart
```diff
@@ -565,12 +571,39 @@ abstract class Component2 implements Component {
...
   ///
-  /// __Note__: Choose either this method or [componentWillReceivePropsWithContext]. They are both called at the same
-  /// time so using both provides no added benefit.
+  /// [prevState] will be null when this lifcecyle method is called before first mount.
   ///
-  /// See: <https://facebook.github.io/react/docs/react-component.html#updating-componentwillreceiveprops>
-  void componentWillReceiveProps(Map newProps) {}
+  /// This method is also static, so features like [this] or [ref] cannot be used in this method
+  ///
+  /// See: <https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops>
```
greglittlefield-wf: I think it'd be a good idea to include the info under the "Deriving state leads to verbose code and makes your components difficult to think about." section of these docs directly within this doc comment.

lib/react_client/react_interop.dart
```diff
@@ -352,9 +368,19 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
     dynamic Function(Component2 component) handleGetInitialState,
     void Function(Component2 component, ReactComponent jsThis) handleComponentWillMount,
     void Function(Component2 component) handleComponentDidMount,
-    // TODO: Should this be removed when we update Component2.componentWillReceiveProps to throw an UnsupportedError? (CPLAT-4765)
-    void Function(Component2 component, JsMap jsNextProps) handleComponentWillReceiveProps,
-    bool Function(Component2 component, JsMap jsNextProps, JsMap jsNextState) handleShouldComponentUpdate,
+    void Function(
+      ComponentStatics componentInstance,
```
greglittlefield-wf: This name isn't right, and this should supply a generic parameter
```suggestion
      ComponentStatics<Component2> componentStatics,
```

test/lifecycle_test.dart
```diff
@@ -760,14 +769,18 @@ void sharedLifecycleTests<T extends react.Component>({
...
-          isComponent2
-              ? matchCall('shouldComponentUpdate',
-                  args: [expectedProps, newState, expectedContext], state: initialState)
-              : matchCall('shouldComponentUpdateWithContext',
-                  args: [expectedProps, newState, expectedContext], state: initialState),
-        ];
+        List calls;
+        if (isComponent2) {
+          calls = [
+            matchCall('getDerivedStateFromProps'),
```
greglittlefield-wf: Can we add `args` expectations to this?

lib/react.js
```diff
@@ -2382,7 +2382,7 @@ module.exports = ReactPropTypesSecret;
 /***/ (function(module, exports, __webpack_require__) {
 
 "use strict";
-/** @license React v16.8.6
+/** @license React v16.8.3
```
aaronlademann-wf: This should not be getting downgraded. It's possible you'll need to blow away your `package-lock.json` file and re-run `npm install && npm build`.

cc/ @greglittlefield-wf

lib/react.dart
```diff
@@ -614,11 +620,44 @@ abstract class Component2 implements Component {
...
-  ///
-  /// See: <https://facebook.github.io/react/docs/react-component.html#updating-componentwillreceiveprops>
+  /// > __DEPRECATED - DO NOT USE__
+  /// >
+  /// > This will be removed once 6.0.0 releases, switching to [Component2.getDerivedStateFromProps] is the path forward
+  /// >
+  /// > React 16 has deprecated [componentWillReceiveProps] in favor of [getDerivedStateFromProps]
+  /// >
+  /// > This will be completely removed when the JS side of it is slated for removal (ReactJS 17 / react.dart 6.0.0)
+  @mustCallSuper
```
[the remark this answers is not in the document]
tainhenning-wk: This forces the dev to call super, and when they call super they will be able to see the deprecation error message.

lib/react_client/react_interop.dart
```diff
@@ -424,6 +424,8 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
 class ComponentStatics<T extends Component> {
   final ComponentFactory<T> componentFactory;
   final Component2 componentInstance;
+
+  /// Required for use by [ReactDartInteropStatics2].
```
aaronlademann-wf: NOTE: This will need to be updated / removed when @joebingham-wk's error boundary PR goes through, since he's moving all the `Component2`-related stuff into its own `ComponentStatics2`.

test/lifecycle_test.dart
```diff
@@ -175,6 +177,52 @@ main() {
...
+
+          component.lifecycleCalls.clear();
+
+          (component as react.Component2).forceUpdate();
+
+          expect(
+              component.lifecycleCalls,
+              equals([
+                matchCall('getDerivedStateFromProps'),
+                matchCall('render', state: {}),
```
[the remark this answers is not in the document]
aaronlademann-wf: Can we also add an arbitrary state key/value pair to this component so that we can assert that returning null from getDerivedStateFromProps does not result in the unexpected mutation of state -- that the pre-existing key/value pairs are not obliterated?

test/lifecycle_test.dart
```diff
@@ -545,6 +606,7 @@ void sharedLifecycleTests<T extends react.Component>({
       expect(
           component.lifecycleCalls,
           equals([
+            isComponent2 ? matchCall('getDerivedStateFromProps') : null,
```
greglittlefield-wf: This should have `args: ` expectations to test that the correct arguments are being passed in

test/lifecycle_test.dart
```diff
@@ -607,6 +669,7 @@ void sharedLifecycleTests<T extends react.Component>({
...
```
greglittlefield-wf: This should have `args: ` expectations to test that the correct arguments are being passed in

greglittlefield-wf: Same goes for most of these; may as well add it to all of them, for consistency

kealjones-wk:
last commit should be `merge in 5.0.0-wip`

lib/react_client.dart
```diff
@@ -623,6 +624,7 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
   return new ReactDartInteropStatics2(
     initComponent: allowInterop(initComponent),
     handleGetInitialState: allowInterop(handleGetInitialState),
+    // TODO: we should review if we need to support the deprecated will methods in component2
```
aaronlademann-wf: Definitely... I think we should revisit the entire impl via CPLAT-2277.

test/lifecycle_test.dart
```diff
@@ -184,18 +184,48 @@ main() {
             'initialDerivedState': 'initial',
           };
 
-          final Map initialProps = unmodifiableMap({'getDerivedStateFromProps': (_, __, ___) => initialDerivedState});
+          const Map updatedDerivedState = const {
+            'initialDerivedState': 'updated',
+          };
+
+          final Map initialProps = {'getDerivedStateFromProps': (_, __, ___) => initialDerivedState};
```
greglittlefield-wf: Could we make this and unmodifiable maps like they used to be? That reduces the risk of these maps changing unexpectedly and causing the wrong values to be checked in the expectations.

test/lifecycle_test.dart
```diff
@@ -176,6 +177,85 @@ main() {
...
+                matchCall('componentDidUpdate', state: updatedDerivedState),
+              ].where((matcher) => matcher != null).toList()));
+        });
+
+        test('null, leaving the existing instance state value intact', () {
+          const Map initialState = const {
+            'foo': 'sudo',
+          };
+          final Map initialProps = unmodifiableMap(
+              {'getDerivedStateFromProps': (_, __, ___) => null, 'getInitialState': (_) => initialState});
```
greglittlefield-wf: #nit trailing commas would help with formatting here
```suggestion
              {'getDerivedStateFromProps': (_, __, ___) => null, 'getInitialState': (_) => initialState,});
```

test/ReactSetStateTestComponent2.js
```diff
@@ -71,8 +71,15 @@ class ReactSetStateTestComponent2 extends React.Component {
         this.recordStateChange(this.state.counter);
     }
 
+    static getDerivedStateFromProps(nextProps, __) {
+        _shouldUpdate = nextProps.shouldUpdate;
+        _shouldUpdate ? _component2UpdatingSetStateLifeCycleCalls.push("getDerivedStateFromProps") : _component2NonUpdatingSetStateLifeCycleCalls.push("getDerivedStateFromProps");
```
greglittlefield-wf: #nit this line is pretty long, could get wrapped
```suggestion
        _shouldUpdate
            ? _component2UpdatingSetStateLifeCycleCalls.push("getDerivedStateFromProps") 
            : _component2NonUpdatingSetStateLifeCycleCalls.push("getDerivedStateFromProps");
```

test/lifecycle_test/component2.dart
```diff
@@ -29,8 +29,9 @@ class _SetStateTest extends react.Component2 with LifecycleTestHelper {
       };
 
   @override
-  componentWillReceiveProps(_) {
-    recordLifecyleCall('componentWillReceiveProps');
+  Map getDerivedStateFromProps(_, __) {
+    recordLifecyleCall('getDerivedStateFromProps');
+    return new Map();
```
greglittlefield-wf: #nit prefer collection literals
```suggestion
return {};
```

greglittlefield-wf:
+10

tainhenning-wk:
+1 a040c83ffe9b73c6a0859f16b76f9d07d3e1996f

PR #175, opened 2019-03-29
CPLAT-3094 Lifecycle Error Boundary

joebingham-wk:
## Description
In order to allow for the possibility of React 16 error boundaries, we needed to add `getDerivedStateFromError` and `componentDidCatch` lifecycle events. However, we also needed to ensure that not every component becomes and error boundary. Consequently, `skipMethods` was introduced to `registerComponent`. 

## Changes
- Add the lifecycle event for `componentDidCatch`.
- Add the lifecycle event for `getDerivedStateFromError`.
- Modify tests to allow for static lifecycle events.
- Add increased logic for `skipMethods`.
- Add `_filterSkipMethods` to ensure that a consumer cannot skip important lifecycle events.
- Add logic in `_dart_helpers.js` removing lifecycle events based upon the final skipMethods list.
- Add test coverage.
- Add an example in `react_test.dart`

## Testing Suggestions
- [ ] CI Passes
- [ ] Manually check the new example to ensure that it behaves as expected.
- [ ] Verify adequate unit test coverage.


kealjones-wk:
As an update with where this currently stands... This implementation is working but the `error` that we get in `getDerivedStateFromError` is a JsObject or NativeJavaScriptObject... See issue on dart-lang/sdk here: https://github.com/dart-lang/sdk/issues/36363

because this is branched from my context pr that has not been merged yet a cleaner diff cant be found here: https://github.com/kealjones-wk/react-dart/compare/react-16-new-context...cleandart:CPLAT-3094-lifecycle-error-boundary

lib/react.dart
```diff
@@ -1461,6 +1491,16 @@ class SyntheticWheelEvent extends SyntheticEvent {
   throw new Exception('setClientConfiguration must be called before registerComponent.');
 };
 
+/// Registers [componentFactory] on both client and server.
+/*ComponentRegistrar*/ Function registerComponent2 = (/*ComponentFactory*/ componentFactory,
+    [/*Iterable<String>*/ skipMethods = const [
```
aaronlademann-wf: Is there a reason that the "version 2" of this can't have all these types? 

I know we had them commented out in "version 1"... but... I feel like we might be able to have typing on this stuff now.

lib/react_client.dart
```diff
@@ -209,6 +209,44 @@ dynamic _convertArgsToChildren(List childrenArgs) {
...
 
+/// Util used with [_registerComponent2] to ensure no imporant lifecycle
+/// methods are skipped. This includes [shouldComponentUpdate],
+/// [componentDidUpdate], and [render] because they utilize
+/// [_updatePropsAndStateWithJs].
+///
+/// Returns the life of lifecycle methods to skip, having removed the
+/// important ones. If an important lifecycle event was set for skipping, a
+/// warning is issued.
+
```
aaronlademann-wf: Unnecessary blank line

lib/react_client.dart
```diff
@@ -209,6 +209,44 @@ dynamic _convertArgsToChildren(List childrenArgs) {
...
+  }
+
+  if (finalList.contains('render')) {
+    finalList.remove('render');
+    warn = true;
+  }
+
+  if (warn) {
+    window.console.warn("WARNING: Crucial lifecycle methods passed into "
+        "skipMethod. shouldComponentUpdate, componentDidUpdate, and render "
```
```suggestion
        "skipMethods. shouldComponentUpdate, componentDidUpdate, and render "
```

lib/react_client.dart
```diff
@@ -708,9 +756,14 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
 /// Creates and returns a new [ReactDartComponentFactoryProxy] from the provided [componentFactory]
 /// which produces a new JS [`ReactClass` component class](https://facebook.github.io/react/docs/top-level-api.html#react.createclass).
 ReactDartComponentFactoryProxy2 _registerComponent2(ComponentFactory<Component2> componentFactory,
-    [Iterable<String> skipMethods = const []]) {
+    [Iterable<String> skipMethods = const [
+      'getDerivedStateFromError',
+      'compo'
+          'nentDidCatch'
```
aaronlademann-wf: eh?

lib/react_client/react_interop.dart
```diff
@@ -411,6 +411,8 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
     )
         handleComponentDidUpdate,
     void Function(Component2 component) handleComponentWillUnmount,
+    void Function(Component2 component, dynamic error, dynamic info) handleComponentDidCatch,
+    void Function(ComponentStatics componentInstance, dynamic error) handleGetDerivedStateFromError,
```
aaronlademann-wf: Is this `void`? I thought it was `JsMap`?

lib/react_client/react_interop.dart
```diff
@@ -423,8 +425,8 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
 /// See [ReactDartInteropStatics], [createReactDartComponentClass].
 class ComponentStatics<T extends Component> {
   final ComponentFactory<T> componentFactory;
-
-  ComponentStatics(this.componentFactory);
+  final Component2 componentInstance;
+  ComponentStatics(this.componentFactory, {this.componentInstance});
```
aaronlademann-wf: What is the reason that a `componentInstance` must be stored in here now?

lib/react_client/react_interop.dart
```diff
@@ -423,8 +425,8 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
 /// See [ReactDartInteropStatics], [createReactDartComponentClass].
 class ComponentStatics<T extends Component> {
   final ComponentFactory<T> componentFactory;
-
-  ComponentStatics(this.componentFactory);
+  final Component2 componentInstance;
```
aaronlademann-wf: Also... if this is necessary - shouldn't it be 

```suggestion
  final T componentInstance;
```

test/ReactSetStateTestComponent2.js
```diff
@@ -79,6 +79,14 @@ class ReactSetStateTestComponent2 extends React.Component {
         this.setState(this.innerTransactionalSetStateCallback.bind(this), this.innerSetStateCallback.bind(this));
     }
 
+    componentDidCatch(_, __) {
+        this.recordLifecycleCall('componentDidCatch');
+    }
+
+    getDerivedStateFromError(_) {
+        this.recordLifecycleCall('getDerivedStateFromError');
```
aaronlademann-wf: Should we return a map from this to test / assert that the state is updated with the returned value?

test/ReactSetStateTestComponent2.js
```diff
@@ -79,6 +79,14 @@ class ReactSetStateTestComponent2 extends React.Component {
         this.setState(this.innerTransactionalSetStateCallback.bind(this), this.innerSetStateCallback.bind(this));
     }
 
+    componentDidCatch(_, __) {
+        this.recordLifecycleCall('componentDidCatch');
```
aaronlademann-wf: Should we record the value of the two arguments here to verify they are being passed along as expected?

test/lifecycle_test.dart
```diff
@@ -346,6 +358,7 @@ void sharedLifecycleTests<T extends react.Component>({
         expect(
             component.lifecycleCalls,
             equals([
+              matchCall('getChildContext'),
```
aaronlademann-wf: Is this left over from a bad merge with Keal's context PR?

test/lifecycle_test.dart
```diff
@@ -362,6 +375,7 @@ void sharedLifecycleTests<T extends react.Component>({
...
```
aaronlademann-wf: Is this left over from a bad merge with Keal's context PR?

test/lifecycle_test/component2.dart
```diff
@@ -133,6 +175,19 @@ class _LifecycleTestWithContext extends _LifecycleTest {
   get contextType => LifecycleTestContext;
 }
 
+ReactDartComponentFactoryProxy2 ErrorComponent = react.registerComponent(() => new _ErrorComponent());
```
aaronlademann-wf: Should this be `registerComponent2`?

lib/react.dart
```diff
@@ -714,6 +714,36 @@ abstract class Component2 implements Component {
...
+  void componentDidCatch(dynamic error, dynamic info) {}
+
+  /// ReactJS lifecycle method that is invoked after an error is thrown by a
+  /// descendant.
+  ///
+  /// Use this method to capture the error and update component state after an
+  /// error is thrown.
+  ///
+  /// __Note__: This method, along with [componentDidCatch] will only
+  /// be called if `skipMethods` in [_registerComponent] is overridden with
```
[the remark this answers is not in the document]
joebingham-wk: Yep, that's what it should have been. I was meaning to reference `_registerComponent` in `react_client.dart` because the private methods in `react_client` are ultimately what get used, but it should definitely be `registerComponent2`.

lib/react.dart
```diff
@@ -714,6 +714,36 @@ abstract class Component2 implements Component {
...
   void componentWillUnmount() {}
 
+  /// ReactJS lifecycle method that is invoked after an error is thrown by a
+  /// descendant.
+  ///
+  /// Use this method primarily for logging errors, but because it takes
+  /// place after the commit phase side-effects are permitted.
+  ///
+  /// __Note__: This method, along with [getDerivedStateFromError] will only
+  /// be called if `skipMethods` in [_registerComponent] is overridden with
```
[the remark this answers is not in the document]
joebingham-wk: Yep, that's what it should have been. I was meaning to reference `_registerComponent` in `react_client.dart` because the private methods in `react_client` are ultimately what get used, but it should definitely be `registerComponent2`.

lib/react.dart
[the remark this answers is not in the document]
joebingham-wk: Yep, same thing as above!

lib/react_client/react_interop.dart
joebingham-wk: Good catch, it was meant to be `JsMap`.

test/lifecycle_test/component2.dart
joebingham-wk: Currently `setClientConfiguration` is set to use `registerComponent` rather than `registerComponent2`. As a result, using `registerComponent2` throws the `setClientConfiguration must be called before registerComponent.` exception. That makes me think it would take a notable workaround to get `registerComponent2` to work, whereas `registerComponent` already has a short circuit that redirects to `registerComponent2` if the component is of the type `Component2`. Did I miss something and it would actually be straightforward or is this ok for now?

test/lifecycle_test.dart
kealjones-wk: It's probably left over from when he branched off my in progress context pr...
https://github.com/cleandart/react-dart/pull/160/files#diff-c00a02aa016b25a86fb30a2093f935a1R349

joebingham-wk: Just to recap standup for future reference, it appears they're needed. The lifecycle calls are part of the actual test results, and it makes sense (to me) for them to be there based on the tests. If anyone wants to double check that, it may be a good QA check, but `getChildContext` is unrelated to `Component2` so it seems more likely that the overlap in branching created this need as opposed to adding new lifecycles to `Component2`.

test/lifecycle_test.dart
joebingham-wk: Same note as above!

lib/react_client/react_interop.dart
joebingham-wk: Because both `Component` and `Component2` rely on `ComponentStatics`, this would work but ends up requiring explicitly typing instances of `ComponentStatics` for the sake of differentiating the usages between `Component` and `Component2`. @greglittlefield-wf suggested just implementing `ComponentStatics2` because then this works easily and removes the need for `ComponentStatics` to have `componentInstance`.

test/ReactSetStateTestComponent2.js
joebingham-wk: I definitely should have done that. To fix it, I ended up mirroring the `component2.dart` functionality that relies on `getDerivedStateFromError` returning a state update correctly in order to render properly. I feel like this and the error boundary test in the `Component2` group of `lifecycle_test.dart` show that the state is updating correctly, but let me know if I misunderstood or there's a different way to approach testing that would add more coverage!

lib/react_client/react_interop.dart
joebingham-wk: Will be addressed in a doc comment.

lib/react_client.dart
```diff
@@ -564,6 +601,14 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
         component.componentWillUnmount();
       });
 
+  void handleComponentDidCatch(Component2 component, dynamic error, dynamic info) => zone.run(() {
+        component.componentDidCatch(error, info);
+      });
+
+  JsMap handleGetDerivedStateFromError(ComponentStatics2 componentStatics, dynamic error) => zone.run(() {
+        return jsBackingMapOrJsCopy(componentStatics.componentInstance.getDerivedStateFromError(error));
```
aaronlademann-wf: Why does this one access the `componentInstance` from `componentStatics`, but `handleComponentDidCatch` passes the `component` directly? Seems inconsistent.  Lets pick one unless there is a reason I'm not seeing for why it has to be different.

lib/react_client/react_interop.dart
```diff
@@ -320,8 +320,8 @@ external ReactClass createReactDartComponentClass(
 /// the JS and Dart component instances.
 @JS('_createReactDartComponentClass2')
 external ReactClass createReactDartComponentClass2(
-    ReactDartInteropStatics2 dartInteropStatics, ComponentStatics<Component2> componentStatics,
-    [JsComponentConfig2 jsConfig]);
+    ReactDartInteropStatics2 dartInteropStatics, ComponentStatics2<Component2> componentStatics,
```
aaronlademann-wf: Shouldn't be any reason we need to provide a generic here since we're making a "v2" version, right?

lib/react_client/react_interop.dart
```diff
@@ -423,10 +425,26 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
...
+/// This object is made accessible to a component's JS ReactClass config, which
+/// passes it to certain methods in [ReactDartInteropStatics2].
+///
+/// [ComponentStatics2] exists in addition to [ComponentStatics] because
+/// [Component2] brings new static lifecycle methods that are exposted via
+/// `componentInstace`. Because [Component] does not need an instance to pull
+/// static lifecycle methods from, it made sense to create [ComponentStatics2].
+///
+/// See [ReactDartInteropStatics2], [createReactDartComponentClass2].
+class ComponentStatics2<T extends Component2> {
```
aaronlademann-wf: No need for the generics here I don't think.

lib/react_client.dart
joebingham-wk: It is different because `getDerivedStateFromError` is static and needs access directly to the instance, whereas the other lifecycle methods don't. The property name wasn't very descriptive to that though, so I updated the name.

lib/react_client/react_interop.dart
joebingham-wk: I believe it may be necessary here because `ComponentFactory` is defined with the `T` generic. Is there a workaround to that?

aaronlademann-wf: @joebingham-wk you would just do this instead:

```dart
class ComponentStatics2 {
  final ComponentFactory<Component2> componentFactory;
  final Component2 instanceForStaticMethods;
  ComponentStatics2(this.componentFactory, {this.instanceForStaticMethods});
}
```

joebingham-wk: Well, that was simple. I should have seen that - I'm sorry for making you type it! Code is updated with that change.

lib/react_client.dart
```diff
@@ -260,7 +297,7 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
 
   /// Wrapper for [Component.getInitialState].
   Component initComponent(ReactComponent jsThis, ReactDartComponentInternal internal, InteropContextValue context,
-          ComponentStatics componentStatics) =>
+          ComponentStatics<Component> componentStatics) =>
```
aaronlademann-wf: Sorry... I missed this before... since we're only going to be using `ComponentStatics` for `Component`, not `Component` _and_ `Component2`, there is no need for the `ComponentStatics` class to have generic parameters.

lib/react_client.dart
```diff
@@ -587,7 +634,7 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
 /// which produces a new JS [`ReactClass` component class](https://facebook.github.io/react/docs/top-level-api.html#react.createclass).
 @Deprecated('6.0.0')
 ReactDartComponentFactoryProxy _registerComponent(ComponentFactory componentFactory,
-    [Iterable<String> skipMethods = const []]) {
+    [Iterable<String> skipMethods = const ['getDerivedStateFromError', 'componentDidCatch']]) {
```
aaronlademann-wf: These methods are not available in a V1 component are they? Also, without a call to `_filterSkipMethods`, I'm not sure what the point of this change is?

lib/react_client/react_interop.dart
```diff
@@ -411,6 +411,8 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
     )
         handleComponentDidUpdate,
     void Function(Component2 component) handleComponentWillUnmount,
+    void Function(Component2 component, dynamic error, dynamic info) handleComponentDidCatch,
+    JsMap Function(ComponentStatics2 componentInstance, dynamic error) handleGetDerivedStateFromError,
```
aaronlademann-wf: `componentInstance` seems like a misleading name since the instance is actually a field within the `ComponentStatics2` instance being passed in.

lib/react_client/react_interop.dart
```diff
@@ -423,10 +425,26 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
...
 }
 
+/// An object that stores static methods and information for a specific component class.
+///
+/// This object is made accessible to a component's JS ReactClass config, which
+/// passes it to certain methods in [ReactDartInteropStatics2].
+///
+/// [ComponentStatics2] exists in addition to [ComponentStatics] because
+/// [Component2] brings new static lifecycle methods that are exposted via
+/// `componentInstace`. Because [Component] does not need an instance to pull
```
aaronlademann-wf:
```suggestion
/// `instanceForStaticMethods`. Because [Component] does not need an instance to pull
```

test/lifecycle_test/component2.dart
```diff
@@ -7,14 +7,21 @@ import 'package:react/react_client.dart';
 
 import 'util.dart';
 
-ReactDartComponentFactoryProxy2 SetStateTest = react.registerComponent(() => new _SetStateTest());
+ReactDartComponentFactoryProxy2 SetStateTest = react.registerComponent(() => new _SetStateTest(), [null]);
```
aaronlademann-wf: Do we have any tests that assert the default value of `skipMethods`?

lib/react_client.dart
joebingham-wk: I had originally included it because `_registerComponent` was allowing for `Component2` instances to go through it and was just forwarding those cases to `_registerComponent2`. So you're definitely right that for V1 it's not necessary, but if someone does use the method for `Component2` then having the defaults set here makes sure they don't accidentally set an error boundary (as an empty List default would have done). If I took `['getDerivedStateFromError', 'componentDidCatch']` away I could just not set a default value for `skipMethods` in `_registerComponent` at all so that `_registComponent2`'s defaults would stick. Would that be better?

js_src/_dart_helpers.js
```diff
@@ -103,13 +103,26 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
...
     render() {
       var result = dartInteropStatics.handleRender(this.dartComponent, this.props, this.state, this.context);
       if (typeof result === 'undefined') result = null;
       return result;
     }
   }
 
+  // Delete methods that the user does not want to include (such as error boundary event).
+  skipMethods.forEach((method) => {
+    if (ReactDartComponent2[method] !== undefined) {
```
greglittlefield-wf: This check is unnecessary; `delete` is a noop for properties that do not exist

lib/react_client.dart
```diff
@@ -209,6 +209,43 @@ dynamic _convertArgsToChildren(List childrenArgs) {
...
+  }
+
+  if (finalList.contains('render')) {
+    finalList.remove('render');
+    shouldWarn = true;
+  }
+
+  if (shouldWarn) {
+    window.console.warn("WARNING: Crucial lifecycle methods passed into "
+        "skipMethods. shouldComponentUpdate, componentDidUpdate, and render "
```
greglittlefield-wf: NICE.

lib/react_client.dart
```diff
@@ -723,8 +771,9 @@ ReactDartComponentFactoryProxy2 _registerComponent2(ComponentFactory<Component2>
 
   /// Create the JS [`ReactClass` component class](https://facebook.github.io/react/docs/top-level-api.html#react.createclass)
   /// with custom JS lifecycle methods.
-  var reactComponentClass = createReactDartComponentClass2(_dartInteropStatics2, componentStatics, jsConfig2)
-    ..displayName = componentInstance.displayName;
+  var reactComponentClass =
+      createReactDartComponentClass2(_dartInteropStatics2, componentStatics, jsConfig2, filteredSkipMethods)
```
greglittlefield-wf: #nit could we add `filteredSkipMethods` to the `JsComponentConfig2` object instead? It would be nice to minimize the number of positional arguments

lib/react_client/react_interop.dart
```diff
@@ -423,10 +425,26 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
 /// See [ReactDartInteropStatics], [createReactDartComponentClass].
 class ComponentStatics<T extends Component> {
```
greglittlefield-wf: @aaronlademann-wf mentioned this above, but can we remove this generic parameter now that this class will never be used by `Component2`?

lib/react_client/react_interop.dart
```diff
@@ -423,10 +425,26 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
...
   ComponentStatics(this.componentFactory);
 }
 
+/// An object that stores static methods and information for a specific component class.
+///
+/// This object is made accessible to a component's JS ReactClass config, which
+/// passes it to certain methods in [ReactDartInteropStatics2].
+///
+/// [ComponentStatics2] exists in addition to [ComponentStatics] because
+/// [Component2] brings new static lifecycle methods that are exposted via
```
greglittlefield-wf: Typo: "exposted"

Although, #nit I think this paragraph can just be removed; one can deduce from the absence of `instanceForStaticMethods` in `ComponentStatics` that this is the case

test/lifecycle_test.dart
```diff
@@ -175,6 +175,29 @@ main() {
...
       });
+
+      test('triggers error lifecycle events when an error is thrown', () {
+        var mountNode = new DivElement();
+        var renderedInstance = react_dom.render(components2.SetStateTest({}), mountNode);
+        LifecycleTestHelper component = getDartComponent(renderedInstance);
+        LifecycleTestHelper.staticLifecycleCalls.clear();
+        component.setState({"shouldThrow": true});
+
+        expect(component.lifecycleCalls, containsAllInOrder(['getDerivedStateFromError', 'componentDidCatch']));
```
greglittlefield-wf: Is there a reason `containsAllInOrder` is used here? I think `contains` would give us more test coverage, and assert that methods aren't called when we don't expect them to.

test/lifecycle_test.dart
```diff
@@ -362,6 +393,7 @@ void sharedLifecycleTests<T extends react.Component>({
         expect(
             component.lifecycleCalls,
             equals([
+              matchCall('getChildContext', props: anything, context: anything),
```
greglittlefield-wf: What are these context-related changes from?

test/lifecycle_test/component2.dart
```diff
@@ -152,6 +207,10 @@ class _LifecycleTest extends react.Component2 with LifecycleTestHelper {
   void componentDidUpdate(prevProps, prevState, [snapshot]) =>
       lifecycleCall('componentDidUpdate', arguments: [new Map.from(prevProps), new Map.from(prevState), snapshot]);
 
+  void componentDidCatch(error, info) => lifecycleCall('componentDidCatch', arguments: [error, new Map.from(info)]);
+
+  Map getDerivedStateFromError(error) => lifecycleCall('getDerivedStateFromError', arguments: [error]);
```
greglittlefield-wf: Are these lifecycle calls and arguments being verified anywhere? I'm not seeing tests for them

test/lifecycle_test.dart
```diff
@@ -175,6 +175,29 @@ main() {
...
+
+      test('triggers error lifecycle events when an error is thrown', () {
+        var mountNode = new DivElement();
+        var renderedInstance = react_dom.render(components2.SetStateTest({}), mountNode);
+        LifecycleTestHelper component = getDartComponent(renderedInstance);
+        LifecycleTestHelper.staticLifecycleCalls.clear();
+        component.setState({"shouldThrow": true});
+
+        expect(component.lifecycleCalls, containsAllInOrder(['getDerivedStateFromError', 'componentDidCatch']));
+        expect(component.state["shouldThrow"], isFalse);
```
greglittlefield-wf: This would be a good chance to use `reason`, providing more detail on why this is being asserted
```suggestion
        expect(component.state["shouldThrow"], isFalse, 
            reason: 'applies the state returned by `getDerivedStateFromError`');
```

test/lifecycle_test.dart
```diff
@@ -175,6 +175,29 @@ main() {
...
+      });
+
+      test('shows the correct error and info when an error is thrown', () {
```
greglittlefield-wf: #nit could we change this test description to better reflect what we're actually testing, and not the implementation details of the test harness? Something like:
```suggestion
      test('passes the correct error/info to lifecycle methods when an error is thrown', () {
```

test/lifecycle_test/component2.dart
```diff
@@ -152,6 +207,10 @@ class _LifecycleTest extends react.Component2 with LifecycleTestHelper {
   void componentDidUpdate(prevProps, prevState, [snapshot]) =>
       lifecycleCall('componentDidUpdate', arguments: [new Map.from(prevProps), new Map.from(prevState), snapshot]);
 
+  void componentDidCatch(error, info) => lifecycleCall('componentDidCatch', arguments: [error, new Map.from(info)]);
```
greglittlefield-wf: > `new Map.from(info)`
Does this even work?

lib/react.dart
```diff
@@ -714,6 +714,36 @@ abstract class Component2 implements Component {
...
+  /// place after the commit phase side-effects are permitted.
+  ///
+  /// __Note__: This method, along with [getDerivedStateFromError] will only
+  /// be called if `skipMethods` in [registerComponent2] is overridden with
+  /// a list (which can be empty). Otherwise, in order to prevent every
+  /// component from being an error boundary, [componentDidCatch] and
+  /// [getDerivedStateFromError] will be ignored.
+  ///
+  /// See: <https://facebook.github.io/react/docs/react-component.html#componentdidcatch>
+  void componentDidCatch(dynamic error, dynamic info) {}
```
greglittlefield-wf: Does `info` have a known structure? Could we type it with a JS interop class?

One thing to keep in mind is that any tightening to lifecycle method args is considered a breaking change, so we have to carefully consider what these types should be.

js_src/_dart_helpers.js
```diff
@@ -103,13 +103,26 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
...
+      delete ReactDartComponent2[method];
```
greglittlefield-wf: Does deletion work for non-static methods?

I would think you'd have to do:
```js
delete ReactDartComponent2.prototype[method];
```

test/lifecycle_test/component2.dart
joebingham-wk: We did not, the test has been added though.

js_src/_dart_helpers.js
greglittlefield-wf: Could we also add some tests that this is working properly?
- lifecycle methods are not called for skipped static methods
- lifecycle methods are not called for skipped instance methods

lib/react_client.dart
```diff
@@ -708,9 +755,10 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
 /// Creates and returns a new [ReactDartComponentFactoryProxy] from the provided [componentFactory]
 /// which produces a new JS [`ReactClass` component class](https://facebook.github.io/react/docs/top-level-api.html#react.createclass).
 ReactDartComponentFactoryProxy2 _registerComponent2(ComponentFactory<Component2> componentFactory,
-    [Iterable<String> skipMethods = const []]) {
+    [Iterable<String> skipMethods = const ['getDerivedStateFromError', 'componentDidCatch']]) {
```
greglittlefield-wf: Could we add a test that asserts that components are not error boundaries by default?

lib/react_client/react_interop.dart
joebingham-wk: Indeed!

test/lifecycle_test.dart
joebingham-wk: We spent some time trying to figure that out. At one point I had to branch off of Keal's wip context branch, and we believe it may be leftovers from that or a reaction to those changes. Ultimately when we looked at the test, the additions made sense and reflect what we'd expect to see!

lib/react_client.dart
aaronlademann-wf: Ah, ok - forwarding the cases to `_registerComponent2` sounds good @joebingham-wk

test/lifecycle_test/component2.dart
joebingham-wk: That was a good catch.. it shouldn't.

js_src/_dart_helpers.js
joebingham-wk: > ```js
> delete ReactDartComponent2.prototype[method];
> ```

Good call! Adding tests now.

test/lifecycle_test.dart
greglittlefield-wf: 👍

test/lifecycle_test.dart
```diff
@@ -183,11 +183,55 @@ main() {
         LifecycleTestHelper.staticLifecycleCalls.clear();
         component.setState({"shouldThrow": true});
 
-        expect(component.lifecycleCalls, containsAllInOrder(['getDerivedStateFromError', 'componentDidCatch']));
-        expect(component.state["shouldThrow"], isFalse);
+        // First render throws an error, caught by `getDerivedStateFromError`.
+        // `getDerivedStateFromError` sets state, starting the update cycle
+        // again. `componentDidCatch` also sets the state (storing error
+        // information), starting the update cycle again.
```
greglittlefield-wf: Awesome explanation! 🥇

test/lifecycle_test.dart
```diff
@@ -175,6 +175,86 @@ main() {
...
+              'render',
+              'getDerivedStateFromError',
+              'shouldComponentUpdate',
+              'render',
+              'componentDidUpdate',
+              'componentDidCatch',
+              'shouldComponentUpdate',
+              'render',
+              'componentDidUpdate'
+            ]));
```
greglittlefield-wf: Could we add a reason here to make it clearer how we're testing this?
```suggestion
            ]), reason: 'should have skipped getSnapshotBeforeUpdate');
```

test/lifecycle_test.dart
```diff
@@ -175,6 +175,86 @@ main() {
...
+      test('passes the correct error/info to lifecycle methods when an error is thrown', () {
+        int lengthOfKey = 15;
+        var mountNode = new DivElement();
+        var renderedInstance = react_dom.render(components2.SetStateTest({}), mountNode);
+        LifecycleTestHelper component = getDartComponent(renderedInstance);
+        Element renderedNode = react_dom.findDOMNode(renderedInstance);
+        LifecycleTestHelper.staticLifecycleCalls.clear();
+        component.setState({"shouldThrow": true});
+
+        expect(renderedNode.children[1].text.contains(getComponent2ErrorMessage()), isTrue);
```
greglittlefield-wf: I would recommend the contains matcher; it gives you much more helpful error messages when this fails
```suggestion
        expect(renderedNode.children[1].text, contains(getComponent2ErrorMessage()));
```
Error message when using `isTrue`:
```
Expected: true
  Actual: <false>
```
Error message when using `contains`:
```
Expected: contains 'bar'
  Actual: 'foo'
```

test/lifecycle_test.dart
```diff
@@ -175,6 +175,86 @@ main() {
...
+        expect(renderedNode.children[2].text.substring(1, lengthOfKey), getComponent2ErrorInfo());
```
greglittlefield-wf: What's `lengthOfKey` for? Could we just use `contains` here as well?

test/lifecycle_test.dart
```diff
@@ -175,6 +175,86 @@ main() {
...
+        expect(renderedNode.children[3].text.contains(getComponent2ErrorFromDerivedState()), isTrue);
```
greglittlefield-wf: Same here, use `contains`:
```suggestion
        expect(renderedNode.children[3].text, contains(getComponent2ErrorFromDerivedState()));
```

test/ReactSetStateTestComponent2.js
```diff
@@ -39,22 +39,27 @@ function getComponent2ErrorMessage(){
 }
 
 function getComponent2ErrorInfo(){
-    return _info.toString();
+    return Object.keys(_info)[0];
```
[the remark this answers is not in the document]
joebingham-wk: I may be misunderstanding, but that would return the value, right? Just making sure because in this case the key is checked in the test rather than the value, so I was attempting to dynamically get the key to make sure it matches what we'd expect it to be!

test/lifecycle_test.dart
joebingham-wk: Most likely - similar to one of the comments above though, adding the `ReactErrorInfo` interop class changed this assertment in particular. So the test changed and now definitely doesn't call for `lengthOfKey`.

As general explanation though - the test was checking to make sure both the JS and Dart objects had the same key because the values will be different call stacks. So `lengthOfKey` was the int needed to get a substring (just the object key) to match the Dart key with the JS side, but I wanted to avoid magic numbers. That variable was my attempt at a descriptive name!

PR #176, opened 2019-03-29
CPLAT-4767 Implement Component2 "init" method to replace componentWillMount

tainhenning-wk:
Description:
-Added init lifecycle method that maps the js constructor to dart
-Deprecated componentWillMount and getInitialState
https://jira.atl.workiva.net/browse/CPLAT-4767

Testing:
 - [ ] CI
 - [ ] Tests passing


lib/react_client.dart
```diff
@@ -496,7 +496,7 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
           ..jsThis = jsThis
           ..props = new JsBackedMap.backedBy(jsThis.props)
           ..context = _unjsifyNewContext(jsThis.context);
-
+        component.init();
```
aaronlademann-wf: I believe there is more to it than this...

From the ticket AC:

```dart
component.init();

// Sync the JS component's state the component's state
if (jsThis.dartComponent.state != null) {
  jsThis.state = new JsBackedMap.from(jsThis.dartComponent.state).jsMap;
}
```

Which also indicates to me that lines 495-498 should probably come after `component.init()` et. al. like so:

```dart
component.init();

// Sync the JS component's state the component's state
if (jsThis.dartComponent.state != null) {
  jsThis.state = new JsBackedMap.from(jsThis.dartComponent.state).jsMap;
}

component
  ..jsThis = jsThis
  ..props = new JsBackedMap.backedBy(jsThis.props)
  ..context = _unjsifyNewContext(jsThis.context);

return component;
```

Does that look right to you @greglittlefield-wf @kealjones-wk ?

lib/react_client.dart
```diff
@@ -491,12 +491,16 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
         component.adapter = new JsComponent2Adapter(jsThis: jsThis);
         // Return the component so that the JS proxying component can store it,
         // avoiding an interceptor lookup.
+        component.init();
```
greglittlefield-wf: I believe we'll want to call this after initializing jsThis/props/context below, otherwise those won't be able to be used in `init`

test/lifecycle_test.dart
```diff
@@ -422,6 +423,8 @@ void sharedLifecycleTests<T extends react.Component>({
         expect(
             component.lifecycleCalls,
             equals([
+              // what should we be expecting for the props here in init?
```
greglittlefield-wf: See above comment

lib/react_client.dart
```diff
@@ -491,12 +491,16 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
...
+
+        if (jsThis.dartComponent != null && jsThis.dartComponent.state != null) {
+          jsThis.setState(new JsBackedMap.from(jsThis.dartComponent.state));
```
greglittlefield-wf: I don't think we want to do `setState` since that will result in additional state transition. I think you want something like the implementation on the ticket:
```dart
// Sync the JS component's state the component's state
if (component.state != null) {
  jsThis.state = new JsBackedMap.from(component.state).jsMap;
}
```

greglittlefield-wf: We'll also want to add tests that initializing state from `init` works as expected

lib/react.dart
```diff
@@ -450,6 +450,7 @@ abstract class Component {
   /// Invoked once before the `Component` is mounted. The return value will be used as the initial value of [state].
   ///
   /// See: <https://facebook.github.io/react/docs/react-component.html#getinitialstate>
+  ///
```
aaronlademann-wf:
```suggestion
```

test/lifecycle_test.dart
```diff
@@ -673,6 +676,46 @@ void sharedLifecycleTests<T extends react.Component>({
...
+
+    if (isComponent2) {
+      test('init testing throw :)', () {
+        const Map initialState = const {
+          'initialState': 'initial',
+        };
+
+        final Map initialProps = unmodifiableMap({
+          'init': (react.Component2 component) => component.state = initialState,
+          "getInitialState": (_) => initialState
```
aaronlademann-wf:
```suggestion
          'getInitialState': (_) => initialState
```

lib/react.dart
```diff
@@ -748,8 +754,27 @@ abstract class Component2 implements Component {
...
 
+  /// This is equivalent to `Constructor` in React 16, this is called before mounting
+  /// See: <https://reactjs.org/docs/react-component.html#constructor>
+  ///
+  /// __Example__:
+  ///
+  ///    class MyClass extends react.Component2 {
+  ///      Map initialState = { foo: 0, bar: 1};
+  ///
+  ///      void init() {
```
aaronlademann-wf:
```suggestion
  ///      @override
  ///      void init() {
```

lib/react.dart
```diff
@@ -748,8 +754,27 @@ abstract class Component2 implements Component {
...
+  @Deprecated('6.0.0')
   Map getInitialState() => const {};
 
+  /// This is equivalent to `Constructor` in React 16, this is called before mounting
+  /// See: <https://reactjs.org/docs/react-component.html#constructor>
+  ///
+  /// __Example__:
+  ///
+  ///    class MyClass extends react.Component2 {
+  ///      Map initialState = { foo: 0, bar: 1};
```
aaronlademann-wf: There's no real value in suggesting that consumers store this as a separate variable.  Should just assign `this.state` to the map within `init` like so:

```dart
@override
void init() {
  this.state = {'foo': 0, 'bar': 1};
}
```

js_src/_dart_helpers.js
```diff
@@ -83,10 +83,17 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
   class ReactDartComponent2 extends React.Component {
     constructor(props, context) {
       super(props, context);
-      // TODO combine these two calls into one
+      this.state = null;
       this.dartComponent = dartInteropStatics.initComponent(this, componentStatics);
-      this.state = dartInteropStatics.handleGetInitialState(this.dartComponent);
+      const initialStateFromGetInitialState =
+      dartInteropStatics.handleGetInitialState(this.dartComponent);
+      if (this.state && Object.keys(initialStateFromGetInitialState).length != 0) {
```
greglittlefield-wf: #nit Could we add a comment to this line?
```suggestion
      // If the returned state is empty, assume it's coming from the default implementation.
      // This is done to avoid changing the default getInitialState implementation to null, which would cause breakages.
      if (this.state && Object.keys(initialStateFromGetInitialState).length != 0) {
```

lib/react.dart
```diff
@@ -748,8 +753,26 @@ abstract class Component2 implements Component {
...
 
+  /// This is equivalent to `Constructor` in React 16, this is called before mounting
+  /// See: <https://reactjs.org/docs/react-component.html#constructor>
+  ///
+  /// __Example__:
+  ///
+  ///    class MyClass extends react.Component2 {
+  ///      @override
+  ///      void init() {
+  ///        this.state = { 'foo': 0, 'bar': 1};
```
greglittlefield-wf: #nit extra space
```suggestion
  ///        this.state = {'foo': 0, 'bar': 1};
```

test/lifecycle_test.dart
```diff
@@ -673,6 +676,46 @@ void sharedLifecycleTests<T extends react.Component>({
...
+
+        final Map initialProps =
+            unmodifiableMap({'init': (react.Component2 component) => component.state = initialState});
+        final Map expectedProps = unmodifiableMap(defaultProps, initialProps, emptyChildrenProps);
+        LifecycleTestHelper component = getDartComponent(render(LifecycleTest(initialProps)));
+
+        expect(
+            component.lifecycleCalls,
+            equals([
+              matchCall('init', props: expectedProps),
```
greglittlefield-wf: Since we're dealing with state mutation, these should all probably have `state: ` expectations as well

test/lifecycle_test.dart
```diff
@@ -673,6 +676,46 @@ void sharedLifecycleTests<T extends react.Component>({
           ].where((matcher) => matcher != null).toList()));
     });
 
+    if (isComponent2) {
+      test('init testing', () {
```
greglittlefield-wf: This name could be more descriptive
```suggestion
      test('initializes state with the value set in `init`', () {
```

test/lifecycle_test.dart
```diff
@@ -673,6 +676,46 @@ void sharedLifecycleTests<T extends react.Component>({
...
+              matchCall('init', props: expectedProps),
+              matchCall('getInitialState', props: expectedProps),
+              matchCall('render', props: expectedProps),
+              matchCall('componentDidMount', props: expectedProps)
+            ].where((matcher) => matcher != null)));
+      });
+    }
+
+    if (isComponent2) {
+      test('init testing throw :)', () {
```
greglittlefield-wf:
```suggestion
      test('throws when state is initialized in both init and getInitialState', () {
```
:)

js_src/_dart_helpers.js
```diff
@@ -83,10 +83,17 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
   class ReactDartComponent2 extends React.Component {
     constructor(props, context) {
       super(props, context);
-      // TODO combine these two calls into one
```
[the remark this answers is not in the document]
greglittlefield-wf: Bump on this; I think we need to reinstate this comment.

test/lifecycle_test.dart
tainhenning-wk: It seems that getInitialState gets in the way is resetting the state to be an empty map, is this ok?

greglittlefield-wf: No, we don't want that to happen, since we want consumers that don't override `getInitialState` to be able to initialize state in `init`.

js_src/_dart_helpers.js
```diff
@@ -83,10 +83,20 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
...
+      this.state = null;
       this.dartComponent = dartInteropStatics.initComponent(this, componentStatics);
-      this.state = dartInteropStatics.handleGetInitialState(this.dartComponent);
+      const initialStateFromGetInitialState =
+      dartInteropStatics.handleGetInitialState(this.dartComponent);
+      // If the returned state is empty, assume it's coming from the default implementation.
+      // This is done to avoid changing the default getInitialState implementation to null, which would cause breakages.
+      if (this.state && Object.keys(initialStateFromGetInitialState).length != 0) {
+        throw 'Error: State cannot be initialized in both init and getInitialState';
+      } else {
```
greglittlefield-wf: RE my comment here https://github.com/cleandart/react-dart/pull/176#discussion_r276860480 

I think we can fix that by doing:
```suggestion
      } else if (!this.state) {
```

kealjones-wk:
+10

PR #177, opened 2019-04-04
Fix js component examples for release mode

kealjones-wk:
We merged a broken version of the JS Components Example, this fixes it and cleans it up a bit.


greglittlefield-wf:
+1

aaronlademann-wf:
+10

JS Components example serves / functions as expected with no errors in DDC or dart2js

PR #178, opened 2019-04-23
Fix componentDidCatch and getDerivedStateFromError error argument typing! 

kealjones-wk:
# Issue
`componentDidCatch` and `getDerivedStateFromError` in `Component2` used to be passed the `error` argument and it was not re-dartified so we would receive a `JsObject` in ddc or `NativeJavaScriptObject` in dart2js so all typing and properties were not available. see: https://github.com/dart-lang/sdk/issues/36363 for more detail.

# Fix
@greglittlefield-wf and I were able to find a solution that is able to re-dartify the error passed in from react and then forward that dartified error to the dart method of `Component2`.
How it works:
- Error/Exception is thrown from dart
- React catches the error JS side
- JS passes error as argument (this does not get dartified)
- Dart then passed the error to JS to re-throw the error
- Dart wraps that call to throw with a try-catch (this re-dartifies the error)
- In the catch on the dart side we then pass the dart error to the component method

# QA
 - [x] Run `pub run build_runner serve example`, go to http://localhost:8080/test/react_test.html click `Trigger Error` 
    - [x] "The Code is a 2" is above the button
    - [x] There is a non-error logged stack trace in the console that matches the React logged stacktrace.
 - [x] Run `pub run build_runner serve -r example`, go to http://localhost:8080/test/react_test.html click `Trigger Error`
    - [x] "The Code is a 2" is above the button
    - [x] There is a non-error logged stack trace in the console that matches the React logged stacktrace.
 - [x] CI Passes


lib/react_client.dart
```diff
@@ -614,11 +614,19 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
       });
 
   void handleComponentDidCatch(Component2 component, dynamic error, ReactErrorInfo info) => zone.run(() {
-        component.componentDidCatch(error, info);
+        try {
+          rethrowError(error);
```
joebingham-wk: Should we add any comments to explain why we're doing this? Some may be able to reason about it, but it seems a little unintuitive.

js_src/_dart_helpers.js
```diff
@@ -6,6 +6,11 @@ var _reactDartSymbolPrefix = 'react-dart.';
 /// A global symbol to identify javascript objects owned by react-dart context,
 /// in order to jsify and unjsify context objects correctly.
 var _reactDartContextSymbol = Symbol(_reactDartSymbolPrefix+'context');
+
```
joebingham-wk: I added a note in `componentDidCatch` about adding a comment to explain the purpose. This kind of goes in hand in hand with that - I think a comment here explaining `_rethrowError` could be useful.

lib/react_client.dart
```diff
@@ -614,11 +614,25 @@ final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
...
 
   void handleComponentDidCatch(Component2 component, dynamic error, ReactErrorInfo info) => zone.run(() {
-        component.componentDidCatch(error, info);
+        // Due to the error object being passed in from ReactJS it is a javascript object that does not get dartified.
+        // To fix this we throw the error again from dart to the JS side and catch it dart side which re-dartifies it.
+        try {
+          throwErrorFromJS(error);
+        } catch (e, stack) {
+          // The dart stack track gets lost so we manually add it to the info object for referance.
+          info.dartStackTrace = stack;
```
greglittlefield-wf: Can we add test coverage that this is populated properly?

lib/react.js
```diff
@@ -101,6 +101,13 @@ var _reactDartSymbolPrefix = 'react-dart.';
 /// A global symbol to identify javascript objects owned by react-dart context,
 /// in order to jsify and unjsify context objects correctly.
 var _reactDartContextSymbol = Symbol(_reactDartSymbolPrefix+'context');
+
+/// A JS side function to allow dart to throw an error from JS in order to catch it Dart side.
```
greglittlefield-wf: #nit
```suggestion
/// A JS side function to allow Fart to throw an error from JS in order to catch it Dart side.
```

lib/react_client.dart
kealjones-wk: also i need to capitalize dart... apparently...

greglittlefield-wf:
QA +1

PR #182, opened 2019-05-06
Deprecate emptyJsMap

aaronlademann-wf:
When deprecating a bunch of JS Object stuff in `4.7.0` we missed `emptyJsMap`.

@greglittlefield-wf


greglittlefield-wf:
+10

PR #187, opened 2019-05-23
CPLAT-5908 Add initializeState to Component2

joebingham-wk:
## Issue

over_react did not have an ideal API to initialize component state within the `init` lifecycle method. A pure over_react solution would have been detrimental to performance. The better solution was to create a direct method in react-dart that can convert a `Map` to a `JsBackedMap` and update the JS layer directly.

## Changes
- Add the `initializeState` method to `Component2`
- Add an `initializeState` method in `Component2Adapter` and override it in `JsCompoent2Adapter`
- Add Tests

## Review
_[See CONTRIBUTING.md][contributing-review-types] for more details on review types (+1 / QA +1 / +10) and code review process._

  

Please review: 
@aaronlademann-wf @greglittlefield-wf @kealjones-wk 

## QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -481,6 +481,7 @@ abstract class Component2Adapter {
   void setState(Map newState, SetStateCallback callback);
   void setStateWithUpdater(StateUpdaterCallback stateUpdater, SetStateCallback callback);
   void forceUpdate(SetStateCallback callback);
+  void initializeState(Map state);
 }
 
 /// Top-level ReactJS [Component class](https://facebook.github.io/react/docs/react-component.html)
@@ -588,6 +589,14 @@ abstract class Component2 implements Component {
   /// See: <https://reactjs.org/docs/react-component.html#setstate>
   void setStateWithUpdater(StateUpdaterCallback updater, [SetStateCallback callback]) {
     adapter.setStateWithUpdater(updater, callback);
+  }
+
+  /// Method used to initialize the [state] of a component.
+  ///
+  /// To avoid unnecessary Map copies and setter calls, the [init] method should use [initializeState] (instead of
+  /// this.state) to directly set the state on the JavaScript side.
+  void initializeState(Map value) {
+    adapter.initializeState(value);
   }
 
   void forceUpdate([SetStateCallback callback]) {
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -597,14 +597,12 @@ abstract class Component2 implements Component {
   /// supporting that is difficult in the Dart implementation, so this method must be used instead.
   ///
   /// __Example__:
-  ///     ```js
+  ///
   ///     // JavaScript
   ///     constructor() {
   ///       this.state = {count: 0};
   ///     }
-  ///     ```
   ///
-  ///     ```dart
   ///     // Dart
   ///     init() {
   ///       initializeState({'count': 0});
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -591,11 +591,26 @@ abstract class Component2 implements Component {
     adapter.setStateWithUpdater(updater, callback);
   }
 
-  /// Method used to initialize the [state] of a component.
+  /// Initializes this component's [state] to [value].
   ///
-  /// To avoid unnecessary Map copies and setter calls, the [init] method should use [initializeState] (instead of
-  /// this.state) to directly set the state on the JavaScript side.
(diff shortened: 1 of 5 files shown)


lib/react.dart
```diff
@@ -590,6 +591,14 @@ abstract class Component2 implements Component {
     adapter.setStateWithUpdater(updater, callback);
   }
 
+  /// Method used to initialize the [state] of a component.
```
greglittlefield-wf: As per the Dart style guide,
- https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs
- https://dart.dev/guides/language/effective-dart/documentation#prefer-using-this-instead-of-the-to-refer-to-a-members-instance

This should be something more like
```dart
/// Initializes this component's [state] to [value].
```

lib/react.dart
```diff
@@ -590,6 +591,14 @@ abstract class Component2 implements Component {
...
+  ///
+  /// To avoid unnecessary Map copies and setter calls, the [init] method should use [initializeState] (instead of
+  /// this.state) to directly set the state on the JavaScript side.
```
greglittlefield-wf: I think this is a little too much implementation detail for consumers.

An example might also be nice.

 What about something like:
```dart
/// Whereas `this.state` can be set directly in JavaScript components during initialization, 
/// supporting that is difficult in the Dart implementation, so this method must be used instead.
///
/// ```js
/// // JavaScript
/// constructor() {
///   this.state = {count: 0};
/// }
/// ```
///
/// ```dart
/// // Dart
/// init() {
///   initializeState({'count': 0});
/// }
```

test/lifecycle_test.dart
```diff
@@ -134,6 +134,15 @@ main() {
         isComponent2: true,
       );
 
+      test('initializes correctly using initializeState', () {
```
greglittlefield-wf: The existing tests using `this.state = ` should be updated to use this new API :
> `test('initializes state with the value set in `init`', () {`
> `test('throws when state is initialized in both init and getInitialState', () {`

After that's done, I don't think we'll need this new test anymore.

The `expect(component.state, isA<JsBackedMap>());` is a great addition, though! I think that should be added to the existing tests.

test/lifecycle_test/component2.dart
```diff
@@ -20,16 +20,17 @@ class _SetStateTest extends react.Component2 with LifecycleTestHelper {
...
-  getInitialState() => {
-        'counter': 1,
-        'shouldThrow': true,
-        'errorFromGetDerivedState': '',
-        'error': '',
-        'info': '',
-      };
+  void init() {
+    lifecycleCall('init');
+    this.initializeState({
```
greglittlefield-wf: #nit Using `this.` is discouraged when it isn't needed: https://dart.dev/guides/language/effective-dart/usage#dont-use-this-when-not-needed-to-avoid-shadowing

No worries, though; we should get more lints turned on in this repo so these kinds of checks can be automated

lib/react.dart
joebingham-wk: Thank you for the links!! It had been a while since I had referred to those and clearly I need a refresher!

test/lifecycle_test/component2.dart
joebingham-wk: 💯

lib/react.dart
```diff
@@ -590,6 +591,28 @@ abstract class Component2 implements Component {
     adapter.setStateWithUpdater(updater, callback);
   }
 
+  /// Initializes this component's [state] to [value].
+  ///
+  /// Whereas `this.state` can be set directly in JavaScript components during initialization,
+  /// supporting that is difficult in the Dart implementation, so this method must be used instead.
+  ///
+  /// __Example__:
+  ///     ```js
```
aaronlademann-wf: I don't think ``` is necessary / works to output code blocks in dart doc comments.  Should just be:

```
  /// __Example__:
  ///
  ///     // JavaScript
  ///     constructor() {
  ///       this.state = {count: 0};
  ///     }
  ///
  ///     // Dart
  ///     init() {
  ///       initializeState({'count': 0});
  ///     }
```

PR #190, opened 2019-06-26
Update setStateWithUpdater to allow null as a return

kealjones-wk:
# Motivation 
setStateWithUpdater for component2 previously passed the return value straight into the jsBackingMapOrJsCopy but if it returned null it would break when passed into that methods 'addAll' method, this short circuits that by returning null if the value is null.

# For discussion: 
an alternate approach to the `if(value == null) return null;` at https://github.com/cleandart/react-dart/pull/190/files#diff-12ac054bb8b76f35c1035afef1d94396R518 could be:
```
      if (value is Map) return jsBackingMapOrJsCopy(value);
      return value;
```


lib/react_client.dart
```diff
@@ -511,12 +511,13 @@ class JsComponent2Adapter extends Component2Adapter {
   @override
   void setStateWithUpdater(StateUpdaterCallback stateUpdater, SetStateCallback callback) {
     final firstArg = allowInterop((jsPrevState, jsProps, [_]) {
-      return jsBackingMapOrJsCopy(stateUpdater(
+      dynamic value = stateUpdater(
         new JsBackedMap.backedBy(jsPrevState),
         new JsBackedMap.backedBy(jsProps),
-      ));
+      );
+      if (value == null) return null;
```
kealjones-wk: instead of this we could do:
```
      if (value is Map) return jsBackingMapOrJsCopy(value);
      return value;
```

thoughts? @greglittlefield-wf @aaronlademann-wf @joebingham-wk @sydneyjodon-wk

kealjones-wk: StateUpdaterCallback has a return type Map
`typedef Map StateUpdaterCallback(Map prevState, Map props);`

joebingham-wk: I feel like the null check is a little cleaner - it's more common practice (from what I've seen) and easier to reason about.

lib/react_client/bridge.dart
```diff
@@ -91,10 +91,12 @@ class Component2BridgeImpl extends Component2Bridge {
   @override
   void setStateWithUpdater(Component2 component, StateUpdaterCallback stateUpdater, SetStateCallback callback) {
     final firstArg = allowInterop((JsMap jsPrevState, JsMap jsProps, [_]) {
-      return jsBackingMapOrJsCopy(stateUpdater(
+      final value = stateUpdater(
         new JsBackedMap.backedBy(jsPrevState),
         new JsBackedMap.backedBy(jsProps),
-      ));
+      );
+      if (value == null) return null;
```
greglittlefield-wf: This needs a test

PR #191, opened 2019-07-01
CPLAT-6232 normalize set state call

joebingham-wk:
# Motivation 
Calling `setState` in React 15 essentially always causes an update (regardless of whether or not a new state is passed in). This behavior changes in React 16. We want to ensure that `Component` does not break expectations while changing behavior in `Component2`.

# Changes
- Remove null check that stops the update if `null` is passed in
- Add tests to show that `Component` will update if `null` is passed but `Component2` will not

@aaronlademann-wf @kealjones-wk @greglittlefield-wf @sydneyjodon-wk


kealjones-wk:
QA +1

greglittlefield-wf:
+1

PR #192, opened 2019-07-03
CPLAT-6313 / CPLAT-4887: Add propTypes

kealjones-wk:
# Motivation
Allow prop validation via react.

# Changes
- Added interop layer to attach propTypes to a component
- Added `propTypes` getter on component2 (only works with component 2 due to requirement on JsBackedMaps)
- Added `jsPropTypesMap` to allow additional customization of how the interop handles typing.
- Updated react_test example to use propTypes
- Added tests


lib/react_client/react_interop.dart
```diff
@@ -51,6 +52,61 @@ abstract class ReactDom {
   static ReactPortal createPortal(dynamic children, Element container) => ReactDOM.createPortal(children, container);
 }
 
+typedef Error PropValidator(Map props, String propName, String componentName, String location, String propFullName);
+typedef Error TypedPropValidator<TProps>(
+    TProps props, String propName, String componentName, String location, String propFullName);
+
+// Do we want to expose these?
```
kealjones-wk: thoughts? after a convo with @greglittlefield-wf my thinking was probably not.

js_src/react.js
```diff
@@ -19,7 +19,7 @@ window.React = React;
 Object.assign(window, DartHelpers);
 
 React.createClass = CreateReactClass; // TODO: Remove this once over_react_test doesnt rely on createClass.
-React.PropTypes = PropTypes; // Only needed to support legacy context until we update.
+React.PropTypes = PropTypes; // Only needed to support legacy context until we update. lol jk we need it for prop validation now.
```
greglittlefield-wf: 😆

lib/react_client/react_interop.dart
kealjones-wk: @greglittlefield-wf did we think we should just take this all out for now?

test/lifecycle_test.dart
```diff
@@ -134,6 +131,42 @@ main() {
...
+
+        tearDown(() {
+          context['console']['error'] = originalConsoleError;
+        });
+
+        test('fails validation with incorrect prop type', () {
+          react_dom.render(components2.PropTypesTest({'intProp': 'test'}), mountNode);
+          expect(consoleErrorCalled, isTrue, reason: 'should have outputted a warning');
+          expect(consoleErrorMessage, contains('intProp should be int'));
+        });
```
greglittlefield-wf: Could we have an additional test here:
```dart
test('passes through all of the expected arguments to the prop validator', () { ... }
```
just to get coverage on this, maybe just dump all of those vars in an error message.

kealjones-wk: `test('passes through all of the expected arguments to the prop validator', () { ... });`
FTFY

kealjones-wk: https://github.com/cleandart/react-dart/pull/192/files#diff-c00a02aa016b25a86fb30a2093f935a1R171-R190 lulz

lib/react.dart
```diff
@@ -759,6 +763,13 @@ abstract class Component2 implements Component {
   /// See: <https://facebook.github.io/react/docs/react-component.html#getinitialstate>
   Map getInitialState() => const {};
 
+  /// Allows usage of PropValidator functions to check the validity of a prop passed to it.
+  /// When an invalid value is provided for a prop, a warning will be shown in the JavaScript console.
+  /// For performance reasons, propTypes is only checked in development mode.
+  ///
+  /// See: <https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes>
+  Map<String, PropValidator<Null>> get propTypes => {};
```
aaronlademann-wf: We should include an example of what usage looks like for this - complete with common examples _(e.g. validating a prop based on the value of another prop, etc.)_.

lib/react.dart
```diff
@@ -759,6 +763,43 @@ abstract class Component2 implements Component {
...
+  ///       get propTypes => {
+  ///         'name': (Map props, propName, componentName, location, propFullName) {
+  ///           if (props[propName].length > 20) {
+  ///             return ArgumentError('(${props[propName]}) is too long. $propName has a max length of 20 characters.');
+  ///           }
+  ///           return null;
+  ///         },
+  ///       };
+  ///
+  /// Example of 2 props being dependant:
```
greglittlefield-wf: typo
```suggestion
  /// Example of 2 props being dependent:
```

lib/react.dart
```diff
@@ -759,6 +763,43 @@ abstract class Component2 implements Component {
...
+  ///       get propTypes => {
+  ///         'someProp': (Map props, propName, componentName, location, propFullName) {
+  ///           if (props[propName] == true && props['anotherProp'] == null) {
+  ///             return ArgumentError('If (${props[propName]}) is true. You must have a value for "anotherProp".');
+  ///           }
+  ///           return null;
+  ///         },
+  ///         'anotherProp': (Map props, propName, componentName, location, propFullName) {
+  ///           if (props[propName] != null && props['someProp'] != true) {
+  ///             return ArgumentError('You must set "someProp" to true, to use $propName.');
```
greglittlefield-wf: #nit unnecessary comma
```suggestion
  ///             return ArgumentError('You must set "someProp" to true to use $propName.');
```

lib/react_client/bridge.dart
```diff
@@ -110,7 +110,28 @@ class Component2BridgeImpl extends Component2Bridge {
...
+        String secret,
+      ) {
+        // Create a Dart consumable version of the JsMap.
+        var convertedProps = JsBackedMap.fromJs(props);
+        // Call the users validator with the newly wrapped props.
+        var error = validator(convertedProps, propName, componentName, location, propFullName);
+        if (error != null) {
+          return JsError(error.toString());
+        }
+        return error;
```
greglittlefield-wf: #nit this is a little confusing since it's always null at this point; what about
```dart
return error == null ? null : JsError(error.toString());
```

lib/react_client/react_interop.dart
```diff
@@ -131,6 +133,19 @@ abstract class ReactDomServer {
   external static String renderToStaticMarkup(ReactElement component);
 }
 
+/// Runtime type checking for React props and similar objects.
+///
+/// See: <https://reactjs.org/docs/typechecking-with-proptypes.html>
+/// See: <https://www.npmjs.com/package/prop-types>
+@JS('React.PropTypes')
+abstract class PropTypes {
+  /// PropTypes.checkPropTypes(...) only console.error.log(...)s a given message once.
```
greglittlefield-wf: #nit
```suggestion
  /// PropTypes.checkPropTypes(...) only console.error(...)s a given message once.
```

lib/react_client/bridge.dart
kealjones-wk:
```suggestion
        return error == null ? null : JsError(error.toString());
```

lib/react_client/bridge.dart
```diff
@@ -110,7 +110,28 @@ class Component2BridgeImpl extends Component2Bridge {
...
+        String location,
+        String propFullName,
+        // This is a required argument of PropTypes but is usally hidden from the JS consumer.
+        String secret,
+      ) {
+        // Create a Dart consumable version of the JsMap.
+        var convertedProps = JsBackedMap.fromJs(props);
+        // Call the users validator with the newly wrapped props.
+        var error = validator(convertedProps, propName, componentName, location, propFullName);
+        if (error != null) {
```
kealjones-wk:
```suggestion
```

lib/react_client/bridge.dart
```diff
@@ -110,7 +110,28 @@ class Component2BridgeImpl extends Component2Bridge {
...
+          return JsError(error.toString());
```
kealjones-wk:
```suggestion
```

lib/react_client/bridge.dart
```diff
@@ -110,7 +110,28 @@ class Component2BridgeImpl extends Component2Bridge {
...
+        }
```
kealjones-wk:
```suggestion
```

PR #193, opened 2019-07-05
CPLAT 4337 Expose React 16 createRef / forwardRef

sydneyjodon-wk:
# Motivation 
Expose React 16 `createRef` and `forwardRef`.

# Changes
- Expose React `createRef` function in react_interop.dart and wrap it in a class to handle typing.
- Wrap React `forwardRef` with `ReactJsComponentFactoryProxy`.
## Testing
- [ ] CI Passes
- [ ] Manual Testing

Please review:
@kealjones-wk @aaronlademann-wf @greglittlefield-wf @joebingham-wk

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -23,7 +23,7 @@ import "package:react/src/react_client/synthetic_event_wrappers.dart" as events;
 import 'package:react/src/typedefs.dart';
 import 'package:react/src/ddc_emulated_function_name_bug.dart' as ddc_emulated_function_name_bug;
 
-export 'package:react/react_client/react_interop.dart' show ReactElement, ReactJsComponentFactory, inReactDevMode, createRef;
+export 'package:react/react_client/react_interop.dart' show ReactElement, ReactJsComponentFactory, inReactDevMode, createRef, forwardRef;
 export 'package:react/react.dart' show ReactComponentFactoryProxy, ComponentFactory;
 export 'package:react/src/react_client/js_backed_map.dart' show JsBackedMap, JsMap, jsBackingMapOrJsCopy;
 
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -930,10 +930,8 @@ _convertBoundValues(Map args) {
 
 void _convertRefValue(Map args) {
   var ref = args['ref'];
-  if (ref != null) {
-    if (ref is react_interop.Ref) {
-      args['ref'] = ref.jsRef;
-    }
+  if (ref is react_interop.Ref) {
+    args['ref'] = ref.jsRef;
   }
 }
 
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -120,7 +120,11 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
       // with the Dart Component instance, not the ReactComponent instance.
       if (ref is _CallbackRef) {
         interopProps.ref = allowInterop((ReactComponent instance) => ref(instance?.dartComponent));
-      } else {
+      }
+      else if (ref is createRef) {
+        interopProps.ref = ref.jsCreateRef;
+      }
+      else {
         interopProps.ref = ref;
       }
     }
@@ -185,6 +189,10 @@ class ReactDartComponentFactoryProxy2<TComponent extends Component2> extends Rea
       // with the Dart Component instance, not the ReactComponent instance.
       if (ref is _CallbackRef) {
         propsForJs['ref'] = allowInterop((ReactComponent instance) => ref(instance?.dartComponent));
+      }
+
+      if (ref is createRef) {
+        propsForJs['ref'] = ref.jsCreateRef;
       }
     }
 
@@ -855,6 +863,7 @@ class ReactDomComponentFactoryProxy extends ReactComponentFactoryProxy {
   static void convertProps(Map props) {
     _convertBoundValues(props);
     _convertEventHandlers(props);
+    _convertRefValue(props);
   }
 }
 
@@ -917,6 +926,16 @@ _convertBoundValues(Map args) {
         return onChange(event);
       }
     };
+  }
+}
+
+_convertRefValue(Map args) {
+  if (args.containsKey('ref')) {
+    var ref = args['ref'];
+
+    if (ref is createRef) {
+      args['ref'] = ref.jsCreateRef;
+    }
   }
 }
 
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -23,7 +23,8 @@ import "package:react/src/react_client/synthetic_event_wrappers.dart" as events;
 import 'package:react/src/typedefs.dart';
 import 'package:react/src/ddc_emulated_function_name_bug.dart' as ddc_emulated_function_name_bug;
 
-export 'package:react/react_client/react_interop.dart' show ReactElement, ReactJsComponentFactory, inReactDevMode, createRef, forwardRef;
+export 'package:react/react_client/react_interop.dart'
+    show ReactElement, ReactJsComponentFactory, inReactDevMode, createRef, forwardRef;
 export 'package:react/react.dart' show ReactComponentFactoryProxy, ComponentFactory;
 export 'package:react/src/react_client/js_backed_map.dart' show JsBackedMap, JsMap, jsBackingMapOrJsCopy;
 
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -14,7 +14,8 @@ import 'dart:js_util';
 import "package:js/js.dart";
 import "package:react/react.dart";
 import 'package:react/react_client/js_interop_helpers.dart';
-import 'package:react/react_client/react_interop.dart';
+import 'package:react/react_client/react_interop.dart' hide Ref;
+import 'package:react/react_client/react_interop.dart' as react_interop show Ref;
 import "package:react/react_dom.dart";
 import 'package:react/react_dom_server.dart';
 import "package:react/src/react_client/event_prop_key_to_event_factory.dart";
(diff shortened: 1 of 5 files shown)


lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,54 @@ abstract class React {
...
+  /// Returns a ref that can be attached to an element via the ref attribute.
+  ///
+  /// When a ref is passed to an element in render, a reference to the node becomes
+  /// accessible at the 'current' attribute of the ref.
+  ///
+  /// See: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
+  external static createRef();
 }
 
+class createRef<CurrentType> {
```
greglittlefield-wf: This class and its members need doc comments.

The class comment should also have an example usage, showcasing the generic typing, and setting a convention for usage.

Something like
```dart
class FooComponent extends react.Component2 {
  final Ref<BarComponent> barRef = createRef();
  final Ref<Element> inputRef = createRef();

  render() => [
    Bar({'ref': barRef}),
    react.input({'ref': inputRef}),
  ];
}
```

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,54 @@ abstract class React {
...
+  }
+
+  CurrentType get current {
+    var jsCurrent = getProperty(jsCreateRef, 'current');
+    if (jsCurrent != null) {
+      if (hasProperty(jsCurrent, 'dartComponent')) {
+        return getProperty(jsCurrent, 'dartComponent');
+      }
+    }
+    return jsCurrent;
```
joebingham-wk: This may be a #superNit, but would it be worthwhile to move `return jsCurrent` into the if block starting on line 58? As is, there is a chance of returning `null` under the guise of `jsCurrent`, which seems like it could be a little insidious. 

If `null` is an acceptable return value (not worthy of throwing), then this: 
```
    if (jsCurrent != null) {
      if (hasProperty(jsCurrent, 'dartComponent')) {
        return getProperty(jsCurrent, 'dartComponent');
      }

      return jsCurrent;
    }
    return null;
```
seems more declarative to what's could actually be happening. Definitely just a suggestion though!

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,54 @@ abstract class React {
...
 
+class createRef<CurrentType> {
+  var jsCreateRef;
+
+  createRef() {
+    jsCreateRef = React.createRef();
+  }
+
+  CurrentType get current {
+    var jsCurrent = getProperty(jsCreateRef, 'current');
```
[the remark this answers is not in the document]
kealjones-wk: This may be worth another try but I think I mentioned this today, this approach was actually the very first thing we tried but because js side react does that object.freeze it actually has panic attacks if you mess around with that at all and I’m not sure what or how but I believe the interceptors are part of it and change the object and so then react looses its mind if when it sees not it’s own object value.

lib/react_client/react_interop.dart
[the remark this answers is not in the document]
kealjones-wk: Oh I guess we could just have a like function that returns a new class instance of this class, I was thinking since you need to construct it similarly to the function we wouldn’t need it but that makes sense. So something like
`CreateRef createRef() => new CreateRef();`
?

test/factory/common_factory_tests.dart
```diff
@@ -214,6 +214,42 @@ void refTests(ReactComponentFactoryProxy factory, {void verifyRefValue(dynamic r
 
     verifyRefValue(renderedInstance.dartComponent.ref('test'));
   });
+
+  test('createRef function creates ref with correct value', () {
+    var refValue = createRef();
```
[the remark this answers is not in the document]
sydneyjodon-wk: I changed `refValue` to `ref` in the `createRef` test, but in the `forwardRef` test there is already a `ref` parameter so I changed the name `refValue` to `refObject`.

lib/react_client/react_interop.dart
sydneyjodon-wk: That makes sense. I ended up removing the null check in the if statement and just checking if `jsCurrent` is null while getting the `dartComponent` property. I can still add that null return if you still think it’s necessary, but it will add another layer of if statements or an if statement at the top to return null if `jsCurrent` is null. Let me know what you think!

joebingham-wk: Nope, it was convenient before but with the additional specificity I think it's great as is!

lib/react_client.dart
```diff
@@ -919,6 +928,15 @@ _convertBoundValues(Map args) {
   }
 }
 
+void _convertRefValue(Map args) {
+  var ref = args['ref'];
+  if (ref != null) {
+    if (ref is react_interop.Ref) {
```
greglittlefield-wf: #nit In Dart, `is`-checks on `null` are always false except for `is Null` or `is Object`, so you don't need an additional type check here.
```dart
void main() {
  print(null is int);      // false
  print(null is Function); // false
  print(null is MyClass);  // false
  print(null is Object);   // true
  print(null is Null);     // true
}

class MyClass {}
```

lib/react_client/react_interop.dart
```diff
@@ -5,11 +5,13 @@
...
 
 import 'dart:html';
+import 'dart:js_util';
 
 import 'package:meta/meta.dart';
 import 'package:js/js.dart';
 import 'package:react/react.dart';
-import 'package:react/react_client.dart' show ComponentFactory;
+import 'package:react/react_client.dart' show ComponentFactory, ReactJsComponentFactoryProxy;
+import 'package:react/react_client.dart';
```
greglittlefield-wf: Duplicate import of the previous line; instead, the `show` should be updated with what we need from this lib.

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,83 @@ abstract class React {
 
   external static bool isValidElement(dynamic object);
   external static ReactClass get Fragment;
+
+  external static JsRef createRef();
+}
+
+/// Creates a [Ref] object that can be attached to an element via the ref attribute.
```
greglittlefield-wf: "prop" would be more accurate here than "attribute".

Also, #nit the term "element" is a little overloaded, so we prefer distinguishing between DOM Elements and ReactElements. 

```suggestion
/// Creates a [Ref] object that can be attached to a [ReactElement] via the ref prop.
```

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,83 @@ abstract class React {
...
+
+/// The return type of [createRef], the constructor wraps [React.createRef] for use in
+/// Dart.
+///
+/// A [Ref] can be attached to an element via the ref attribute. When the element renders,
+/// a reference to that node becomes accessible at the [current] attribute of the [Ref].
+///
+/// See: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
+class Ref<CurrentType> {
+  JsRef jsRef;
```
greglittlefield-wf: This needs a doc comment

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,83 @@ abstract class React {
...
+///
+/// Learn more: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
+Ref<CurrentType> createRef<CurrentType>() {
+  return new Ref<CurrentType>();
+}
+
+/// The return type of [createRef], the constructor wraps [React.createRef] for use in
+/// Dart.
+///
+/// A [Ref] can be attached to an element via the ref attribute. When the element renders,
```
greglittlefield-wf: Same as above regarding verbiage here.

Also, some Dart style guide tips:
- you don't need to reference `[Ref]` since that's the class your documenting: https://dart.dev/guides/language/effective-dart/documentation#avoid-redundancy-with-the-surrounding-context
- prefer using the word "this" within doc comments to refer to the current class instance https://dart.dev/guides/language/effective-dart/documentation#prefer-using-this-instead-of-the-to-refer-to-a-members-instance
- prefer brevity https://dart.dev/guides/language/effective-dart/documentation#prefer-brevity


Another idea for how to word this:
```dart
/// When this is provided as the ref prop, a reference to the rendered component 
/// will be available via [current].
```

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,83 @@ abstract class React {
...
+/// See: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
+class Ref<CurrentType> {
+  JsRef jsRef;
+
+  Ref() {
+    jsRef = React.createRef();
+  }
+
+  /// When the element renders, a reference to that node becomes accessible at the
+  /// [current] attribute of the [Ref].
```
greglittlefield-wf: Related to above style guide tips, as well as "PREFER starting library or type comments with noun phrases." https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases
 this comment could be something more like:
```dart
/// A reference to the latest instance of the rendered component.
///
/// See: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
```

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,83 @@ abstract class React {
...
+  ///
+  /// See: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
+  CurrentType get current {
+    final jsCurrent = jsRef.current;
+
+    if (jsCurrent is! Element) {
+      final dartCurrent = jsCurrent?.dartComponent;
```
greglittlefield-wf: This is a dynamic call (jsCurrent has a static type `dynamic` here), which we want to avoid since there's no static analysis as to whether we're doing something wrong. So, we should cast it.

```suggestion
      final dartCurrent = (jsCurrent as ReactComponent)?.dartComponent;
```

lib/react_client/react_interop.dart
```diff
@@ -34,8 +36,83 @@ abstract class React {
...
+
+      if (dartCurrent != null) {
+        return dartCurrent;
```
greglittlefield-wf: If we do a cast above, we need another cast here, since react.Component isn't necessarily a subtype of `CurrentType`:
```suggestion
        return dartCurrent as CurrentType;
```

lib/react_client/react_interop.dart
sydneyjodon-wk: Thank you for the tips and references!

greglittlefield-wf:
QA +1

Merging

PR #194, opened 2019-07-15
Finalize Component2 API - part 1 of X

greglittlefield-wf:
## Motivation
Component2 was implemented via a large number of PRs, and has some parts that need to be cleaned up before release.

## Changes 
- Remove `bind` (will be deprecated in master in a follow-up PR) and support for bound values
- Remove addition of legacy `context` arg to `Component2.shouldComponentUpdate`
- Undo deprecations for:
    - `Component.redraw` - there's no replacement until `Component2`, which which `forceUpdate`
    - `Component.setState` - no deprecation is needed since users will get type errors in `Component2` if they pass in a non-`Map`
    - deprecated lifecycle methods - consumers will get deprecation errors after switching to `Component2`, and having these creates more deprecation noise since `Component` itself is already deprecated
- Make all deprecated/unsupported members in Component2 properly throw `UnsupportedError`
    - Some threw UnimplementedError which isn't semantically correct, and some didn't do anything
- Add `@mustCallSuper` to all lifecycle methods so that consumers that override them get the right runtime errors when calling super
- Update some doc comments for deprecated members

## Testing
- CI passes
- Examples still work


lib/react.dart
```diff
@@ -858,168 +767,294 @@ abstract class Component2 implements Component {
...
   ///
-  /// Will be removed when [Component] is removed in the `6.0.0` release.
+  /// Replace calls to this method with either:
+  ///
+  /// - [forceUpdate] (preferred) - forces rerender and bypasses [shouldComponentUpdate]
+  /// - `setState({})` - same behavior as this method: causes rerender but goes through [shouldComponentUpdate]
+  @override
   @Deprecated('6.0.0')
-  void replaceState(Map newState, [SetStateCallback callback]) => throw new UnimplementedError();
+  void redraw([SetStateCallback callback]) {
```
greglittlefield-wf: This is the same before, but was just moved

lib/react_client.dart
```diff
@@ -860,7 +861,6 @@ class ReactDomComponentFactoryProxy extends ReactComponentFactoryProxy {
 
   /// Prepares the bound values, event handlers, and style props for consumption by ReactJS DOM components.
```
aaronlademann-wf: #supernit this comment could be updated since "bound" values don't exist anymore... But that can certainly be done in one of the follow-up PRs.

lib/react_client.dart
```diff
@@ -906,35 +906,6 @@ _setValueToProps(Map props, val) {
...
-      boundValue[1](_getValueFromDom(event.target));
-
-      if (onChange != null) {
-        return onChange(event);
-      }
-    };
-  }
-}
-
-void _convertRefValue(Map args) {
```
aaronlademann-wf: Seems like this is still being called on line 865

lib/react_client/react_interop.dart
```diff
@@ -507,13 +507,7 @@ class ReactDartInteropStatics2 implements ReactDartInteropStatics {
...
         handleGetDerivedStateFromProps,
-    bool Function(
-      Component2 component,
-      JsMap jsNextProps,
-      JsMap jsNextState,
-      dynamic jsNextContext,
-    )
-        handleShouldComponentUpdate,
+    bool Function(Component2 component, JsMap jsNextProps, JsMap jsNextState) handleShouldComponentUpdate,
     // TODO: Should this be removed when we update Component2.componentWillUpdate to throw an UnsupportedError? (CPLAT-4766)
```
aaronlademann-wf: Is this TODO still valid?

lib/react_client.dart
greglittlefield-wf: Reinstated in force-pushed merge

lib/react_client.dart
greglittlefield-wf: Updated

lib/react_client/react_interop.dart
greglittlefield-wf: It was on the wrong line, but there was some leftover code from otehr lifecycle methods that I cleaned up in 06bce3f2e3b140d362a40e18f18729e0e1a4c8c8

PR #195, opened 2019-07-17
Finalize Component2 API - part 2 of X

greglittlefield-wf:
## Motivation
Component2 was implemented via a large number of PRs, and has some parts that need to be cleaned up before release.

## Changes
- Component2Bridge
    - Rename Component2Adapter -> Component2Bridge (a better description of the design pattern it uses)
    - Move into its own file
    - Use Expando for storing it so that it isn't part of Component's public API
    - Add stub for jsifyPropTypes (to be consumed in @kealjone-wk's PR)
    - Allow injection of custom bridge implementations within registerComponent2
- Re-wire up registerComponent2 since we will need it after all to be able to inject custom bridges
- Clean up and document JSBackedMap implementation
    - Make it always `Map<dynamic, dynamic>` since we can't trust that the keys/values coming from the JS map match the type parameters the consumer might use
- Misc cleanup
    - Make some interop-related members added in 5.1.0-wip visibleForTesting
    - Add some missing deprecations to JS interop files
    - Remove unused `bind`-related APIs

#### Boyscouting
- Ignore deprecations from current package to cut down on lints
- Fix miscellaneous lints, including removing unused code
- Rename unexported `Ref` typdef that conflicted with the new `Ref` type, which was confusing and made imports messy
- Make `_ReactDartInteropStatics2` a class with static methods instead of a closure, 
    - Is cleaner, more grokkable, and easier to work with in IDEs
    - Removes the need for a JS interop class that mirrors all of the function types for type safety

## Testing
- CI passes
- Examples still work


lib/react.dart
```diff
@@ -624,7 +648,7 @@ abstract class Component2 implements Component {
   ///      return null;
   ///    }
   ///
-  Map getDerivedStateFromProps(Map nextProps, Map prevState) {}
+  Map getDerivedStateFromProps(Map nextProps, Map prevState) => null;
```
greglittlefield-wf: These are fixes to the `missing_return` lint

lib/react_client.dart
```diff
@@ -476,72 +476,24 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
...
+abstract class _ReactDartInteropStatics2 {
+  // TODO 3.1.0-wip expose for testing?
+  static final zone = Zone.root;
 
-  JsComponent2Adapter({
-    this.jsThis,
-  });
-
-  @override
-  void forceUpdate(SetStateCallback callback) {
```
greglittlefield-wf: These adapter method implementations are pretty much unchanged, and just moved to bridge.dart

lib/react_client/js_backed_map.dart
```diff
@@ -0,0 +1,153 @@
...
+/// Keys should be of primitive types (String, num, bool, null), as any non-primitive Dart objects will be
+/// canonicalized to their JavaScript string representations upon read/write.
+/// For performance reasons, keys are not validated.
+///
+/// `null` is allowed as a key.
+///
+/// Iteration order is arbitrary, and is based on the current browser's implementation.
+///
+/// Two JsBackedMap instances are considered equal if they are backed by the same [jsObject].
+class JsBackedMap extends MapBase<dynamic, dynamic> {
```
greglittlefield-wf: This implementation was moved, and pretty much just got things removed from it.

However, it could use a thorough re-review.

lib/react_client/react_interop.dart
```diff
@@ -398,6 +410,7 @@ class ReactDartContextInternal {
 
 /// Throws the error passed to it from Javascript.
 /// This allows us to catch the error in dart which re-dartifies the js errors/exceptions.
+@alwaysThrows
```
greglittlefield-wf: This fixes false positives with the `missing_return` lint

test/lifecycle_test/component2.dart
```diff
@@ -207,16 +208,12 @@ class _ErrorComponent extends react.Component2 {
 ReactDartComponentFactoryProxy2 LifecycleTest = react.registerComponent(() => new _LifecycleTest());
 
 class _LifecycleTest extends react.Component2 with LifecycleTestHelper {
-  void componentWillMount() => lifecycleCall('componentWillMount');
```
greglittlefield-wf: Deprecated lifecycle methods won't ever be called

test/lifecycle_test.dart
```diff
@@ -66,18 +68,14 @@ main() {
         int secondStateUpdateCalls;
         Map initialProps;
         Map newState1;
-        Map expectedState1;
```
greglittlefield-wf: These variables and statements were unused and had no side-effects, so I removed them

lib/react.dart
```diff
@@ -997,7 +1021,7 @@ abstract class Component2 implements Component {
   /// Will be removed when [Component] is removed in the `6.0.0` release.
   @override
   @Deprecated('6.0.0')
-  Ref get ref => throw _unsupportedError('ref');
+  RefMethod get ref => throw _unsupportedError('ref');
```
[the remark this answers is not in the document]
greglittlefield-wf: Nope, for functions it doesn't matter how you declare them

kealjones-wk:
+10

lib/react_client.dart
```diff
@@ -476,72 +476,24 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
...
+    component.context = _unjsifyNewContext(jsContext);
   }
-}
-
-final ReactDartInteropStatics2 _dartInteropStatics2 = (() {
-  final zone = Zone.current;
 
   /// Wrapper for [Component.getInitialState].
-  Component2 initComponent(ReactComponent jsThis, ComponentStatics2 componentStatics) => zone.run(() {
+  static Component2 initComponent(ReactComponent jsThis, ComponentStatics2 componentStatics) => //
```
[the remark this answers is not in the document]
greglittlefield-wf: Yean, it can be used to force dartfmt to insert a line break. I put these here because IMO it was easier to read to have all the `zone.run` calls on their own lines, as opposed to having just some of them.

> It seems to break pattern on line 550.
Good point; I'll add it to all of them for consistency's sake

I can also add a comment that makes it clearer what it's doing:
```dart
// dartfmt
```

PR #196, opened 2019-07-24
CPLAT-2779 Compatibility stuff for over_react_redux

kealjones-wk:
- Adds ReactRedux to ReactDom bundle (sorry native)
- Adds optional boolean `alwaysReturnChildrenAsList` to `ReactJsComponentFactoryProxy` 
- Updates JS bundles


package.json
```diff
@@ -15,7 +15,9 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-dom-factories": "^1.0.2",
+    "react-redux": "^7.1.0",
     "react-test-renderer": "^16.8.3",
+    "redux": "^4.0.4",
```
kealjones-wk: we don't use this but NPM complains if it isn't a peer dependency :/

lib/src/context.dart
```diff
@@ -0,0 +1,147 @@
...
+///             return react.span({}, [
+///               '$value', // Outputs: 'new context value'
+///             ]),
+///           });
+///         ]);
+///       }
+///     }
+///
+/// Learn more at: https://reactjs.org/docs/context.html
+class ReactDartContext {
```
greglittlefield-wf: Idea: since this is now being namespaced, what about just calling it `Context`?

example/test/react_test_components.dart
```diff
@@ -272,7 +272,7 @@ int calculateChangedBits(currentValue, nextValue) {
   return result;
 }
 
-var TestNewContext = createContext({'renderCount': 0}, calculateChangedBits);
+var TestNewContext = react.createContext({'renderCount': 0}, calculateChangedBits);
```
sydneyjodon-wk: #nit remove unused `import 'package:react/react_client.dart'` on line 5

lib/src/context.dart
kealjones-wk: I use `Context` for the over_react version, so i kept this one as `ReactReduxContext` just so there was no confusion between the two but i could, but should i?

kealjones-wk:
Failures are because of dart1, screw dart1

lib/src/context.dart
```diff
@@ -0,0 +1,147 @@
...
+///       render() {
+///         return react.span({}, [
+///           '${this.context}', // Outputs: 'test'
+///         ]);
+///       }
+///     }
+///
+/// ___ OR ___
+///
+///     Context MyContext = createContext();
```
greglittlefield-wf: #nit this should have the react prefix
```suggestion
///     react.Context MyContext = react.createContext();
```

lib/src/context.dart
```diff
@@ -0,0 +1,147 @@
...
+
+/// Creates a [Context] object. When React renders a component that subscribes to this [Context]
+/// object it will read the current context value from the closest matching Provider above it in the tree.
+///
+/// The `defaultValue` argument is only used when a component does not have a matching [Context.Provider]
+/// above it in the tree. This can be helpful for testing components in isolation without wrapping them.
+///
+/// __Example__:
+///
+///     Context MyContext = createContext('test');
```
greglittlefield-wf: #nit this should have the react prefix
```suggestion
///     react.Context MyContext = react.createContext('test');
```

lib/src/context.dart
```diff
@@ -0,0 +1,147 @@
...
+///             ]),
+///           });
+///         ]);
+///       }
+///     }
+///
+/// Learn more: https://reactjs.org/docs/context.html#reactcreatecontext
+Context createContext([
+  dynamic defaultValue,
+  int Function(dynamic currentValue, dynamic nextValue) calculateChangedBits,
```
greglittlefield-wf: Could we make these generic just like we do in over_react?

https://github.com/Workiva/over_react/pull/322/files#diff-1711dd72be000baed949393d3073152cR226

lib/src/context.dart
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// A JavaScript symbol that we use as the key in a JS Object to wrap the Dart.
+@JS()
+external get _reactDartContextSymbol;
+
+/// A context utility for to assist with common needs of ReactDartContext.
+///
+/// __For internal/advanced use only.__
+class ContextHelpers {
```
greglittlefield-wf: #nit could be `abstract` since it only contains statics and should not be implemented
```suggestion
abstract class ContextHelpers {
```

lib/src/context.dart
kealjones-wk: but greg, idk if you realize, but this is react-dart... :P

PR #200, opened 2019-08-01
defaultProps getter

kealjones-wk:
# why?
why not?

# changes
literally like 1 line of code then just updates to tests and example


test/lifecycle_test.dart
```diff
@@ -196,7 +196,8 @@ main() {
           'initialState': 'initial',
         };
 
-        final Map initialProps = unmodifiableMap({'getInitialState': (_) => initialState});
+        final Map initialProps =
+            unmodifiableMap({'getInitialState': (_) => initialState, 'initialState': (_) => initialState});
```
aaronlademann-wf: Why is having both `getInitialState` and `initialState` as a key in this map necessary?

test/lifecycle_test.dart
```diff
@@ -448,7 +449,7 @@ void sharedLifecycleTests<T extends react.Component>({
 }) {
   group('(shared behavior)', () {
     group('default props', () {
-      test('getDefaultProps() is only called once per component factory and cached', () {
+      test('defaultProps is only called once per component factory and cached', () {
```
aaronlademann-wf: We typically leave test coverage around for deprecated stuff - I'd like to see the `getDefaultProps` test moved somewhere that it can still be run for `Component` only.

test/lifecycle_test.dart
```diff
@@ -876,24 +872,6 @@ void sharedLifecycleTests<T extends react.Component>({
       });
     }
 
-    if (isComponent2) {
-      test('throws when state is initialized in both init and getInitialState', () {
```
aaronlademann-wf: Don't we need a test asserting that calling `getInitialState` throws in `Component2`?

lib/react_client.dart
```diff
@@ -483,21 +482,14 @@ abstract class _ReactDartInteropStatics2 {
...
-        if (component.state != null) {
-          jsThis.state = new JsBackedMap.from(component.state).jsObject;
-        }
+        component..state = new JsBackedMap.backedBy(jsThis.state);
 
+        Component2Bridge.bridgeForComponent[component] = componentStatics.bridgeFactory(component);
         return component;
       });
 
-  static JsMap handleGetInitialState(Component2 component) => // dartfmt
```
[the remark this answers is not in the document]
kealjones-wk: That was new in 5.1.0-wip, does not exist in 5.0.0 or master

lib/react_client/bridge.dart
```diff
@@ -84,12 +83,6 @@ class Component2BridgeImpl extends Component2Bridge {
     }
   }
 
-  @override
-  void initializeState(Component2 component, Map state) {
```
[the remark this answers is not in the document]
kealjones-wk: nope. introduced in 5.1.0-wip, doesnt exist in master or 5.0.0-wip

test/lifecycle_test.dart
kealjones-wk: because this test suite is shared it runs against component1 and component2 and because component2 does not have a getInitalState we need to provide both and only the one that it uses will be executed

test/lifecycle_test.dart
kealjones-wk: The test for both component2 defaultProps and component1 getDefaultProps is still there maybe we could instead just do: 
```suggestion
      test('is only called once per component factory and cached', () {
```

test/lifecycle_test.dart
kealjones-wk: @greglittlefield-wf  is going to add the throws tests for `getInitialState` and `getDefaultProps`

lib/react_client/react_interop.dart
```diff
@@ -582,6 +587,7 @@ class JsComponentConfig {
 class JsComponentConfig2 {
   external factory JsComponentConfig2({
     dynamic contextType,
+    JsMap initialState,
```
greglittlefield-wf: This isn't used and is left over from older code, right?
```suggestion
```

lib/react.dart
```diff
@@ -517,6 +517,20 @@ abstract class Component2 implements Component {
...
+  ///
+  /// This method is invoked before any instances are created and thus cannot rely on [props]. In addition, be aware
+  /// that any complex objects returned by `defaultProps` will be shared across instances, not copied.
+  ///
+  /// See: <https://reactjs.org/docs/react-component.html#defaultprops>
+  Map get defaultProps => const {};
+
+  /// Invoked once before the `Component` is mounted. The return value will be used as the initial value of [state].
+  ///
+  /// See: <https://facebook.github.io/react/docs/react-component.html#getinitialstate>
```
greglittlefield-wf: This link is broken. Should link to these sections instead?
- https://reactjs.org/docs/react-without-es6.html#setting-the-initial-state
- https://reactjs.org/docs/react-component.html#constructor

and maybe add a note along the lines of 
> this is similar to getInitialState, and replaces the `this.state =` seen in React ES6 examples
?

greglittlefield-wf: Or, maybe have an example like the one removed below?
```dart
  /// __Example__:
  ///
  ///     // ES6 component
  ///     constructor() {
  ///       this.state = {count: 0};
  ///     }
  ///
  ///     // Dart component
  ///     @override
  ///     get initialState => {'count': 0};
```

lib/react.dart
```diff
@@ -523,12 +523,32 @@ abstract class Component2 implements Component {
   /// This method is invoked before any instances are created and thus cannot rely on [props]. In addition, be aware
   /// that any complex objects returned by `defaultProps` will be shared across instances, not copied.
   ///
-  /// See: <https://reactjs.org/docs/react-component.html#defaultprops>
```
greglittlefield-wf: I see that you removed these links, but I think it's good to have links back to the official React docs where possible. Could you reinstate them (and potentially use the newer ones I posted in a comment for `initialState`?

PR #205, opened 2019-08-20
Fix context setter typing to match getter and not fail implicit_casts

greglittlefield-wf:
## Motivation 
With `implicit_casts: false` in consumer repos, all components exhibit the following error:
> error: The return type of getter 'Component.context' is 'dynamic' which isn't assignable to the type 'Map<dynamic, dynamic>' of its setter 'Component.context'.

## Changes
Update setter to be dynamic.

## Testing
- CI passes
- Verify that the issue does not reproduce when pulled into a consumer repo with `implicit_casts: false`
    ```
    dependency_overrides:
      react:
        git:
          url: git@github.com:cleandart/react-dart.git
          ref: greglittlefield-wf-patch-1
    ```

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -62,7 +62,7 @@ abstract class Component {
   /// >
   /// > It is strongly recommended that you do not use this, and instead wait for `Component2.context`.
   @experimental
-  set context(Map value) => _context = value;
+  set context(dynamic value) => _context = value;
 
   /// ReactJS [Component] props.
   ///
--- a/test/lifecycle_test.dart
+++ b/test/lifecycle_test.dart
@@ -682,11 +682,7 @@ external List getUpdatingSetStateLifeCycleCalls();
 external List getNonUpdatingSetStateLifeCycleCalls();
 
 /// A test helper to record lifecycle calls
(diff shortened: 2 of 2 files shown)


test/lifecycle_test.dart
```diff
@@ -682,11 +682,7 @@ external List getUpdatingSetStateLifeCycleCalls();
 external List getNonUpdatingSetStateLifeCycleCalls();
 
 /// A test helper to record lifecycle calls
-abstract class LifecycleTestHelper {
-  Map context;
```
greglittlefield-wf: This was causing analysis errors since it was overriding the context setter without `covariant`.

lib/react.dart
```diff
@@ -62,7 +62,7 @@ abstract class Component {
   /// >
   /// > It is strongly recommended that you do not use this, and instead wait for `Component2.context`.
   @experimental
-  set context(Map value) => _context = value;
+  set context(dynamic value) => _context = value;
```
greglittlefield-wf: This is actually how it is in 5.0.0-wip; we must have missed it when backporting related changes

PR #209, opened 2019-09-11
CPLAT-7369 Add Check for Dart Map

joebingham-wk:
## Why?
`unconvertJsProps` will take in a Dart `Component`, when it shouldn't. This PR adds a check and throws if that happens.

## Changes
- Add an if statement checking for `Component` only props, and throw if they're found.

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -855,8 +855,12 @@ Map unconvertJsProps(/* ReactElement|ReactComponent */ instance) {
   // Convert the nested style map so it can be read by Dart code.
   var style = props['style'];
   if (style != null) {
-    props['style'] = Map<String, dynamic>.from(JsBackedMap.backedBy(style));
-  }
+      if (style is! Map) {
+        props['style'] = Map<String, dynamic>.from(JsBackedMap.backedBy(style));
+      } else {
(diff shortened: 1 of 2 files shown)


lib/react_client.dart
```diff
@@ -846,6 +846,12 @@ final Expando<Function> _originalEventHandlers = new Expando();
 /// Dart synthetic counterparts.
 Map unconvertJsProps(/* ReactElement|ReactComponent */ instance) {
   var props = Map.from(JsBackedMap.backedBy(instance.props));
+
+  // Catch if a Dart component has been passed in.
+  if (props['internal'] is ReactDartComponentInternal || (props['style'] != null && props['style'] is Map)) {
```
aaronlademann-wf: Pardon my ignorance - but I'm not following why a `Map` value for `props['style']` indicates that a Dart component instance has been passed in here?  Possibly deserves a bit more info in the code comment, IMO.

joebingham-wk: Good call. I added a comment explaining my logic. Let me know if there's a better way to detect `Component2` - it's not the best to rely on the style prop, but it does stop the error that created the need for the PR.

PR #211, opened 2019-09-16
Remove Component.bind

greglittlefield-wf:
`Component.bind` was slated for removal in 5.0.0, and got removed in 5.1.0-wip, but not 5.0.0-wip. We need to remove it.

Note that  the deprecation in master is "5.0.0", while 5.0.0-wip has the out-of-date "6.0.0":
https://github.com/cleandart/react-dart/blob/master/lib/react.dart#L125-L126


kealjones-wk:
+10

aaronlademann-wf:
Stable phase of CI passed. Merging.

PR #212, opened 2019-09-18
CPLAT-7459 Test coverage improvements and miscellaneous fixes

greglittlefield-wf:
### Motivation
There were some missing areas of test coverage for some of the new APIs added in 5.x, as well as some issues discovered while improving test coverage in the corresponding React 16 branches of over_react.

### Changes
- Move export of `createRef`/`forwardRef` to `react.dart` instead of `react_client.dart` to avoid conflicting with over_react's exports of similarly-named functions. This change makes them consistent with other react-dart APIs replaced by over_react (findDomNode, Component, etc.)
- Add missing test coverage for refs for JS components, fix conversion of Dart `Ref` objects
- Fix `forwardRef` passing through a `JsRef` instead of a `Ref` to the consumer-provided function
- Add `ReactDartComponentVersion` w/ constants and helpers around `ReactClass.dartComponentVersion`
- `Ref` class boyscouting:
    - doc comment cleanup
    - update generic parameter to follow [single-letter conventions](https://dart.dev/guides/language/effective-dart/design#do-follow-existing-mnemonic-conventions-when-naming-type-parameters), since the word "current" in that context if anything makes the parameter more confusing

### Testing
- Unit tests pass
- Tests in over_react PR pass: https://github.com/Workiva/over_react/pull/338

### DIFF
--- a/lib/react_client/react_interop.dart
+++ b/lib/react_client/react_interop.dart
@@ -187,8 +187,44 @@ class ReactClass {
   @Deprecated('6.0.0')
   external set dartDefaultProps(Map value);
 
+  /// A string to distinguish between different Dart component implementations / base classes.
+  ///
+  /// See [ReactDartComponentVersion] for values.
+  ///
+  /// __For internal use only.__
+  @protected
   external String get dartComponentVersion;
+  @protected
   external set dartComponentVersion(String value);
+}
+
+/// Constants for use with [ReactClass.dartComponentVersion] to distinguish
+/// different versions of Dart component implementations / base classes.
+///
+/// __For internal use only.__
+@protected
+@sealed
+abstract class ReactDartComponentVersion {
+  /// A [Component]-based component.
+  @protected
+  static const String component = '1';
+
+  /// A [Component2]-based component.
+  @protected
+  static const String component2 = '2';
+
+  /// Returns [ReactClass.dartComponentVersion] if [type] is the [ReactClass] for a Dart component
+  /// (a react-dart [ReactElement] or [ReactComponent]), and null otherwise.
+  @protected
+  static String fromType(dynamic type) {
+    // This check doesn't do much since ReactClass is an  an anonymous JS object,
+    // but it lets us safely cast to ReactClass.
+    if (type is ReactClass) {
+      return type.dartComponentVersion;
+    }
+
+    return null;
+  }
 }
 
 /// A JS interop class used as an argument to [React.createClass].
--- a/lib/react_client/react_interop.dart
+++ b/lib/react_client/react_interop.dart
@@ -104,17 +104,17 @@ class JsRef {
 ///
 /// See: <https://reactjs.org/docs/forwarding-refs.html>.
 ReactJsComponentFactoryProxy forwardRef(Function(Map props, Ref ref) wrapperFunction) {
-  var hoc = _jsForwardRef(allowInterop((JsMap props, Ref ref) {
-    var dartProps = new JsBackedMap.backedBy(props);
-
-    return wrapperFunction(dartProps, ref);
+  var hoc = _jsForwardRef(allowInterop((JsMap props, JsRef ref) {
+    final dartProps = JsBackedMap.backedBy(props);
+    final dartRef = Ref.fromJs(ref);
+    return wrapperFunction(dartProps, dartRef);
   }));
 
   return new ReactJsComponentFactoryProxy(hoc, shouldConvertDomProps: false);
 }
 
 @JS('React.forwardRef')
-external ReactClass _jsForwardRef(Function(JsMap props, Ref ref) wrapperFunction);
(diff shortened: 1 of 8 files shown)


lib/react_client/react_interop.dart
```diff
@@ -57,39 +57,43 @@ abstract class React {
 ///     }
 ///
 /// Learn more: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
-Ref<CurrentType> createRef<CurrentType>() {
-  return new Ref<CurrentType>();
+Ref<T> createRef<T>() {
```
kealjones-wk: :P

lib/react_client/react_interop.dart
```diff
@@ -57,39 +57,43 @@ abstract class React {
...
   /// A JavaScript ref object returned by [React.createRef].
-  JsRef jsRef;
+  final JsRef jsRef;
 
-  Ref() {
-    jsRef = React.createRef();
-  }
+  Ref() : jsRef = React.createRef();
+
+  Ref.fromJs(this.jsRef);
```
kealjones-wk: nice

kealjones-wk:
+10

lib/react_client/react_interop.dart
aaronlademann-wf: Feels very _meta_

lib/react_client/react_interop.dart
aaronlademann-wf: #supernit

The example in the doc comment for `createRef` shows left-side typing like this:

```dart
final Ref<BarComponent> barRef = createRef();
```

When I believe best practice would be something like:

```dart
final barRef = createRef<BarComponent>();
```

lib/react_client/react_interop.dart
```diff
@@ -57,39 +57,43 @@ abstract class React {
 ///     }
 ///
 /// Learn more: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
-Ref<CurrentType> createRef<CurrentType>() {
-  return new Ref<CurrentType>();
+Ref<T> createRef<T>() {
+  return new Ref<T>();
```
aaronlademann-wf: #supernit

The `new` keyword is no longer necessary

lib/react_client/react_interop.dart
```diff
@@ -57,39 +57,43 @@ abstract class React {
...
 /// When this is provided as the ref prop, a reference to the rendered component
 /// will be available via [current].
 ///
-/// See: <https://reactjs.org/docs/refs-and-the-dom.html#creating-refs>.
-class Ref<CurrentType> {
+/// See [createRef] for usage examples and more info.
+class Ref<T> {
   /// A JavaScript ref object returned by [React.createRef].
-  JsRef jsRef;
+  final JsRef jsRef;
```
aaronlademann-wf: nice nice nice nice nice

lib/react_client/react_interop.dart
```diff
@@ -183,10 +187,46 @@ class ReactClass {
...
+
+  /// A [Component2]-based component.
+  @protected
+  static const String component2 = '2';
+
+  /// Returns [ReactClass.dartComponentVersion] if [type] is the [ReactClass] for a Dart component
+  /// (a react-dart [ReactElement] or [ReactComponent]), and null otherwise.
+  @protected
+  static String fromType(dynamic type) {
+    // This check doesn't do much since ReactClass is an  an anonymous JS object,
```
aaronlademann-wf:
```suggestion
    // This check doesn't do much since ReactClass is an anonymous JS object,
```

test/factory/common_factory_tests.dart
```diff
@@ -228,6 +229,10 @@ void refTests(ReactComponentFactoryProxy factory, {void verifyRefValue(dynamic r
 
   test('forwardRef function passes a ref through a component to one of its children', () {
     var ForwardRefTestComponent = forwardRef((props, ref) {
+      // Extra type checking since JS refs being passed through
+      // aren't caught by built-in type checking.
+      expect(ref, isA<Ref>());
```
aaronlademann-wf: 🎉

lib/react_client/react_interop.dart
greglittlefield-wf: https://dart.dev/guides/language/effective-dart/design#prefer-type-annotating-public-fields-and-top-level-variables-if-the-type-isnt-obvious

> When in doubt, add a type annotation. Even when a type is obvious, you may still wish to explicitly annotate. If the inferred type relies on values or declarations from other libraries, you may want to type annotate your declaration so that a change to that other library doesn’t silently change the type of your own API without you realizing.

Fields are kind of a grey area, and we tend toward annotating them in our codebases, so that's why I went with this typing.

lib/react_client/react_interop.dart
greglittlefield-wf: Yeah at some point we should just go through and remove all of these

PR #215, opened 2019-10-02
CPLAT-7571 Remove on Keyword

joebingham-wk:
# Motivation
`TypedSnapshot uses` the on keyword to enforce that the mixin is only used with children of `Component2`. However, because `PropTypes` is implemented in `UiComponent2`, `TypedSnapshot` does not mix with `UiComponent2` well. This is related to https://github.com/dart-lang/sdk/issues/38098.

# Changes
- Remove the `on` keyword, and ultimately the dependency upon using the mixin with `Component2` based classes.

@aaronlademann-wf @greglittlefield-wf @kealjones-wk @sydneyjodon-wk

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -1174,11 +1174,14 @@ abstract class Component2 implements Component {
 ///       /* Include other standard component logic */
 ///
 ///     }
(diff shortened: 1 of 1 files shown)


greglittlefield-wf:
+10

PR #221, opened 2019-10-24
CPLAT-3874: Function Components

kealjones-wk:
## Motivation
Support Function Components in react-dart

## Changes
- Added `ReactDartFunctionComponentFactoryProxy` class
- Added `registerFunctionComponent` method
- Abstracted `ReactDartComponentFactoryProxy2`'s `build` and `generateExtendedJsProps`  logic into mixin for re-use with `ReactDartFunctionComponentFactoryProxy`
- Updated `isDartComponent` logic to support function components
- Added tests


lib/react.dart
```diff
@@ -20,6 +20,7 @@ export 'package:react/react_client/react_interop.dart' show forwardRef, createRe
 typedef Error PropValidator<TProps>(
     TProps props, String propName, String componentName, String location, String propFullName);
 
+typedef FunctionComponent = dynamic Function(Map props);
```
aaronlademann-wf: Should the return type be `ReactElement`? Or do function components support fragments as well?

lib/react.dart
```diff
@@ -1186,6 +1190,9 @@ mixin TypedSnapshot<TSnapshot> {
 
 /// Creates a ReactJS virtual DOM instance (`ReactElement` on the client).
 abstract class ReactComponentFactoryProxy implements Function {
+  /// The JS component factory used by this factory to build [ReactElement]s.
+  ReactJsComponentFactory reactComponentFactory;
```
aaronlademann-wf: should we name this something like `reactJsComponentFactory` instead?

lib/react_client.dart
```diff
@@ -734,8 +746,57 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+  ReactDartFunctionComponentFactoryProxy(reactFunction, [String componentName])
+      : this.reactFunction = reactFunction,
+        this.componentName = componentName,
+        this.reactComponentFactory = React.createFactory(reactFunction);
+
+  get type => reactFunction;
+}
+
+/// Creates a function component from the given [dartFunctionComponent] that can be used with React.
+JsFunctionComponent _wrapFunctionComponent(DartFunctionComponent dartFunctionComponent, {String componentName}) {
```
aaronlademann-wf: Is `componentName` analogous to the JS `static displayName` field? If so, can we name it `displayName` to align with the JS stuff?

lib/react_client.dart
```diff
@@ -734,8 +746,57 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+  assert(() {
+    componentName ??= getProperty(dartFunctionComponent, 'name');
```
aaronlademann-wf: I'm confused why this part would only get set in dev mode... what will the `componentName` be in production?

lib/react_client.dart
```diff
@@ -734,8 +746,57 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+/// Creates and returns a new [ReactDartComponentFactoryProxy] from the provided [dartFunctionComponent]
+/// which produces a new `JsFunctionComponent`.
+ReactDartFunctionComponentFactoryProxy _registerFunctionComponent(DartFunctionComponent dartFunctionComponent,
+        {String componentName}) =>
+    new ReactDartFunctionComponentFactoryProxy(
+        _wrapFunctionComponent(dartFunctionComponent, componentName: componentName));
+
 /// Creates and returns a new [ReactDartComponentFactoryProxy] from the provided [componentFactory]
-/// which produces a new JS [`ReactClass` component class](https://facebook.github.io/react/docs/top-level-api.html#react.createclass).
+/// which produces a new JS [`ReactClass` component class].
```
```suggestion
/// which produces a new JS `ReactClass` component class.
```

lib/react.dart
kealjones-wk: its typed as dynamic because dart doesn't support union types or inheriting primitive types, React16 allows for a lot of things be returned... string, list, numbers, fragment, react element, etc...

lib/react_client.dart
kealjones-wk: ferrr surrreee

lib/react.dart
kealjones-wk: So this was added to this class in order for it to be accessible by the `JsBackedMapsComponentFactoryBuilder` mixin that uses `on ReactComponentFactoryProxy` this field is implemented in almost every `ReactDart...FactoryProxy` type class that extends from `ReactComponentFactoryProxy` with this exact same type and name, changing it would be a breaking change?

lib/react_client.dart
kealjones-wk: This was done in order to reduce any unnecessary additional calls when in production. We do a similar thing with Component2's display name: https://github.com/cleandart/react-dart/blob/5.1.0-wip/lib/react.dart#L608-L615

It will be `Anonymous`:
Screen Shot 2019-10-24 at 4 14 09 PM

That is, unless one is provided via the `componentName`/`displayName` factory arg, then it will use that... which i am thinking in over_react, we will populate with the same string that we would use to generate the UiFactory.

js_src/_dart_helpers.js
```diff
@@ -10,10 +10,14 @@ var _reactDartContextSymbol = Symbol(_reactDartSymbolPrefix+'context');
...
 /// Used within Component2 error boundry methods to dartify the error argument.
 /// See: https://github.com/dart-lang/sdk/issues/36363
-function _throwErrorFromJS(error){
+function _throwErrorFromJS(error) {
   throw error;
 }
 
+/// A JS variable that can be used with dart interop in order to force returning a
+/// javascript `null`. This prevents dart2js from possibly converting dart `null` into `undefined`.
+var _jsNull = null;
```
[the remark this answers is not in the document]
kealjones-wk: b3931e9f830f2441981e9e14672dc5de

js_src/_dart_helpers.js
```diff
@@ -10,10 +10,14 @@ var _reactDartContextSymbol = Symbol(_reactDartSymbolPrefix+'context');
 /// A JS side function to allow Dart to throw an error from JS in order to catch it Dart side.
 /// Used within Component2 error boundry methods to dartify the error argument.
 /// See: https://github.com/dart-lang/sdk/issues/36363
-function _throwErrorFromJS(error){
+function _throwErrorFromJS(error) {
   throw error;
 }
 
+/// A JS variable that can be used with dart interop in order to force returning a
```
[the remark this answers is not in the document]
kealjones-wk: lol `Fart`? @greglittlefield-wf would be proud

lib/react_client.dart
```diff
@@ -11,6 +11,7 @@ import "dart:collection";
 import "dart:html";
 import 'dart:js';
 import 'dart:js_util';
+import 'dart:js_util' as js_util;
```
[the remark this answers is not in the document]
kealjones-wk: lol cause im used to using that library with the namespace for its autocomplete lol my bad.

lib/react_client.dart
```diff
@@ -592,7 +609,7 @@ abstract class _ReactDartInteropStatics2 {
 }
 
 /// Creates and returns a new [ReactDartComponentFactoryProxy] from the provided [componentFactory]
-/// which produces a new JS [`ReactClass` component class](https://facebook.github.io/react/docs/top-level-api.html#react.createclass).
+/// which produces a new JS `ReactClass` component class.
```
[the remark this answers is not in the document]
kealjones-wk: nope

lib/react_client.dart
```diff
@@ -611,7 +628,7 @@ ReactDartComponentFactoryProxy _registerComponent(
     contextKeys: componentInstance.contextKeys,
   );
 
-  /// Create the JS [`ReactClass` component class](https://facebook.github.io/react/docs/top-level-api.html#react.createclass)
+  /// Create the JS `ReactClass` component class
```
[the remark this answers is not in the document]
kealjones-wk: no again

lib/react_client/react_interop.dart
```diff
@@ -218,10 +219,13 @@ abstract class ReactDartComponentVersion {
   @protected
   static String fromType(dynamic type) {
     // This check doesn't do much since ReactClass is an anonymous JS object,
-    // but it lets us safely cast to ReactClass.
+    // but it lets us safely cast to ReactClass
```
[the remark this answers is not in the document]
kealjones-wk: lol i hate periods

test/factory/dart_function_factory_test.dart
```diff
@@ -0,0 +1,93 @@
...
+  setClientConfiguration();
+
+  group('ReactDartFunctionComponentFactoryProxy', () {
+    test('has a type corresponding to the backing JS Function', () {
+      expect(FunctionFoo.type, equals(FunctionFoo.reactFunction));
+    });
+
+    group('- common factory behavior -', () {
+      group('Function Component -', () {
+        group('- common factory behavior -', () {
```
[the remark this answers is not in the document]
kealjones-wk: inception

test/factory/dart_function_factory_test.dart
```diff
@@ -0,0 +1,93 @@
...
+          expect(ReactDartComponentVersion.fromType(react.div({}).type), isNull);
+          expect(ReactDartComponentVersion.fromType(JsFoo({}).type), isNull);
+          expect(ReactDartComponentVersion.fromType(FunctionFoo({}).type), ReactDartComponentVersion.component2);
+          expect(ReactDartComponentVersion.fromType(Foo({}).type), ReactDartComponentVersion.component);
+          expect(ReactDartComponentVersion.fromType(Foo2({}).type), ReactDartComponentVersion.component2);
+        });
+      });
+
+      group('test utils', () {
+        // todo move somewhere else
```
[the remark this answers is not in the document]
kealjones-wk: this is straight copy pasta from dart_factory_test.dart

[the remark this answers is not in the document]
aaronlademann-wf: Lets just remove it then

test/lifecycle_test.dart
```diff
@@ -429,6 +429,78 @@ main() {
...
+      });
+
+      group('', () {
+        ReactDartFunctionComponentFactoryProxy PropsTest;
+
+        setUpAll(() {
+          PropsTest = react.registerFunctionComponent((Map props) {
+            return [props['testProp'], props['children']];
+          });
+        });
```
aaronlademann-wf: You'll should add a `tearDownAll` that nulls `PropsTest`.

test/lifecycle_test.dart
```diff
@@ -429,6 +429,78 @@ main() {
...
+
+      group('recieves a JsBackedMap from the props argument', () {
+        var propTypeCheck;
+        ReactDartFunctionComponentFactoryProxy PropsArgTypeTest;
+        setUp(() {
+          PropsArgTypeTest = react.registerFunctionComponent((Map props) {
+            propTypeCheck = props;
+            return null;
+          });
+        });
```
aaronlademann-wf: You should add a `tearDown` that nulls `PropsArgTypeTest` and `propTypeCheck`.

lib/react.dart
```diff
@@ -20,7 +20,9 @@ export 'package:react/react_client/react_interop.dart' show forwardRef, createRe
 typedef Error PropValidator<TProps>(
     TProps props, String propName, String componentName, String location, String propFullName);
 
+typedef FunctionComponent = dynamic Function(Map props);
```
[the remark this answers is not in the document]
kealjones-wk: It doesn't, i updated the name here to be DartFunctionComponent then remove it from react_client.dart since it imports react.dart :)

lib/react_client.dart
```diff
@@ -148,10 +161,12 @@ class ReactDartComponentFactoryProxy2<TComponent extends Component2> extends Rea
   ReactDartComponentFactoryProxy2(ReactClass reactClass)
       : this.reactClass = reactClass,
         this.reactComponentFactory = React.createFactory(reactClass),
-        this.defaultProps = new JsBackedMap.fromJs(reactClass.defaultProps);
+        this.defaultProps = new JsBackedMap.fromJs(reactClass?.defaultProps);
```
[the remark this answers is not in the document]
kealjones-wk: if someone passes in `null`? lol ill remove this i think this was left over from before when i was considering just using `ReactDartComponentFactoryProxy2` instead of creating another factory.

lib/react_client.dart
```diff
@@ -734,8 +751,56 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+
+  /// The JS component factory used by this factory to build [ReactElement]s.
+  final ReactJsComponentFactory reactComponentFactory;
+
+  ReactDartFunctionComponentFactoryProxy(JsFunctionComponent reactFunction, [String displayName])
+      : this.reactFunction = reactFunction,
+        this.displayName = displayName,
+        this.reactComponentFactory = React.createFactory(reactFunction);
+
+  get type => reactFunction;
```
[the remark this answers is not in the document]
kealjones-wk: There is a comment inherited from `ReactComponentFactoryProxy`? "The type of component created by this factory." is the inherited comment enough or should i copy and paste it over too here?

lib/react_client.dart
```diff
@@ -734,8 +751,56 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+}
+
+/// Creates a function component from the given [dartFunctionComponent] that can be used with React.
+JsFunctionComponent _wrapFunctionComponent(DartFunctionComponent dartFunctionComponent, {String displayName}) {
+  assert(() {
+    displayName ??= getProperty(dartFunctionComponent, 'name');
+    return true;
+  }());
+  jsFunctionComponent(JsMap jsProps, [JsMap _legacyContext]) =>
+      (dartFunctionComponent(jsProps != null ? JsBackedMap.backedBy(jsProps) : JsBackedMap()) ?? jsNull);
```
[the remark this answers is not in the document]
kealjones-wk: in dart world, no user ever sees that argument, we have it here as the wrapper we pass to javascript so that dart doesnt panic when it gets 2 arguments from react instead of 1.

lib/react_client.dart
```diff
@@ -734,8 +751,56 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+  JsFunctionComponent interopFunction = allowInterop(jsFunctionComponent);
+  if (displayName != null) {
+    // This is a work-around to display the correct name in the React DevTools.
+    callMethod(getProperty(window, 'Object'), 'defineProperty', [
+      interopFunction,
+      'name',
```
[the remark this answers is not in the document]
kealjones-wk: Nope, this is actually modifying the `name` property of the javascript function. because it is just a js function it doesnt have special properties like react component classes do, so its just the generic `name`.

lib/react_client/react_interop.dart
```diff
@@ -7,6 +7,7 @@ library react_client.react_interop;
 
 import 'dart:html';
 
+import 'package:js/js_util.dart';
```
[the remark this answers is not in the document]
kealjones-wk: its not, js_util.dart has `getProperty` and a bunch of other nice interop helper functions that arent exported by js.dart. js.dart only really exports like `allowInterop` and the `@JS` annotation and then like 2 other methods.

js_src/_dart_helpers.js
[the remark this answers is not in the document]
kealjones-wk: Id also like to point out...
https://github.com/cleandart/react-dart/pull/221/files#diff-5ca8a047db3f6ee8d65a46bba4471236R162

lib/react_with_addons.js
```diff
@@ -155,7 +155,11 @@ var _reactDartContextSymbol = Symbol(_reactDartSymbolPrefix + 'context'); /// A
 
 function _throwErrorFromJS(error) {
   throw error;
-}
+} /// A JS variable that can be used with dart interop in order to force returning a
+/// javascript `null`. This prevents dart2js from possibly converting dart `null` into `undefined`.
```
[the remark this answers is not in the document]
kealjones-wk: webpack ¯\_(ツ)_/¯

js_src/_dart_helpers.js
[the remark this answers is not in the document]
joebingham-wk: See the comment above that about those files not being generated well ;)

lib/react_client.dart
```diff
@@ -35,6 +35,13 @@ export 'package:react/react.dart' show ReactComponentFactoryProxy, ComponentFact
 /// See: <https://facebook.github.io/react/docs/more-about-refs.html#the-ref-callback-attribute>
 typedef _CallbackRef<T>(T componentOrDomNode);
 
+/// The ReactJS function signature.
+///
+/// - [props] will always be supplied as the first argument
+/// - [legacyContext] has been deprecated and should not be used but remains for backward compatibility and is necessary
+/// to match darts generated call signature based on the number of args react provides.
```
greglittlefield-wf: #nit
```suggestion
/// to match Dart's generated call signature based on the number of args React provides.
```

lib/react_client.dart
```diff
@@ -181,6 +191,8 @@ class ReactDartComponentFactoryProxy2<TComponent extends Component2> extends Rea
   /// Returns a JavaScript version of the specified [props], preprocessed for consumption by ReactJS and prepared for
   /// consumption by the [react] library internals.
   static JsMap generateExtendedJsProps(Map props) {
+    if (props == null) return newObject();
```
greglittlefield-wf: Do we need this null-check here?

I would rather update any call-sites that might pass in null as opposed to opening up the possibility of something passing in null on accident, and silently failing instead of getting a null-error.

lib/react_client.dart
```diff
@@ -734,8 +746,54 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
   }
 }
 
+class ReactDartFunctionComponentFactoryProxy extends ReactComponentFactoryProxy
```
greglittlefield-wf: Needs a doc comment

lib/react_client.dart
```diff
@@ -151,7 +159,9 @@ class ReactDartComponentFactoryProxy2<TComponent extends Component2> extends Rea
         this.defaultProps = new JsBackedMap.fromJs(reactClass.defaultProps);
 
   ReactClass get type => reactClass;
+}
 
+mixin JsBackedMapsComponentFactoryBuilder on ReactComponentFactoryProxy {
```
greglittlefield-wf: Needs a doc comment

lib/react_client.dart
```diff
@@ -181,6 +191,8 @@ class ReactDartComponentFactoryProxy2<TComponent extends Component2> extends Rea
   /// Returns a JavaScript version of the specified [props], preprocessed for consumption by ReactJS and prepared for
   /// consumption by the [react] library internals.
   static JsMap generateExtendedJsProps(Map props) {
```
greglittlefield-wf: This was previously a static member on `ReactDartComponentFactoryProxy2`; will moving it into this class break over_react? We might just want to proxy it on that class for that reason, and plus, it'd be a little more convenient to get it from there instead of knowing to go get it from the mixin.

lib/react.dart
greglittlefield-wf: This is more of an implementation detail that we might not always want (for instance, we could just use `React.createClass` directly).

What about adding this onto `ReactComponentFactoryProxy` as an abstract getter, and then the classes that mix it in will override it with their current implementations?

lib/react_client.dart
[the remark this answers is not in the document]
greglittlefield-wf: That should be good. This needs an `@override` annotation, though.

lib/react_client.dart
```diff
@@ -734,8 +751,56 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+        this.displayName = displayName,
+        this.reactComponentFactory = React.createFactory(reactFunction);
+
+  get type => reactFunction;
+}
+
+/// Creates a function component from the given [dartFunctionComponent] that can be used with React.
+JsFunctionComponent _wrapFunctionComponent(DartFunctionComponent dartFunctionComponent, {String displayName}) {
+  assert(() {
+    displayName ??= getProperty(dartFunctionComponent, 'name');
```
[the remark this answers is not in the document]
greglittlefield-wf: This could use a similar comment to the one on existing code https://github.com/cleandart/react-dart/blob/a3b562304c2f73f95ee6253ef1f0f62141e622c5/lib/react.dart#L603-L607 to make it clear why we're doing this.

lib/react_client.dart
```diff
@@ -734,8 +746,54 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+    return true;
+  }());
+  jsFunctionComponent(JsMap jsProps, [JsMap _legacyContext]) =>
+      (dartFunctionComponent(jsProps != null ? JsBackedMap.backedBy(jsProps) : JsBackedMap()) ?? jsNull);
```
greglittlefield-wf: When would `jsProps` be null?

greglittlefield-wf: We should have a line comment linking to the undefined/null issue as an explanation for `?? jsNull`

greglittlefield-wf: Unnecessary parens
```suggestion
      dartFunctionComponent(jsProps != null ? JsBackedMap.backedBy(jsProps) : JsBackedMap()) ?? jsNull;
```

lib/react_client.dart
```diff
@@ -734,8 +746,54 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
+  JsFunctionComponent interopFunction = allowInterop(jsFunctionComponent);
+  if (displayName != null) {
+    // This is a work-around to display the correct name in the React DevTools.
+    callMethod(getProperty(window, 'Object'), 'defineProperty', [
```
greglittlefield-wf: Can we use a JS-interop binding for this? Those read better and have better static analysis.
```dart
@JS('Object.defineProperty')
external void _defineProperty(dynamic object, String propertyName, dynamic descriptor);
```

test/factory/dart_function_factory_test.dart
```diff
@@ -0,0 +1,90 @@
...
+      expect(FunctionFoo.type, equals(FunctionFoo.reactFunction));
+    });
+
+    group('Function Component -', () {
+      group('- common factory behavior -', () {
+        commonFactoryTests(FunctionFoo, isFunctionComponent: true);
+      });
+    });
+
+    group('utils', () {
```
greglittlefield-wf: These utils tests (minus the addition of the function component cases) are duplicated from another file. Could we split these utils out into a separate file? Sorry I didn't do that in the first place.

test/lifecycle_test.dart
```diff
@@ -429,6 +429,87 @@ main() {
...
+
+      group('recieves a JsBackedMap from the props argument', () {
+        var propTypeCheck;
+        ReactDartFunctionComponentFactoryProxy PropsArgTypeTest;
+
+        setUp(() {
+          PropsArgTypeTest = react.registerFunctionComponent((Map props) {
+            propTypeCheck = props;
+            return null;
+          });
```
greglittlefield-wf: `propTypeCheck` should also be nulled out here. to ensure the test starts with a clean slate

test/lifecycle_test.dart
```diff
@@ -429,6 +429,87 @@ main() {
...
+        });
+
+        tearDownAll(() {
```
greglittlefield-wf: #nit this teardown isn't really necessary since `PropsArgTypeTest` is always overwritten in setUp

test/lifecycle_test.dart
```diff
@@ -429,6 +429,87 @@ main() {
...
+          PropsArgTypeTest = null;
+        });
+
+        test('when provided with an empty dart map', () {
+          react_dom.render(PropsArgTypeTest({}), mountNode);
+          expect(propTypeCheck, isA<JsBackedMap>());
+        });
+
+        test('when provided null', () {
```
greglittlefield-wf: Same question as above: will this case ever be encountered?

dart_test.yaml
```diff
@@ -0,0 +1,3 @@
+tags:
+  # Tags for functional components
+  functionComponent:
```
greglittlefield-wf: #nit Is this tags necessary? Seems like a test name filter would be sufficient

lib/react_client.dart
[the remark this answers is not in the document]
kealjones-wk: after re-thinking about this... we chose to not use runtimeType due to dart2js de-optimization, but here its just a property getter, that should be pretty fast, do we even need the assert? ive confirmed it still works with -r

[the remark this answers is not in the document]
kealjones-wk: > However the dart2js deoptimization concerns are less about runtime speed at the call site and more about increased size of code output. I'll try to find a reference for this and post back.

I believe that the output code from a getProperty is pretty minor, in fact i think it ends up becoming like `self.dartFunctionComponent.name` ? or something like that hence my thought on removing the assert.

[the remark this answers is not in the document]
kealjones-wk: > Actually, it looks like `toString` might be an exception: [dart-lang/linter#1475](https://github.com/dart-lang/linter/issues/1475) so maybe we don't need the assert after all!

also that sounds super promising! do you mean we could potentially remove the assert from component2 as well since we use the toString?

lib/react_client.dart
kealjones-wk: c75

lib/react_client.dart
```diff
@@ -626,116 +893,8 @@ ReactDartComponentFactoryProxy _registerComponent(
   return new ReactDartComponentFactoryProxy(reactComponentClass);
 }
 
-class ReactJsContextComponentFactoryProxy extends ReactJsComponentFactoryProxy {
```
kealjones-wk: This wasn't removed, it was just moved up to group it with other [FactoryProxy]s

lib/react_client.dart
```diff
@@ -776,60 +935,12 @@ ReactDartComponentFactoryProxy2 _registerComponent2(
   return new ReactDartComponentFactoryProxy2(reactComponentClass);
 }
 
-/// Creates ReactJS [ReactElement] instances for DOM components.
-class ReactDomComponentFactoryProxy extends ReactComponentFactoryProxy {
```
kealjones-wk: This wasn't removed, it was just moved up to group it with other [FactoryProxy]s

example/test/function_component_test.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'package:react/react.dart' as react;
+import 'package:react/react_dom.dart' as react_dom;
+import 'package:react/react_client.dart';
+
+import 'react_test_components.dart';
+
+void main() {
+  setClientConfiguration();
+  var inputValue = 'World';
+  // TODO: replace this with hooks/useState when they are added.
```
aaronlademann-wf: Can we add a note / AC item to make sure this gets addressed via #223?

example/test/react_test_components.dart
```diff
@@ -282,9 +282,9 @@ class _NewContextProviderComponent extends react.Component2 {
   }
 
   render() {
-    Map provideMap = {'renderCount': this.state['renderCount']};
+    Map<String, dynamic> provideMap = {'renderCount': this.state['renderCount']};
```
aaronlademann-wf: I'm not 100% sure for the reason this was changed, but the `<String, dynamic>` typing can be inferred:

```suggestion
    final provideMap = {'renderCount': this.state['renderCount']};
```

example/test/react_test_components.dart
```diff
@@ -282,9 +282,9 @@ class _NewContextProviderComponent extends react.Component2 {
...
 
-    Map complexValues = {
+    Map<String, dynamic> complexValues = {
```
aaronlademann-wf: I'm not 100% sure for the reason this was changed, but the `<String, dynamic>` typing can be inferred:

```suggestion
    final complexValues = {
```

js_src/_dart_helpers.js
```diff
@@ -131,7 +137,7 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
   }
 
   // Delete methods that the user does not want to include (such as error boundary event).
-  jsConfig.skipMethods.forEach((method) => {
+  jsConfig.skipMethods.forEach(method => {
```
aaronlademann-wf: I'm unfamiliar with this syntax - is it kosher to ditch the parentheses? 

Out of curiosity, what code change on the Dart side resulted in this change?

lib/react.dart
```diff
@@ -20,7 +20,12 @@ export 'package:react/react_client/react_interop.dart' show forwardRef, createRe
 typedef Error PropValidator<TProps>(
     TProps props, String propName, String componentName, String location, String propFullName);
 
+/// A React component declared using a function that takes in [props] and returns rendered output.
+///
+/// See <https://facebook.github.io/react/docs/components-and-props.html#functional-and-class-components>.
+typedef DartFunctionComponent = dynamic Function(Map props);
```
aaronlademann-wf: I think I already asked this - but its not notated in the comment anywhere... what is/are the expected return type(s)? Is `dynamic` the best we can do for the return signature?

lib/react.dart
```diff
@@ -1186,6 +1194,12 @@ mixin TypedSnapshot<TSnapshot> {
 
 /// Creates a ReactJS virtual DOM instance (`ReactElement` on the client).
 abstract class ReactComponentFactoryProxy implements Function {
+  /// The JS component factory used by this factory to build [ReactElement]s.
+  ///
+  /// Deprecated: [ReactComponentFactoryProxy]s now use inline [React.createElement()].
```
aaronlademann-wf: This deprecation message doesn't really spell out what should be used instead.  Can we add an example of what consumers should do if they are using this currently?

lib/react.dart
```diff
@@ -1766,6 +1780,11 @@ ComponentRegistrar2 registerComponent2 = (
   throw new Exception('setClientConfiguration must be called before registerComponent.');
 };
 
+/// Registers [componentFactory] on both client and server.
```
aaronlademann-wf: This doc comment looks like copypasta from `ComponentRegistrar`. At the very least, the bit about it registering stuff for server side rendering should be removed.  It would be even better if the comment fully explained how it should be used.

test/factory/dart_function_factory_test.dart
```diff
@@ -0,0 +1,30 @@
...
+        commonFactoryTests(FunctionFoo, isFunctionComponent: true);
+      });
+    });
+  });
+}
+
+final FunctionFoo = react.registerFunctionComponent(_FunctionFoo);
+
+_FunctionFoo(Map props) {
+  return react.div({});
```
aaronlademann-wf: Do the `commonFactoryTests` not test any sort of prop forwarding? Seems like the `props` value is completely ignored here - which I would expect to cause at least 1 failing test.

js_src/_dart_helpers.js
```diff
@@ -1,19 +1,25 @@
...
 
 /// A JS side function to allow Dart to throw an error from JS in order to catch it Dart side.
 /// Used within Component2 error boundry methods to dartify the error argument.
 /// See: https://github.com/dart-lang/sdk/issues/36363
-function _throwErrorFromJS(error){
+function _throwErrorFromJS(error) {
   throw error;
 }
 
+/// A JS variable that can be used with Fart interop in order to force returning a
```
joebingham-wk: :sigh:
```suggestion
/// A JS variable that can be used with Dart interop in order to force returning a
```

lib/react_client/react_interop.dart
```diff
@@ -474,6 +478,11 @@ class JsError {
   external JsError(message);
 }
 
+/// A JS variable that can be used with dart interop in order to force returning a javascript `null`.
```
joebingham-wk: #nit
```suggestion
/// A JS variable that can be used with Dart interop in order to force returning a JavaScript `null`.
```

lib/react_client/react_interop.dart
```diff
@@ -474,6 +478,11 @@ class JsError {
...
+/// Use this if dart2js is possibly converting dart `null` into `undefined`.
```
joebingham-wk: #nit
```suggestion
/// Use this if dart2js is possibly converting Ddart `null` into `undefined`.
```

test/util_test.dart
```diff
@@ -0,0 +1,71 @@
...
+
+    test('isDartComponent', () {
+      expect(isDartComponent(react.div({})), isFalse);
+      expect(isDartComponent(JsFoo({})), isFalse);
+      expect(isDartComponent(Foo({})), isTrue);
+      expect(isDartComponent(Foo2({})), isTrue);
+    });
+  });
+}
+
```
joebingham-wk: #nit extra line
```suggestion
```

test/util_test.dart
```diff
@@ -0,0 +1,71 @@
...
+class _Foo2 extends react.Component2 {
+  @override
+  render() => react.div({});
+}
+
+final JsFoo = ReactJsComponentFactoryProxy(React.createClass(ReactClassConfig(
+  displayName: 'JsFoo',
+  render: allowInterop(() => react.div({})),
+)));
+
```
joebingham-wk: #nit extra line
```suggestion
```

js_src/_dart_helpers.js
kealjones-wk: nope

example/test/function_component_test.dart
kealjones-wk: @sydneyjodon-wk already addressed this in her pr

example/test/react_test_components.dart
kealjones-wk: This is changed because its inferred as <String, int> but the example changes the `value` for the "complex" values to <String, Map<String, dynamic>> and since its casted to int it will throw if you don't initially type it as <String, dynamic>

js_src/_dart_helpers.js
kealjones-wk: Yeah it it's kosher, because it gets babelified this get de-sugared. its syntactic sugar for that is part of ES6, if there is only a single argument, you can even leave off the parenthesis around the argument, it's mainly just formatting change, nothing functionally besides `_jsNull` changed here.

lib/react.dart
kealjones-wk: Yeah this is similar to the return of render, it can technically be a number of different things...
Typescripts render / return type is `ReactNode`:

```typescript

    type ReactText = string | number;
    type ReactChild = ReactElement | ReactText;

    interface ReactNodeArray extends Array<ReactNode> {}
    type ReactFragment = {} | ReactNodeArray;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

so something like:
```
/* union ReactReturn = ReactElement | List<ReactReturn> | String | num | bool | Null; */

Comment (aaronlademann-wf) [reply lib/react.dart]: cool

File: test/factory/dart_function_factory_test.dart (hunk 45 above)
Comment (kealjones-wk) [reply test/factory/dart_function_factory_test.dart]: yeah they dont, this is basicly copypasta from `dart_factory_test.dart`
```
final Foo = react.registerComponent(() => new _Foo()) as ReactDartComponentFactoryProxy;

class _Foo extends react.Component {
  @override
  render() => react.div({});
}

final Foo2 = react.registerComponent(() => new _Foo2()) as ReactDartComponentFactoryProxy2;

class _Foo2 extends react.Component2 {
  @override
  render() => react.div({});
}
```

the props are tested pretty exhaustively in `lifecycle_test.dart` though

File: example/test/react_test_components.dart (hunk 40 above)
Comment (kealjones-wk) [reply example/test/react_test_components.dart]: https://github.com/cleandart/react-dart/pull/221/files#r342151687

File: lib/react.dart [hunk 51]
```diff
@@ -1766,6 +1791,18 @@ ComponentRegistrar2 registerComponent2 = (
   throw new Exception('setClientConfiguration must be called before registerComponent.');
 };
 
+/// Registers [componentFactory] on client.
+///
+/// Example:
+/// ```
+///   var myFunctionComponent = registerFunctionComponent((Map props) {
+///         return ['I am a function component', ...props.children];
+///       });
```
joebingham-wk: #nit I could be wrong, but wouldn't the ending `});` usually has less indentation?
```suggestion
///   });
```

lib/react_client.dart
```diff
@@ -135,69 +162,371 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
...
+}
+
+class ReactJsContextComponentFactoryProxy extends ReactJsComponentFactoryProxy {
+  /// The JS class used by this factory.
+  @override
+  final ReactClass type;
+  final bool isConsumer;
+  final bool isProvider;
+
+  /// Deprecated: Use [React.createElement()] instead and pass in [type] as
```
joebingham-wk: The first line is usually a description of what the API is. Could we explicitly add that line and then add the deprecation below it?

example/test/react_test_components.dart
aaronlademann-wf: @kealjones-wk I'm  not seeing this being inferred as `<String, int>`...

image

lib/react.dart
aaronlademann-wf: If the triple backticks are being used, all of the indentation (lines 1798 and 1799 also) are off...

example/test/react_test_components.dart
kealjones-wk: at runtime in dart2js

greglittlefield-wf: That actually seems like a Dart bug... We should report this if it hasn't been reported already.

test/util_test.dart
```diff
@@ -0,0 +1,69 @@
...
+  group('util test', () {
+    test('ReactDartComponentVersion.fromType', () {
+      expect(ReactDartComponentVersion.fromType(react.div({}).type), isNull);
+      expect(ReactDartComponentVersion.fromType(JsFoo({}).type), isNull);
+      expect(ReactDartComponentVersion.fromType(Foo({}).type), ReactDartComponentVersion.component);
+      expect(ReactDartComponentVersion.fromType(Foo2({}).type), ReactDartComponentVersion.component2);
+      expect(ReactDartComponentVersion.fromType(FunctionFoo({}).type), ReactDartComponentVersion.component2);
+    });
+
+    test('isDartComponent2', () {
```
greglittlefield-wf: These three tests are missing test cases for `FunctionFoo`

lib/react_client.dart
```diff
@@ -135,69 +162,371 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
...
-    if (childrenArgs.isEmpty) {
-      children = childrenArgs;
-    } else if (childrenArgs.length == 1) {
-      final singleChild = listifyChildren(childrenArgs[0]);
-      if (singleChild is List) {
-        children = singleChild;
+  static JsMap generateExtendedJsProps(Map props) => _generateJsProps(props);
+}
+
+JsMap _generateJsProps(Map props,
```
greglittlefield-wf: #nit any reason this is split out from `JsBackedMapComponentFactoryMixin._generateJsProps`?

lib/react_client.dart
```diff
@@ -135,69 +162,371 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
...
+    {bool convertEventHandlers = false,
```
greglittlefield-wf: These arguments don't seem to be used at all; are they needed?

lib/react.dart
```diff
@@ -1186,6 +1194,23 @@ mixin TypedSnapshot<TSnapshot> {
...
+  /// ```
+  ///   YourReactComponentFactoryProxy.reactComponentFactory(props, children);
+  /// ```
+  ///
+  /// After:
+  /// ```
+  ///   React.createElement(YourReactComponentFactoryProxy.type, props, children);
+  /// ```
+  @Deprecated('6.0.0')
+  ReactJsComponentFactory reactComponentFactory;
```
greglittlefield-wf: Can this be removed with the latest changes? This doesn't seem to be used anywhere, and shouldn't be added to the public API if not necessary

lib/react.dart
```diff
@@ -1766,6 +1791,18 @@ ComponentRegistrar2 registerComponent2 = (
   throw new Exception('setClientConfiguration must be called before registerComponent.');
 };
 
+/// Registers [componentFactory] on client.
+///
+/// Example:
+/// ```
+///   var myFunctionComponent = registerFunctionComponent((Map props) {
```
greglittlefield-wf: Should the example show declaring it as a separate function or passing in `displayName` so that there's a displayName built in?

lib/react_client.dart
```diff
@@ -135,69 +162,371 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
...
+  // dart2js uses null and undefined interchangeably, meaning returning `null` from dart
+  // may show up in js as `undefined`, ReactJS doesnt like that and expects a js `null` to be returned,
+  // and throws if it gets `undefined`. `jsNull` is an interop variable that holds a JS `null` value
+  // to force `null` as the return value if user returns a Dart `null`.
+  // See: https://github.com/dart-lang/sdk/issues/27485
+  jsFunctionComponent(JsMap jsProps, [JsMap _legacyContext]) =>
+      dartFunctionComponent(JsBackedMap.backedBy(jsProps)) ?? jsNull;
+  JsFunctionComponent interopFunction = allowInterop(jsFunctionComponent);
+  if (displayName != null) {
+    // This is a work-around to display the correct name in the React DevTools.
```
greglittlefield-wf: Can we add a test for this and the `getProperty` logic above?

lib/react_client.dart
```diff
@@ -135,69 +162,371 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
...
+
+/// Converts a list of variadic children arguments to children that should be passed to ReactJS.
+///
+/// Returns:
+///
+/// - `null` if there are no args and [shouldAlwaysBeList] is false
+/// - `[]` if there are no args and [shouldAlwaysBeList] is true
+/// - the single child if only one was specified
+/// - otherwise, the same list of args, will all top-level children validated
+dynamic _generateChildren(List childrenArgs, {bool shouldAlwaysBeList = false}) {
```
greglittlefield-wf: This seems like it's just an improvement to the existing `_convertArgsToChildren`; could we combine the two implementations into one to reduce duplication of code? I'd vote for keeping the name `_convertArgsToChildren` since it's a little more descriptive.

example/test/react_test_components.dart
kealjones-wk: Sorry, so this issue was ACTUALLY due to providing a generic type to `createContext` that was typed as `Map<String, dynamic>` and then using Map as the type, it used `Map<dynamic, dynamic>` hence the issue i ran into, fix was to type these two vars to finals and then update the generic to just be `Map`

lib/react_client.dart
kealjones-wk: yeah so i originally replaced every FactoryProxy with my _generateChildren and _generateProps, and they worked for the most part (all tests in this repo pass) but in over_react there are like 2 tests that fail cause one expects a `empty` but gets `[null]` and i didnt want to deal with it so i removed it all thinking you probably wouldn't even like that much code changes lol

lib/react_client.dart
kealjones-wk: https://github.com/cleandart/react-dart/pull/221#discussion_r342731336

lib/react_client.dart
kealjones-wk: I had originally set out in this PR to update all function factories to be able to use `_generateChildren` and `_generateJsProps` in order to reduce code duplication, it works, all the tests will pass but then in OverReact 2 or so tests fail because they expect `empty` but get `[null]` this was discovered on friday and rather than fix it i opted to revert it because of another issue with something else i previously did in this PR. so here we are. I could probably pursue it again and find the issue, it is probably something silly.

lib/react.dart
kealjones-wk: it would be a breaking change to remove that public field, hence the deprecation... or would it not be breaking for some reason?

lib/react_client.dart
kealjones-wk: https://github.com/cleandart/react-dart/pull/221#discussion_r342769366

lib/react.dart
kealjones-wk: @greglittlefield-wf ^

aaronlademann-wf: I agree with Keal - this would be a breaking change.

lib/react_client.dart
```diff
@@ -35,6 +35,13 @@ export 'package:react/react.dart' show ReactComponentFactoryProxy, ComponentFact
 /// See: <https://facebook.github.io/react/docs/more-about-refs.html#the-ref-callback-attribute>
 typedef _CallbackRef<T>(T componentOrDomNode);
 
+/// The ReactJS function signature.
```
aaronlademann-wf: What does this mean?

test/factory/dart_function_factory_test.dart
```diff
@@ -0,0 +1,52 @@
+// ignore_for_file: deprecated_member_use_from_same_package
+// ignore_for_file: invalid_use_of_protected_member
+import 'package:js/js_util.dart';
```
joebingham-wk: #nit Rogue import placement.
```
@TestOn('browser')
import 'package:js/js_util.dart';
import 'package:react/react.dart' as react;
import 'package:react/react_client.dart';
import 'package:test/test.dart';

import 'common_factory_tests.dart';
```

lib/react_client.dart
kealjones-wk: that typedef is the function signature required to pass to react js? I dont fully understand the question 😬

aaronlademann-wf:
+1 refresh

lib/react.dart
kealjones-wk: wait a second, this was Added, gregs probably correct, i added this here before i did a few other things, i think this should be removed. and it wouldnt be a break as it exists on all the other proxys themselves.

kealjones-wk: this was already removed, this is outdated...

kealjones-wk:
+10

test/factory/dart_function_factory_test.dart
```diff
@@ -0,0 +1,54 @@
...
+      expect(FunctionFoo.type, equals(FunctionFoo.reactFunction));
+    });
+
+    group('Function Component -', () {
+      group('- common factory behavior -', () {
+        commonFactoryTests(FunctionFoo, isFunctionComponent: true);
+      });
+
+      group('displayName', () {
+        test('gets automaticly populated when not provided', () {
```
greglittlefield-wf: #nit typo
```suggestion
        test('gets automatically populated when not provided', () {
```

lib/react_client.dart
```diff
@@ -706,31 +758,10 @@ class ReactJsComponentFactoryProxy extends ReactComponentFactoryProxy {
...
-      _convertRefValue(potentiallyConvertedProps);
-    } else {
-      potentiallyConvertedProps = props;
-    }
-
-    // jsifyAndAllowInterop also handles converting props with nested Map/List structures, like `style`
-    return factory(jsifyAndAllowInterop(potentiallyConvertedProps), children);
+    dynamic children = _generateChildren(childrenArgs, shouldAlwaysBeList: alwaysReturnChildrenAsList);
+    JsMap convertedProps = _generateJsProps(props,
+        convertEventHandlers: shouldConvertDomProps || props['ref'] != null, convertCallbackRefValue: false);
```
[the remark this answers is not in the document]
kealjones-wk: mainly because it was in the logic before... but lets try it without!

test/factory/dart_function_factory_test.dart
kealjones-wk: embarrassing

PR #223, opened 2019-10-30
CPLAT-8034 Implement/Expose useState Hook

sydneyjodon-wk:
## Motivation
Support useState hook in react-dart

## Changes
- Added `StateHook` class
- Added `useState` and `useStateLazy` functions
- Wrote tests.


example/test/react_test_components.dart
```diff
@@ -392,16 +392,6 @@ class _NewContextTypeConsumerComponent extends react.Component2 {
   }
 }
 
-var helloGregFunctionComponent = react.registerFunctionComponent(HelloGreg, displayName: 'HelloGreg');
-
-HelloGreg(Map props) {
```
kealjones-wk: why you take away `HelloGreg`!?

lib/react_client/react_interop.dart
```diff
@@ -105,7 +108,7 @@ class JsRef {
 ///
 /// See: <https://reactjs.org/docs/forwarding-refs.html>.
 ReactJsComponentFactoryProxy forwardRef(Function(Map props, Ref ref) wrapperFunction) {
-  var hoc = _jsForwardRef(allowInterop((JsMap props, JsRef ref) {
+  var hoc = React.forwardRef(allowInterop((JsMap props, JsRef ref) {
```
kealjones-wk: I like this better. @greglittlefield-wf do you see any possible issues here?

lib/hooks.dart
```diff
@@ -0,0 +1,92 @@
...
+///
+///       return react.div({}, [
+///         count.value,
+///         react.button({'onClick': (_) => count.set(0)}, ['Reset']),
+///         react.button({'onClick': (_) => count.setTx((prev) => prev + 1)}, ['+']),
+///       ]);
+///     }
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#lazy-initial-state>.
+StateHook<T> useStateInit<T>(T init()) => new StateHook.init(init);
```
kealjones-wk: Just for funzies couldn't we just rename `StateHook` to `useState` (ignore that it goes against convention) then we wouldn't need the class and two additional functions `useState` and `useStateInit` it would just be the two `useState` constructors... @greglittlefield-wf thoughts? i know that could be gross but it technically keeps the typing and all that

example/test/react_test_components.dart
kealjones-wk: 3f98ri

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// When called inside a [DartFunctionComponent], adds a local state to the component.
+///
+/// Returns a [StateHook] with [StateHook.value] initialized to [initialValue].
+///
+/// __Example__:
+///
+///     UseStateTestComponent(Map props) {
+///       final count = react.useState(0);
```
aaronlademann-wf: This example implies that the `package:react/react.dart` import is namespaced as `react`.  That's ok with me - as long as the example shows that namespaced import.

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+///
+///       return react.div({}, [
+///         count.value,
+///         react.button({'onClick': (_) => count.set(0)}, ['Reset']),
+///         react.button({'onClick': (_) => count.setTx((prev) => prev + 1)}, ['+']),
```
aaronlademann-wf: Is this also possible without using `setTx`?

```dart
react.button({'onClick': (_) => count.set(count.value + 1)}, ['+']),
```

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+///       ]);
+///     }
+///
+/// Learn more: <https://reactjs.org/docs/hooks-state.html>.
+StateHook<T> useState<T>(T initialValue) => new StateHook(initialValue);
```
aaronlademann-wf: #nit unnecessary `new` keyword

```suggestion
StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
```

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+
+/// When called inside a [DartFunctionComponent], adds a local state to the component.
+///
+/// Returns a [StateHook] with [StateHook.value] initialized to the return value of [init].
```
```dart
/// Adds local state to a [DartFunctionComponent] 
/// by returning a [StateHook] with [StateHook.value] initialized to the return value of [init].
```

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+///     UseStateTestComponent(Map props) {
+///       final count = react.useStateInit(() {
+///         var initialState = someExpensiveComputation(props);
+///         return initialState;
+///       }));
+///
+///       return react.div({}, [
+///         count.value,
+///         react.button({'onClick': (_) => count.set(0)}, ['Reset']),
+///         react.button({'onClick': (_) => count.setTx((prev) => prev + 1)}, ['+']),
```
aaronlademann-wf: Is this also possible without using `setTx`?

```dart
react.button({'onClick': (_) => count.set(count.value + 1)}, ['+']),
```

lib/react_client/react_interop.dart
greglittlefield-wf: Nope, looks good to me!

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
+@JS()
+library hooks;
```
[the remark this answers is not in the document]
greglittlefield-wf: After going back and forth on it, I suggested this offline to makes things a little more organized as opposed to just lumping more things into `react.dart`. Also, since they're APIs that will only be used sometimes, I figured it made sense for them to be their own self-contained entrypoint.

But, I could be convinced to move them into `react.dart`.

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+  /// The second item in the pair returned by [React.useState].
+  void Function(dynamic) _setValue;
+
+  StateHook(T initialValue) {
+    final result = React.useState(initialValue);
+    _value = result[0];
+    _setValue = result[1];
+  }
+
+  StateHook.init(T init()) {
```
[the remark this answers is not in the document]
greglittlefield-wf: No, this allows you to specify an initial value that's lazy-computed. We should link to these docs in the doc comment: https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

I think `StateHook.lazy` might be a better name for this constrructor

(Sorry, a lot of these bad names are from my spike, sorry Sydney 😅)

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  /// Updates [value] to [newValue].
+  ///
+  /// See: <https://reactjs.org/docs/hooks-state.html#updating-state>.
+  void set(T newValue) => _setValue(newValue);
+
+  /// Updates [value] to the return value of [computeNewValue].
+  ///
+  /// See: <https://reactjs.org/docs/hooks-reference.html#functional-updates>.
+  void setTx(T computeNewValue(T oldValue)) => _setValue(allowInterop(computeNewValue));
```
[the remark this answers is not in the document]
greglittlefield-wf: It was mean to be "transactional", but it was clear when I had `setStateTx` that that wasn't clear haha...

Perhaps `setWithUpdater` would be better? Or even just `update`? But with the latter it might not be the immediately obvious the difference between `set` and `update`.

Thoughts? I think that for these kinds of naming decisions, we should gather input from more rather than fewer people. @joebingham-wk @kealjones-wk

lib/hooks.dart
greglittlefield-wf: I would rather not stray against the conventions, since it would be confusing for consumers, especially if they used it as a type. Also, it would prevent them from using these as tearoffs, which is minor but could cause issues.

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+///
+///       return react.div({}, [
+///         count.value,
+///         react.button({'onClick': (_) => count.set(0)}, ['Reset']),
+///         react.button({'onClick': (_) => count.setTx((prev) => prev + 1)}, ['+']),
+///       ]);
+///     }
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#lazy-initial-state>.
+StateHook<T> useStateInit<T>(T init()) => new StateHook.init(init);
```
greglittlefield-wf: See above comment: we probably should name this something like `useStateLazy` to match.

lib/hooks.dart
```diff
@@ -0,0 +1,97 @@
...
+/// * Only call Hooks at the top level.
+/// * Only call Hooks from inside a [DartFunctionComponent].
+///
+/// Learn more: <https://reactjs.org/docs/hooks-state.html>.
+class StateHook<T> {
+  /// The first item of the pair returned by [React.useState].
+  T _value;
+
+  /// The second item in the pair returned by [React.useState].
+  void Function(dynamic) _setValue;
```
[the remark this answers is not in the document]
greglittlefield-wf: You might be right... my brain hurts when thinking about Dart 2 function types haha, so I'm not sure.

Works for me so long as there are no type errors in DDC or dart2js!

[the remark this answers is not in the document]
sydneyjodon-wk: Changing this type breaks the tests for `setTx`. I think it's because `_setValue` is the updater function returned by the js useState and can take an argument of type `T newValue` or type `T computeNewValue(T oldValue)`.

lib/hooks.dart
sydneyjodon-wk: You're right, I just wrote it that way because that was the example of functional updates in the React docs:
https://reactjs.org/docs/hooks-reference.html#functional-updates
Should I change it to `set` and use a different example for `setTx`?

lib/hooks.dart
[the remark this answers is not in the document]
kealjones-wk: Im plus1 on the `StateHook.lazy` and using a function like `setLazy`

lib/hooks.dart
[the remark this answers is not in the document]
kealjones-wk: I feel the same it makes sense to have them seperate but i also enjoy the convenience of getting them when importing `react.dart`... plus they would get tree shaken out if they weren't used right?

lib/hooks.dart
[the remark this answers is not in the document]
kealjones-wk: i think `setWithUpdater` makes a lot of sense as it correlates to component2's new `setStateWithUpdater`, also we never made a codemod for that did we .... should we?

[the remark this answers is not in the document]
aaronlademann-wf: I like `setWithUpdater`

lib/hooks.dart
aaronlademann-wf: I think so, yeah.

aaronlademann-wf: But we could also just make a follow-up ticket to improve the examples.

lib/hooks.dart
```diff
@@ -0,0 +1,101 @@
...
+import 'package:js/js.dart';
+import 'package:react/react.dart';
+import 'package:react/react_client/react_interop.dart';
+
+/// The return value of [useState].
+///
+/// The current value of the state is available via [value] and
+/// functions to update it are available via [set] and [setWithUpdater].
+///
+/// Note are two rules for using Hooks (<https://reactjs.org/docs/hooks-rules.html>):
```
joebingham-wk: #supertinytyponit I think there's a missing `there` in the comment, or something similar to that
```suggestion
/// Note there are two rules for using Hooks (<https://reactjs.org/docs/hooks-rules.html>):
```

sydneyjodon-wk: Good catch! Thanks!

lib/hooks.dart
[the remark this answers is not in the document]
kealjones-wk: 3fzoer
https://github.com/dart-lang/sdk/issues/4938
https://github.com/dart-lang/language/issues/83

lib/hooks.dart
sydneyjodon-wk: Okay, I updated the example here

example/test/function_component_test.dart
```diff
@@ -1,31 +1,50 @@
...
+var useStateTestFunctionComponent = react.registerFunctionComponent(UseStateTestComponent, displayName: 'useStateTest');
+
+UseStateTestComponent(Map props) {
+  final count = useState(0);
+
+  return react.div({}, [
+    count.value,
+    react.button({'onClick': (_) => count.set(0)}, ['Reset']),
+    react.button({
+      'onClick': (_) => count.setWithUpdater((prev) {
```
greglittlefield-wf: #nit it's better to avoid the state change altogether when disabled:
```dart
      'onClick': (_) {
        if (props['enabled']) {
          count.setWithUpdater((prev) => prev + 1);
        }
      },
```

lib/hooks.dart
```diff
@@ -0,0 +1,109 @@
...
+///
+/// ```
+/// UseStateTestComponent(Map props) {
+///   final count = useState(0);
+///
+///   return react.div({}, [
+///     count.value,
+///     react.button({'onClick': (_) => count.set(0)}, ['Reset']),
+///     react.button({
+///       'onClick': (_) => count.setWithUpdater((prev) {
```
greglittlefield-wf: #nit We should probably show avoiding the state change altogether when disabled:
```dart
      'onClick': (_) {
        if (props['enabled']) {
          count.setWithUpdater((prev) => prev + 1);
        }
      },
```

Or for this example, just do:
```dart
'onClick': (_) => count.setWithUpdater((prev) => prev + 1)
```

lib/hooks.dart
greglittlefield-wf: This is actually a case where you would need to use `setTx`.

Let's say that two setState calls were batched together:
```dart
react.button({'onClick': (_) {
  count.setTx((prev) => prev + 1);
  count.setTx((prev) => prev + 1);
}, ['+2']),
```

Clicking on the button would result in the value being incremented by two.

However, without setTx:
```dart
react.button({'onClick': (_) {
  count.set(count.value + 1);
  count.set(count.value + 1);
}, ['+2']),
```
Clicking on the button would result in the value being incremented by one, which is not expected.

Here, it would be best practice to use transactional setState calls.

React [recommends using transactional state updaters where possible](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous), and it also uses them in [this corresponding example](https://reactjs.org/docs/hooks-reference.html#functional-updates), so I think we should keep `setTx` here

kealjones-wk:
+10

greglittlefield-wf:
+1

PR #227, opened 2019-11-06
CPLAT-8035 Implement/Expose useEffect Hook

sydneyjodon-wk:
## Motivation
Support useEffect hook in react-dart

## Changes
- Added `useEffect` function
- Wrote tests.


lib/hooks.dart
```diff
@@ -99,3 +99,13 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#lazy-initial-state>.
 StateHook<T> useStateLazy<T>(T init()) => StateHook.lazy(init);
+
+///
+void useEffect(void Function() sideEffect) {
+  return React.useEffect(allowInterop(sideEffect));
+}
```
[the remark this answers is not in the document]
sydneyjodon-wk: That's much nicer! Thank you!

example/test/function_component_test.dart
```diff
@@ -28,9 +43,9 @@ void main() {
     react_dom.render(
         react.Fragment({}, [
           react.h1({'key': 'functionComponentTestLabel'}, ['Function Component Tests']),
-          react.h2({'key': 'useStateTestLabel'}, ['useState Hook Test']),
-          useStateTestFunctionComponent({
+          hookTestFunctionComponent({
             'key': 'useStateTest',
+            'enabled': true,
```
kealjones-wk: This isn't needed?

lib/hooks.dart
```diff
@@ -101,3 +101,55 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
...
+///   useEffect(() {
+///     if (count.value % 2 == 0) {
+///       evenOdd.set('even');
+///     } else {
+///       evenOdd.set('odd');
+///     }
+///     return () {
+///       print('count is changing...');
+///     };
+///   }, [count.value]);
```
kealjones-wk: this isn't really required.
```suggestion
///   });
```

example/test/function_component_test.dart
```diff
@@ -5,19 +5,34 @@ import 'package:react/react.dart' as react;
...
+      print('count changed to ' + count.value.toString());
+      evenOdd.set('even');
+    } else {
+      print('count changed to ' + count.value.toString());
+      evenOdd.set('odd');
+    }
+    return () {
+      print('count is changing...');
+    };
+  }, [count.value]);
```
kealjones-wk:
```suggestion
  });
```
i understand why this was added, but it is not really necessary.

lib/hooks.dart
sydneyjodon-wk: I added that because the side effect is updating `evenOdd.value`. Without the dependency, every time `count.value` changes, the side effect will run twice (once because `count.value` changed and again because the side effect changed `evenOdd.value`). If the side effect did something like increment `evenOdd.value`, it would run in an infinite loop unless it had this dependency stopping it from running when `evenOdd.value` changes.

greglittlefield-wf: Ooh could we add this as a comment in the example?

lib/hooks.dart
```diff
@@ -101,3 +101,55 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
...
+///   return react.div({}, [
+///     react.p({}, [count.value.toString() + ' is ' + evenOdd.value.toString()]),
+///     react.button({'onClick': (_) => count.set(count.value + 1)}, ['+']),
+///   ]);
+/// }
+/// ```
+///
+/// See: <https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects>.
+void useEffect(dynamic Function() sideEffect, [List<Object> dependencies]) {
+  var finalSideEffect = allowInterop(() {
```
greglittlefield-wf: #nit `final` isn't a very descriptive name. What about `wrappedSideEffect`?

test/hooks_test.dart
```diff
@@ -92,5 +92,171 @@ main() {
...
+              },
+            }, [
+              '+'
+            ]),
+          ]);
+        });
+
+        react_dom.render(UseEffectTest({}), mountNode);
+
+        await pumpEventQueue();
```
greglittlefield-wf: This should have a comment saying why it's necessary

test/hooks_test.dart
```diff
@@ -92,5 +92,171 @@ main() {
...
+      });
+
+      tearDownAll(() {
+        UseEffectTest = null;
```
greglittlefield-wf: #nit no need to null this out

test/hooks_test.dart
sydneyjodon-wk: Apparently I didn't need them anymore haha

lib/hooks.dart
```diff
@@ -144,7 +144,7 @@ void useEffect(dynamic Function() sideEffect, [List<Object> dependencies]) {
     if (result is Function) {
       return allowInterop(result);
     }
-    return null;
+    return jsUndefined;
```
greglittlefield-wf: Could we add a comment saying why undefined is needed?

lib/hooks.dart
```diff
@@ -102,6 +102,60 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
...
+///   final evenOdd = useState('even');
+///
+///   useEffect(() {
+///     if (count.value % 2 == 0) {
+///       evenOdd.set('even');
+///     } else {
+///       evenOdd.set('odd');
+///     }
+///     return () {
+///       print('count is changing...');
```
aaronlademann-wf: Is this the "cleanup" function? Might be good to notate that either with a comment, or update the string here to say something like: 

```suggestion
///       print('count is changing... do some cleanup if you need to');
```

test/hooks_test.dart
```diff
@@ -93,6 +89,162 @@ main() {
...
+
+      test('side effect (with empty dependency list) is called after the first render', () {
+        expect(useEffectWithEmptyDepsCallCount, 1);
+        expect(useEffectCleanupWithEmptyDepsCallCount, 0, reason: 'component has not been unmounted or re-rendered');
+      });
+
+      group('after state change,', () {
+        setUpAll(() async {
+          react_test_utils.Simulate.click(countButtonRef);
+        });
```
aaronlademann-wf: Why is this function body asynchronous?

test/hooks_test.dart
```diff
@@ -93,6 +89,162 @@ main() {
...
+      int useEffectCallCount;
+      int useEffectCleanupCallCount;
+      int useEffectWithDepsCallCount;
+      int useEffectCleanupWithDepsCallCount;
+      int useEffectWithDepsCallCount2;
+      int useEffectCleanupWithDepsCallCount2;
+      int useEffectWithEmptyDepsCallCount;
+      int useEffectCleanupWithEmptyDepsCallCount;
+
+      setUpAll(() async {
```
aaronlademann-wf: Why is this function body asynchronous?

test/hooks_test.dart
```diff
@@ -93,6 +89,162 @@ main() {
...
+
+        test('side effect (with empty dependency list) is not called again', () {
+          expect(useEffectWithEmptyDepsCallCount, 1,
+              reason: 'side effect is only called once for empty dependency list');
+          expect(useEffectCleanupWithEmptyDepsCallCount, 0, reason: 'component has not been unmounted');
+        });
+      });
+
+      group('after component is unmounted,', () {
+        setUpAll(() async {
```
aaronlademann-wf: Why is this function body asynchronous?

lib/hooks.dart
```diff
@@ -102,6 +102,60 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
...
+  var wrappedSideEffect = allowInterop(() {
+    var result = sideEffect();
+    if (result is Function) {
+      return allowInterop(result);
+    }
+    return jsUndefined;
+  });
+
+  if (dependencies != null) {
+    return React.useEffect(wrappedSideEffect, dependencies);
```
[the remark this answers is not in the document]
sydneyjodon-wk: Yes, passing an empty list is used if you just want to run an effect and clean it up only once (on mount and unmount)

PR #229, opened 2019-11-08
Release react 5.1.0

aaronlademann-wf:
This __Dart 2 only__, stable release of react includes:

### Full React JS 16.x Component Lifecycle Support

- The new `Component2` class replaces the now deprecated `Component` class.
  - Faster
    - Initial renders ~10% faster
    - Re-renders ~7 - 30% faster
  - Improved developer experience
     - Props and state values now show up in the React DevTools just as they would if you were using React JS, and primitive values (strings, numbers, booleans) can be edited live, just like in React JS!
    
        React DevTools with Component2

  - Easier to maintain
  - Easier integration with JS libs
    - `ReactJsComponentFactoryProxy` makes it easy to use JS components with Dart!
      - [Check out this example of MaterialUI components!](https://github.com/cleandart/react-dart/blob/5.1.0-wip/example/js_components/js_components.dart#L115-L145)
  - Supports new lifecycle methods, allowing us to use Concurrent Mode in the future
    - <s>`componentWillMount`</s> => `componentDidMount`
    - <s>`componentWillReceiveProps`</s> => `getDerivedStateFromProps` _(new)_
    - <s>`componentWillUpdate`</s> => `getSnapshotBeforeUpdate` _(new)_
    - `componentDidCatch` / `getDerivedStateFromError` _(new)_
      - Adds support for [error boundaries](https://reactjs.org/docs/error-boundaries.html).
      
### [Portals](https://reactjs.org/docs/portals.html)

- _"Portals provide a first-class way to render children into a DOM node that exists outside the DOM hierarchy of the parent component."_

### Improved, stable [Context](https://reactjs.org/docs/context.html) API

- _"Context provides a way to pass data through the component tree without having to pass props down manually at every level. … Context is primarily used when some data needs to be accessible by many components at different nesting levels. Apply it sparingly because it makes component reuse more difficult."_

### [Fragments](https://reactjs.org/docs/fragments.html)

- _"A common pattern in React is for a component to return multiple elements. Fragments let you group a list of children without adding extra nodes to the DOM."_
- Component.render can now return Fragments (multiple children) or other values like strings and lists instead of just a single ReactElement

### No more Dart 1 SDK support

> [Full list of 5.1.0 Changes](https://github.com/cleandart/react-dart/milestone/1?closed=1)


lib/react_client.dart
```diff
@@ -348,11 +471,155 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
       handleRender: allowInterop(handleRender));
 })();
 
+abstract class _ReactDartInteropStatics2 {
+  // TODO 3.1.0-wip expose for testing?
```
greglittlefield-wf: #followup need to follow up on this and expose it

lib/react_client.dart
```diff
@@ -114,16 +120,10 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
 
       // If the ref is a callback, pass ReactJS a function that will call it
       // with the Dart Component instance, not the ReactComponent instance.
-      //
-      // Use _CallbackRef<Null> to check arity, since parameters could be non-dynamic, and thus
-      // would fail the `is _CallbackRef<dynamic>` check.
-      // See https://github.com/dart-lang/sdk/issues/34593 for more information on arity checks.
-      if (ref is _CallbackRef<Null>) {
```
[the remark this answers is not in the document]
greglittlefield-wf: Resolved in https://github.com/cleandart/react-dart/pull/229/commits/76e598d0cd48145856942261f08529c46174ca4f

CHANGELOG.md
```diff
@@ -1,3 +1,38 @@
...
+__Improved, stable [Context](https://reactjs.org/docs/context.html) API__
+
+- _"Context provides a way to pass data through the component tree without having to pass props down manually at every level. … Context is primarily used when some data needs to be accessible by many components at different nesting levels. Apply it sparingly because it makes component reuse more difficult."_
+
+__[Fragments](https://reactjs.org/docs/fragments.html)__
+
+- _"A common pattern in React is for a component to return multiple elements. Fragments let you group a list of children without adding extra nodes to the DOM."_
+- Component.render can now return Fragments (multiple children) or other values like strings and lists instead of just a single ReactElement
+
+__New and improved ref API: [React.createRef](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs)__
```
greglittlefield-wf: React Redux included in the JS bundle via ReactRedux
```suggestion
__New and improved ref API: [React.createRef](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs)__

__[React Redux](https://react-redux.js.org/) is now included in the JS bundles and exposed via `window.ReactRedux`.__

File: lib/react_client/react_interop.dart [hunk 4]
```diff
@@ -194,8 +344,13 @@ class ReactPortal {
 @JS()
 @anonymous
 class ReactComponent {
+  // TODO: Cast as Component2 in 6.0.0
   external Component get dartComponent;
+  // TODO how to make this JsMap without breaking stuff?
```
aaronlademann-wf: Has this ship sailed?

aaronlademann-wf: Offline: we have worked around this by making `InteropProps` implement `JsMap` for now.  Might be worth re-visiting in 6.0.0 though.

joebingham-wk:
+1

PR #232, opened 2019-11-18
CPLAT-8037 Implement/Expose useReducer Hook

sydneyjodon-wk:
## Motivation
Support useReducer hook in `react-dart`.

## Changes
- Added `ReducerHook` class.
- Added `useReducer` and `useReducerLazy` functions.
- Wrote tests.


lib/hooks.dart
```diff
@@ -101,3 +101,141 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
...
+///
+/// > __Note:__ there are two [rules for using Hooks](https://reactjs.org/docs/hooks-rules.html):
+/// >
+/// > * Only call Hooks at the top level.
+/// > * Only call Hooks from inside a [DartFunctionComponent].
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#usereducer>.
+class ReducerHook {
+  /// The first item of the pair returned by [React.userReducer].
+  Map _state;
```
greglittlefield-wf: State/actions don't have to be maps, do they? If not, this class should use generics:
```dart
class ReducerHook<TState, TAActions> {
  TState _state;
  void Function(TActions ) _dispatch;
}
```
as should the useReucer* functions

lib/hooks.dart
```diff
@@ -101,3 +101,141 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
...
+///       })
+///     }, [
+///       'reset'
+///     ]),
+///   ]);
+/// }
+/// ```
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#lazy-initialization>.
+ReducerHook useReducerLazy<T>(Map Function(Map state, Map action) reducer, dynamic initialArg, Function init) =>
```
greglittlefield-wf: Init could be typed more specifically than just `Function`

lib/hooks.dart
```diff
@@ -102,6 +102,147 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
...
+///     default:
+///       return state;
+///   }
+/// }
+///
+/// UseReducerTestComponent(Map props) {
+///   final state = useReducer(reducer, {'count': 0});
+///
+///   return react.Fragment({}, [
+///     state.state['count'],
```
greglittlefield-wf: #nit having `state.state` is a little confusing; I wonder if we could come up with a different name for the hook return value...

`store`? 😄 🤷‍♂ Nah, that's probably not better...

lib/hooks.dart
```diff
@@ -158,6 +158,147 @@ void useEffect(dynamic Function() sideEffect, [List<Object> dependencies]) {
...
+///       'onClick': (_) => state.dispatch({'type': 'decrement'})
+///     }, [
+///       '-'
+///     ]),
+///   ]);
+/// }
+/// ```
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#usereducer>.
+ReducerHook<TState, TActions, TInit> useReducer<TState, TActions, TInit>(
```
greglittlefield-wf: #nit "Action" should be singular here, so as not to confuse it with flux
```suggestion
ReducerHook<TState, TAction, TInit> useReducer<TState, TAction, TInit>(
```

Same on the next line and in the below function

greglittlefield-wf:
+10

PR #233, opened 2019-11-18
CPLAT-8038 Implement/Expose useCallback Hook

sydneyjodon-wk:
## Motivation
Support useCallback hook in react-dart

## Changes
- Added `useCallback` function
- Wrote tests.

### DIFF
--- a/lib/hooks.dart
+++ b/lib/hooks.dart
@@ -102,9 +102,6 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#lazy-initial-state>.
 StateHook<T> useStateLazy<T>(T init()) => StateHook.lazy(init);
 
-
-
 Function useCallback(Function callback, List dependencies) {
   return React.useCallback(allowInterop(callback), dependencies);
 }
-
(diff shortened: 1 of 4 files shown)


lib/hooks.dart
```diff
@@ -101,3 +101,37 @@ StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#lazy-initial-state>.
 StateHook<T> useStateLazy<T>(T init()) => StateHook.lazy(init);
+
+/// Returns a memoized version of [callback] that only changes if one of the [dependencies] has changed.
+///
+/// Note there are two rules for using Hooks (<https://reactjs.org/docs/hooks-rules.html>):
+///
+/// * Only call Hooks at the top level.
+/// * Only call Hooks from inside a [DartFunctionComponent].
```
kealjones-wk: not a big deal, and i only just realized this after reviewing all these, that maybe this should be formatted like this: 

```
/// > __Note:__ there are two [rules for using Hooks](https://reactjs.org/docs/hooks-rules.html):
/// >
/// > * Only call Hooks at the top level.
/// > * Only call Hooks from inside a [DartFunctionComponent].
```

sydneyjodon-wk: I updated this in a couple places and I'll make sure the rest are updated in the overall hooks branch

test/hooks_test.dart
```diff
@@ -92,5 +92,109 @@ main() {
...
+        react_test_utils.Simulate.click(incrementWithDepButtonRef);
+        expect(countRef.text, '2');
+      });
+
+      group('after depending state changes,', () {
+        setUpAll(() {
+          react_test_utils.Simulate.click(incrementDeltaButtonRef);
+        });
+
+        test('callback stays the same if state not in dependency list', () {
```
greglittlefield-wf: Nice tests here!

PR #236, opened 2019-12-04
CPLAT-8698 Expose the componentZone

joebingham-wk:
### Motivation
To make testing easier, we want to expose the componentZone field to allow the active zone to be configured to match that of a test.

### Changes
- Add `currentComponentZone` and set `_ReactDartInteropStatics2`'s `componentZone` equal to it. 

Please review @greglittlefield-wf @aaronlademann-wf @kealjones-wk @sydneyjodon-wk

### DIFF
--- a/lib/react_client.dart
+++ b/lib/react_client.dart
@@ -62,13 +62,12 @@ dynamic listifyChildren(dynamic children) {
 /// __Example:__
 ///
 ///     test('zone test', () {
-///       currentComponentZone = Zone.current;
+///       componentZone = Zone.current;
 ///
 ///       // ... test your component
 ///     }
-///
 @visibleForTesting
-Zone currentComponentZone = Zone.root;
+Zone componentZone = Zone.root;
 
 /// Use [ReactDartComponentFactoryProxy2] instead.
 ///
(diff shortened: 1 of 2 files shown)


test/lifecycle_test.dart
```diff
@@ -14,7 +14,6 @@ import 'package:react/react.dart' as react;
 import 'package:react/react_client.dart';
 import 'package:react/react_client/js_backed_map.dart';
 import 'package:react/react_client/react_interop.dart' as react_interop;
-import 'package:react/react_client/react_interop.dart';
```
joebingham-wk: Just cleaning up a lil lint

lib/react_client.dart
```diff
@@ -53,6 +54,22 @@ dynamic listifyChildren(dynamic children) {
...
+/// __Example:__
+///
+///     test('zone test', () {
+///       currentComponentZone = Zone.current;
+///
+///       // ... test your component
+///     }
+///
+@visibleForTesting
+Zone currentComponentZone = Zone.root;
```
greglittlefield-wf: Could we just call this `componentZone`? `current` doesn't really add much

lib/react_client.dart
```diff
@@ -473,9 +490,8 @@ final ReactDartInteropStatics _dartInteropStatics = (() {
 })();
 
 abstract class _ReactDartInteropStatics2 {
-  // TODO 3.1.0-wip expose for testing?
   /// The zone in which all component lifecycle methods are run.
-  static final componentZone = Zone.root;
+  static final componentZone = currentComponentZone;
```
greglittlefield-wf: This variable only gets initialized once since it's `static`, so it won't respect updates to `currentComponentZone`. We should support that since
- we're not guaranteed that all tests run in the same zone (they could be descendant zones, etc)
- we may want to set it to two different zones during the execution of the program

PR #237, opened 2019-12-05
CPLAT-8036 Add useContext Hook

joebingham-wk:
### Motivation
Expose the `useContext` hook so that functional components can tap into it.

### Changes
- Add `useContext` function
- Add an example
- Add tests

### Testing
- Verify the example works as expected

Please Review @greglittlefield-wf @kealjones-wk @aaronlademann-wf @sydneyjodon-wk

### DIFF
--- a/example/test/function_component_test.dart
+++ b/example/test/function_component_test.dart
@@ -62,7 +62,7 @@ int calculateChangedBits(currentValue, nextValue) {
 
 var TestNewContext = react.createContext<Map>({'renderCount': 0}, calculateChangedBits);
 
-var newContextProviderComponent = react.registerComponent(() => new _NewContextProviderComponent());
+var newContextProviderComponent = react.registerComponent(() => _NewContextProviderComponent());
 
 class _NewContextProviderComponent extends react.Component2 {
   get initialState => {'renderCount': 0, 'complexMap': false};
--- a/example/test/function_component_test.dart
+++ b/example/test/function_component_test.dart
@@ -12,9 +12,10 @@ UseStateTestComponent(Map props) {
 
   return react.div({}, [
     count.value,
-    react.button({'onClick': (_) => count.set(0)}, ['Reset']),
+    react.button({'onClick': (_) => count.set(0), 'key': 'ust1'}, ['Reset']),
     react.button({
       'onClick': (_) => count.setWithUpdater((prev) => prev + 1),
+      'key': 'ust2'
     }, [
       '+'
     ]),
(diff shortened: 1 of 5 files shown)


lib/hooks.dart
```diff
@@ -135,3 +135,31 @@ StateHook<T> useStateLazy<T>(T init()) => StateHook.lazy(init);
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#usecallback>.
 Function useCallback(Function callback, List dependencies) => React.useCallback(allowInterop(callback), dependencies);
+
+/// Returns the value of the nearest [Context.Provider] for the provided [Context] object every time that context is
```
```suggestion
/// Returns the value of the nearest [Context.Provider] for the provided [context] object every time that context is
```

example/test/function_component_test.dart
```diff
@@ -37,29 +33,96 @@ UseCallbackTestComponent(Map props) {
...
+    react.button({'onClick': increment, 'key': 'ucbt3'}, ['Increment count']),
+    react.button({'onClick': incrementDelta, 'key': 'ucbt4'}, ['Increment delta']),
+  ]);
+}
+
+var useContextTestFunctionComponent =
+    react.registerFunctionComponent(UseContextTestComponent, displayName: 'useContextTest');
+
+UseContextTestComponent(Map props) {
+  final context = useContext(TestNewContext);
```
aaronlademann-wf: It bothers me a bit that `context` will be dynamic unless you provide a type on the left side, and even then - there'd be no way to ensure that the type doesn't mismatch with the underlying type that was returned by `createContext`.  Makes me wonder what the point of the generic parameter is on `createContext`.

By making these changes locally... all the typing seems to line up and makes `context` correctly inferred as a `Map` in this situation as a result of `createContext<Map>` being called on line 63.  

```diff
// context.dart

- class Context {
+ class Context<T> {
// ...
}

- Context createContext<TValue>([
+ Context<TValue> createContext<TValue>([
// ...
```

```diff
// hooks.dart

- dynamic useContext(Context context) => // ...
+ T useContext<T>(Context<T> context) => // ...
```

@greglittlefield-wf @kealjones-wk @joebingham-wk is there some reason that I forgot about that led us to omit generic typing from the `Context` class?

lib/hooks.dart
```diff
@@ -135,3 +135,31 @@ StateHook<T> useStateLazy<T>(T init()) => StateHook.lazy(init);
...
+///   final context = useContext(newContext);
+///
+///   return react.div({}, [
+///     react.div({}, ['The context value is $context']), // initially renders: 'The context value is 0'
+///   ]);
+/// }
+/// ```
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#usecontext>.
+dynamic useContext(Context context) => ContextHelpers.unjsifyNewContext(React.useContext(context.jsThis));
```
aaronlademann-wf: See my other comment about adding generic typing...

```suggestion
T useContext<T>(Context<T> context) => ContextHelpers.unjsifyNewContext(React.useContext(context.jsThis));
```

example/test/function_component_test.dart
kealjones-wk: Honestly I don't recall, I feel like there would have been a reason because the `Context` object in OverReact is typed exactly like that and i don't know why i would have omitted it in React-Dart unless it was just a minor oversight (very likely). @joebingham-wk if you can add this typing id say do it, but test the local context examples in browser and test it with over_react (tests & real examples if already in place) to make sure it doesn't break anything (i don't think it should).

lib/hooks.dart
```diff
@@ -135,3 +135,31 @@ StateHook<T> useStateLazy<T>(T init()) => StateHook.lazy(init);
...
+///
+/// > __Note:__ there are two [rules for using Hooks](https://reactjs.org/docs/hooks-rules.html):
+/// >
+/// > * Only call Hooks at the top level.
+/// > * Only call Hooks from inside a [DartFunctionComponent].
+///
+/// __Example__:
+///
+/// ```
+/// Context newContext = createContext(0, calcuateChangedBits);
```
greglittlefield-wf: `calculateChangedBits` isn't used by most consumers, so we should omit it.

Also, giving a name to the context would help give the example more clarity. For instance:
```dart
/// ```
/// Context countContext = createContext(0);
///
/// UseCallbackTestComponent(Map props) {
///   final count = useContext(countContext);
///
///   return react.div({}, [
///     'The count from context is $count', // initially renders: 'The count from context is 0'
///   ]);
/// }
/// ```
```

test/hooks_test.dart
```diff
@@ -196,5 +197,90 @@ main() {
...
+          expect(currentCount, 3);
+          providerRef.increment();
+          expect(currentCount, 4);
+        });
+      });
+    });
   });
 }
+
+int calculateChangedBits(currentValue, nextValue) {
```
greglittlefield-wf: Is this necessary?

test/hooks_test.dart
```diff
@@ -196,5 +197,90 @@ main() {
         });
       });
     });
+
+    group('useContext -', () {
+      var mountNode = DivElement();
```
greglittlefield-wf: This needs to be unmounted in `tearDown`.

test/hooks_test.dart
joebingham-wk: Nope, good call!

test/hooks_test.dart
```diff
@@ -196,5 +197,85 @@ main() {
...
+                providerRef = ref;
+              }
+            }, [
+              useContextTestFunctionComponent({'key': 't1'}, []),
+            ]),
+            mountNode);
+      });
+
+      tearDown(() {
+        mountNode = null;
```
greglittlefield-wf: This still needs an unmount call
```suggestion
        react_dom.unmountComponentAtNode(mountNode);
        mountNode = null;
```

joebingham-wk: 🤦‍♂ I'm so used to writing tests in over_react with utils that handle that for me. Sorry for missing that!

greglittlefield-wf: Right?? No worries!

PR #239, opened 2019-12-11
Release react 5.2.1 (hotfix)

aaronlademann-wf:
This hotfix temporarily pins the JS react-redux dependency to version `7.1.0` to prevent widespread test failures as a result of [`Provider` being converted into a function component with hooks](https://github.com/reduxjs/react-redux/pull/1377).

---

@kealjones-wk @joebingham-wk @greglittlefield-wf @dustyholmes-wf


kealjones-wk:
+1

PR #244, opened 2020-01-02
Composition events

theisensanders-wf:
## Motivation
Add support for [Composition events](https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent).

## Changes
- `onCompositionStart`, `onCompositionUpdate`, and `onCompositionEnd` event handlers.
- Synthetic event type
- Simulate test utils

Note: w3 spec events are `compositionstart`, `compositionupdate`, and `compositionend` (all lower-case), but react.js uses camel case for event names

Please Review @greglittlefield-wf @kealjones-wk @aaronlademann-wf @sydneyjodon-wk @joebingham-wk


theisensanders-wf:
Turns out this doesn't solve my immediate issue since I'd be adding the listeners on dynamically added elements, not react elements. BUT I do think this is still a good change to have since react.js supports these listeners even though dart2js [does not](https://raw.githubusercontent.com/dart-lang/sdk/master/sdk/lib/html/dart2js/html_dart2js.dart).

lib/react.dart
```diff
@@ -1577,6 +1577,27 @@ class SyntheticKeyboardEvent extends SyntheticEvent {
             isTrusted, nativeEvent, target, timeStamp, type) {}
 }
 
+class SyntheticCompositionEvent extends SyntheticEvent {
+  final data;
```
greglittlefield-wf: Are you okay with making this a string? According to MDN, that's the type: https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent/data
```suggestion
  final String data;
```

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -49,6 +49,12 @@ class SyntheticKeyboardEvent extends SyntheticEvent {
   external num get charCode;
 }
 
+@JS()
+@anonymous
+class SyntheticCompositionEvent extends SyntheticEvent {
+  external get data;
```
greglittlefield-wf: Same as above
```suggestion
  external String get data;
```

lib/react.dart
theisensanders-wf: Yes good call. Looks like react.js also defines it as a string
https://reactjs.org/docs/events.html#composition-events

theisensanders-wf: @greglittlefield-wf fixed

PR #245, opened 2020-01-06
CPLAT-8040 Implement/Expose useRef Hook

sydneyjodon-wk:
## Motivation
Support useRef hook in `react-dart`.

## Changes
- Added `useRef` function.
- Added a `initUseRef` constructor to the `Ref` object.
- Added a setter for the `current` property of `Ref`.
- Wrote tests.


lib/hooks.dart
```diff
@@ -219,3 +219,39 @@ Function useCallback(Function callback, List dependencies) => React.useCallback(
...
+/// > * Only call Hooks from inside a [DartFunctionComponent].
+///
+/// __Example__:
+///
+/// ```
+/// UseRefTestComponent(Map props) {
+///   final input = useState('');
+///   final inputRef = useRef();
+///   final prevInputRef = useRef();
+///   final prevInput = prevInputRef.current;
```
greglittlefield-wf: These variables are named pretty similarly, which makes it a little hard to tell what's going on. I think further distinguishing them and also adding some typing to the refs would help:

```dart
final inputValue = useState('');

final inputRef = useRef<InputElement>();
final prevInputValueRef = useRef<String>();

useEffect(() {
  prevInputValueRef.current = inputValue.value;
});

return react.Fragment({}, [
  react.p({}, ['Current Input: ${inputValue.value}, Previous Input: ${prevInputValueRef.current}']),
  react.input({'ref': inputRef}),
  react.button({'onClick': (_) => inputValue.set(inputRef.current.value)}, ['Update']),
]);
```

lib/hooks.dart
```diff
@@ -219,3 +219,39 @@ Function useCallback(Function callback, List dependencies) => React.useCallback(
...
+///   return react.Fragment({}, [
+///     react.p({}, ['Current Input: ${input.value}, Previous Input: ${prevInput}']),
+///     react.input({'ref': inputRef}),
+///     react.button({'onClick': (_) => input.set(inputRef.current.value)}, ['Update']),
+///   ]);
+/// }
+/// ```
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#useref>.
+Ref useRef([dynamic initialValue]) => new Ref.useRefInit(initialValue);
```
greglittlefield-wf: Ah, I didn't realize this wasn't generic above. I think we could make this generic:
```suggestion
Ref<T> useRef<T>([T initialValue]) => Ref.useRefInit(initialValue);
```

(`Ref.useRefInit` would also have to be made generic instead of just accepting `dynamic`)

lib/react_client/react_interop.dart
```diff
@@ -81,6 +83,11 @@ class Ref<T> {
 
   Ref() : jsRef = React.createRef();
 
+  /// Constructor for [useRef], calls [React.useRef] to initialize [current] to [initialValue].
+  ///
+  /// See: <https://reactjs.org/docs/hooks-reference.html#useref>.
+  Ref.useRefInit(dynamic initialValue) : jsRef = React.useRef(initialValue);
```
greglittlefield-wf: See above
```suggestion
  Ref.useRefInit(T initialValue) : jsRef = React.useRef(initialValue);
```

lib/react_client/react_interop.dart
```diff
@@ -89,15 +96,20 @@ class Ref<T> {
...
         return dartCurrent as T;
       }
     }
     return jsCurrent;
   }
+
+  /// Sets the value of [current].
+  ///
+  /// See: <https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables>.
+  set current(dynamic value) => jsRef.current = value;
```
greglittlefield-wf: This should be `T`, so they can't statically set an incompatible value
```suggestion
  set current(T value) => jsRef.current = value;
```

test/hooks_test.dart
```diff
@@ -407,6 +407,85 @@ main() {
...
+            ]),
+          ]);
+        });
+
+        react_dom.render(UseRefTest({}), mountNode);
+      });
+
+      group('correctly initializes a Ref object', () {
+        test('with current property set to null if no initial value given', () {
+          expect(noInitRef is Ref, isTrue);
```
greglittlefield-wf: There's a matcher for this that will give you better error messages when this fails, instead of just saying `Expected: true. Actual: false`: 
```suggestion
          expect(noInitRef, isA<Ref>());
```

test/hooks_test.dart
```diff
@@ -407,6 +407,85 @@ main() {
...
+          expect(noInitRef.current, null);
+        });
+
+        test('with current property set to the initial value given', () {
+          expect(initRef is Ref, isTrue);
```
greglittlefield-wf: Same as above
```suggestion
          expect(initRef, isA<Ref>());
```

test/hooks_test.dart
```diff
@@ -407,6 +407,85 @@ main() {
...
+          expect(initRef.current, mountNode);
+        });
+      });
+
+      group('the returned Ref', () {
+        test('can be attached to elements via the ref attribute', () {
+          expect(domElementRef.current is InputElement, isTrue);
```
greglittlefield-wf: Same as above
```suggestion
          expect(domElementRef.current, isA<InputElement>());
```

greglittlefield-wf:
Looks like there are merge conflicts on this one now too after merging another PR 😅

test/hooks_test.dart
sydneyjodon-wk: Nice! Thanks!

PR #246, opened 2020-01-07
CPLAT-8044 Implement/Expose useDebugValue Hook

sydneyjodon-wk:
## Motivation
Support useDebugValue hook in `react-dart`.

## Changes
- Added `useDebugValue` function.
- Wrote tests.

### QA Checklist
- [ ] Steps from PR author: 
    - [ ] Run `pbr serve example` and open http://localhost:8080/test/function_component_test.html
    - [ ] Verify that debug value shows up under Hooks in React Dev Tools
image

#### Note
The debug value does not show up in dart2js because the custom hook `useFriendStatus` is not showing. 
image

Please review: @greglittlefield-wf @aaronlademann-wf @joebingham-wk


lib/hooks.dart
```diff
@@ -472,3 +473,77 @@ void useLayoutEffect(dynamic Function() sideEffect, [List<Object> dependencies])
...
+///           }, []),
+///         ]),
+///     displayName: 'useDebugValueTest');
+/// ```
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#usedebugvalue>.
+dynamic useDebugValue<T>(T value, [dynamic Function(T) format]) {
+  if (format == null) {
+    return React.useDebugValue(value);
+  }
```
joebingham-wk: #nit this could be a one liner, but I believe that's just preference :) 
```
if (format == null) return React.useDebugValue(value);
```

lib/react_client/react_interop.dart
```diff
@@ -55,6 +55,7 @@ abstract class React {
   external static JsRef useRef([dynamic initialValue]);
   external static dynamic useMemo(dynamic Function() createFunction, [List<dynamic> dependencies]);
   external static void useLayoutEffect(dynamic Function() sideEffect, [List<Object> dependencies]);
+  external static dynamic useDebugValue(dynamic value, [Function format]);
```
joebingham-wk: Definitely check with @aaronlademann-wf or @greglittlefield-wf on this one, but I think we could make this typing a little more specific. I believe we could use generics to specify that the expected function should take in the `T` generic. Something like:
```
external static dynamic useDebugValue<T>(T value, [dynamic Function(T) format]);
```

test/hooks_test.dart
```diff
@@ -642,6 +642,76 @@ main() {
...
       testEffectHook(useLayoutEffect);
     });
+
+    group('useDebugValue -', () {
+      ReactDartFunctionComponentFactoryProxy UseDebugValueTest1;
+      ReactDartFunctionComponentFactoryProxy UseDebugValueTest2;
+      StateHook<bool> isOnline1;
+      StateHook<bool> isOnline2;
+
+      StateHook useFriendStatus(friendID) {
```
joebingham-wk: Unused parameter - was there a reason to keep it around?

example/test/function_component_test.dart
```diff
@@ -272,6 +272,63 @@ RandomUseEffectTestComponent(Map props) {
   ]);
 }
 
+final ChatAPI = {
```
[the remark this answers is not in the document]
sydneyjodon-wk: That's a good point haha. I just did it this way because that's how the example in React was done, but a class is much better.

test/hooks_test.dart
sydneyjodon-wk: nope lol

lib/react_client/react_interop.dart
sydneyjodon-wk: When I wrapped the hook I used generics like this to specify the typing for the dart version of `useDebugValue`. I'm not sure if it's necessary to also use generics here when this function won't be exposed and is just being wrapped by the dart version. It seems like the other hooks that used generics to tighten the typing in the dart function didn't also use generics here. Is there a reason these React hooks should be generically typed as well?

joebingham-wk: It also crossed my mind if this was necessary, so if you feel strongly I definitely don't 😄 

I threw it out there because I believe that it _is_ exposed, being outside of the `src` directory. No one should ever use it directly, but, that doesn't always stop people ;) So if it doesn't get it in the way and we can enforce a usage pattern, it made sense to me to just throw it in there but it's also probably not worth thinking too much about haha

sydneyjodon-wk: Interesting. So should I add generics to all the hooks in the React class that need them?

joebingham-wk: I'd probably check with Greg on his thoughts there. I'd lean towards it being nice to have and a small change if we're going to add it to this one, but I still think you could be right that it's not necessary and Greg will agree there - so if that's the case I don't want ya to spend that time unnecessarily.

This one in particular struck me as a special case because we'd be trying to enforce that `value` is both its own parameter as well as the only parameter for the `format` callback. I don't think any other hook quite has that situation.. so that also could be a case for not worrying about the others

lib/react_client/react_interop.dart
```diff
@@ -55,6 +55,7 @@ abstract class React {
   external static JsRef useRef([dynamic initialValue]);
   external static dynamic useMemo(dynamic Function() createFunction, [List<dynamic> dependencies]);
   external static void useLayoutEffect(dynamic Function() sideEffect, [List<Object> dependencies]);
+  external static dynamic useDebugValue<T>(T value, [dynamic Function(T) format]);
```
joebingham-wk:
```suggestion
  // NOTE: The use of generics on the `useDebugValue` interop will break the hook.
  external static dynamic useDebugValue(dynamic value, [Function format]);
```

PR #247, opened 2020-01-14
CPLAT-8041 Implement/Expose useImperativeHandle Hook

sydneyjodon-wk:
## Motivation
Support useImperativeHandle hook in `react-dart`.

## Changes
- Added `useImperativeHandle` function.
- Wrote tests.

Please review: @greglittlefield-wf @aaronlademann-wf @joebingham-wk


example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
   ]);
 }
 
+var FancyInput = react.forwardRef((props, ref) {
+  var inputRef = useRef();
```
joebingham-wk: #nit this could be `final`
```suggestion
  final inputRef = useRef();
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
...
+    'placeholder': props['placeholder'],
+    'style': {'borderColor': props['hasError'] ? 'crimson' : '#999'},
+  });
+});
+
+var useImperativeHandleTestFunctionComponent =
+    react.registerFunctionComponent(UseImperativeHandleTestComponent, displayName: 'useImperativeHandleTest');
+
+UseImperativeHandleTestComponent(Map props) {
+  var city = useState('');
```
joebingham-wk: #nit this could be final
```suggestion
  final city = useState('');
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
...
+  var state = useState('');
```
joebingham-wk: #nit this could be final
```suggestion
  final state = useState('');
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
...
+  var error = useState('');
```
joebingham-wk: #nit this could be final
```suggestion
  final error = useState('');
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
...
+  var message = useState('');
```
joebingham-wk: #nit this could be final
```suggestion
  final message = useState('');
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
...
+  return react.input({
+    'ref': inputRef,
+    'value': props['value'],
+    'onChange': (e) => props['update'](e.target.value),
+    'placeholder': props['placeholder'],
+    'style': {'borderColor': props['hasError'] ? 'crimson' : '#999'},
+  });
+});
+
+var useImperativeHandleTestFunctionComponent =
```
joebingham-wk: #nit this could be final
```suggestion
final useImperativeHandleTestFunctionComponent =
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
   ]);
 }
 
+var FancyInput = react.forwardRef((props, ref) {
+  var inputRef = useRef();
+
+  useImperativeHandle(
+    ref,
+    () => ({
+      'focus': () {
```
joebingham-wk: #nit this could be a one liner, but is already good as is :) 

```
'focus': () => inputRef.current.focus(),
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,80 @@ UseRefTestComponent(Map props) {
   ]);
 }
 
+var FancyInput = react.forwardRef((props, ref) {
```
joebingham-wk: #nit this could be final
```suggestion
final FancyInput = react.forwardRef((props, ref) {
```

lib/hooks.dart
```diff
@@ -396,3 +396,54 @@ T useContext<T>(Context<T> context) => ContextHelpers.unjsifyNewContext(React.us
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#useref>.
 Ref<T> useRef<T>([T initialValue]) => Ref.useRefInit(initialValue);
+
+/// Customizes the [ref] value that is exposed to parent components when using [forwardRef] by setting [ref.current]
+/// to the return value of [createHandle].
+///
+/// In most cases, imperative code using refs should be avoided.
```
joebingham-wk: React's hook docs don't do this, but how would you feel about linking to the ref documentation here to describe when they _should_ use refs?
```suggestion
/// In most cases, imperative code using refs should be avoided.
/// > For more information, see <https://reactjs.org/docs/refs-and-the-dom.html>.
```

lib/hooks.dart
```diff
@@ -396,3 +396,54 @@ T useContext<T>(Context<T> context) => ContextHelpers.unjsifyNewContext(React.us
...
+///       'update': inputValue.set,
+///       'ref': fancyInputRef,
+///     }, []),
+///     react.button({'onClick': (_) => fancyInputRef.current['focus']()}, ['Focus Input']),
+///   ]);
+/// }
+/// ```
+///
+/// Learn more: <https://reactjs.org/docs/hooks-reference.html#useimperativehandle>.
+void useImperativeHandle(Ref ref, Function() createHandle, [List dependencies]) =>
```
joebingham-wk: I'm not the typing master like @greglittlefield-wf is, but I feel like the `createHandle` typing could use a return type and the `dependencies` could explicitly type its `List`
```suggestion
void useImperativeHandle(Ref ref, void Function() createHandle, [List<dynamic> dependencies]) =>
```

lib/react_client/react_interop.dart
```diff
@@ -53,6 +53,7 @@ abstract class React {
   external static Function useCallback(Function callback, List dependencies);
   external static ReactContext useContext(ReactContext context);
   external static JsRef useRef([dynamic initialValue]);
+  external static void useImperativeHandle(JsRef ref, Function() createHandle, [List dependencies]);
```
joebingham-wk: Based on whether or not my typing information of `hooks.useImperativeHandle` is correct, this would be updated to match:
```suggestion
  external static void useImperativeHandle(JsRef ref, void Function() createHandle, [List<dynamic> dependencies]);
```

test/hooks_test.dart
```diff
@@ -666,6 +666,111 @@ main() {
...
+            }, [
+              '+'
+            ]),
+            EmptyDepsComponent({'ref': emptyDepsRef}),
+            DepsComponent({'ref': depsRef}),
+          ]);
+        });
+
+        react_dom.render(UseImperativeHandleTest({}), mountNode);
+      });
```
joebingham-wk: #nit I'd add a line here, personally 
```suggestion
      });

```

lib/hooks.dart
sydneyjodon-wk: Good idea!

example/test/function_component_test.dart
```diff
@@ -183,6 +183,134 @@ UseRefTestComponent(Map props) {
...
 }
 
+final FancyInput = react.forwardRef((props, ref) {
+  final inputRef = useRef();
+
+  useImperativeHandle(
+    ref,
+    () {
+      print('FancyInput: useImperativeHandle re-assigns ref.current');
+      return {'focus': () => inputRef.current.focus()};
```
greglittlefield-wf: This matches the JS implementation by using a Map, but I think it might be helpful to also show a Dart object being used (perhaps in another example), to illustrate that you can use typed objects and not just Maps.

```dart
class FancyInputApi {
  final void Function() focus;
  FancyInputApi._(this.focus);
}

final FancyInput = react.forwardRef((props, ref) {
  final inputRef = useRef();

  useImperativeHandle(
    ref,
    () {
      print('FancyInput: useImperativeHandle re-assigns ref.current');
      return FancyInputApi(() => inputRef.current.focus());
    },

    /// Because the return value of createHandle never changes, it is not necessary for ref.current
    /// to be re-set on each render so this dependency list is empty.
    [],
  );
});
```

It is admittedly a bit clunkier than using a Map, but it does provide static typing.

Thoughts?

example/test/function_component_test.dart
```diff
@@ -183,6 +183,134 @@ UseRefTestComponent(Map props) {
...
+final useImperativeHandleTestFunctionComponent =
+    react.registerFunctionComponent(UseImperativeHandleTestComponent, displayName: 'useImperativeHandleTest');
+
+UseImperativeHandleTestComponent(Map props) {
+  final city = useState('');
+  final state = useState('');
+  final error = useState('');
+  final message = useState('');
+
+  Ref cityEl = useRef();
```
greglittlefield-wf: I think the `-El` suffix misleading here, since it's not actually an element, but rather a Map. Can we add typing to these as well to further show that? 

Maybe something like:
```dart
final cityRef = useRef<Map>();
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,134 @@ UseRefTestComponent(Map props) {
...
+  return react.div({}, count.value);
+});
+
+final useImperativeHandleTestFunctionComponent2 =
+    react.registerFunctionComponent(UseImperativeHandleTestComponent2, displayName: 'useImperativeHandleTest2');
+
+UseImperativeHandleTestComponent2(Map props) {
+  final diff = useState(1);
+
+  Ref fancyCounterRef = useRef();
```
greglittlefield-wf: Same as above, this could be typed:
```suggestion
  final fancyCounterRef = useRef<Map>();
```

example/test/function_component_test.dart
```diff
@@ -183,6 +183,134 @@ UseRefTestComponent(Map props) {
...
+}
+
+final FancyCounter = react.forwardRef((props, ref) {
+  final count = useState(0);
+
+  useImperativeHandle(
+    ref,
+    () {
+      print('FancyCounter: useImperativeHandle re-assigns ref.current');
+      return ({
```
greglittlefield-wf: #nit unnecessary parentheses

lib/hooks.dart
```diff
@@ -396,3 +396,59 @@ T useContext<T>(Context<T> context) => ContextHelpers.unjsifyNewContext(React.us
...
+///
+/// __Example__:
+///
+/// ```
+/// var FancyInput = forwardRef((props, ref) {
+///   var inputRef = useRef<InputElement>();
+///
+///   useImperativeHandle(
+///     ref,
+///     () => ({
```
greglittlefield-wf: #nit unnecessary parentheses

example/test/function_component_test.dart
sydneyjodon-wk: That makes sense!

lib/hooks.dart
```diff
@@ -472,3 +472,60 @@ void useLayoutEffect(dynamic Function() sideEffect, [List<Object> dependencies])
...
+/// For more information, see <https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs>.
+///
+/// > __Note:__ there are two [rules for using Hooks](https://reactjs.org/docs/hooks-rules.html):
+/// >
+/// > * Only call Hooks at the top level.
+/// > * Only call Hooks from inside a [DartFunctionComponent].
+///
+/// __Example__:
+///
+/// ```
```
```suggestion
/// ```dart
```

PR #249, opened 2020-01-21
Proxy event.persist() when wrapping synthetic JS events

aaronlademann-wf:
While playing around with using `forwardRef` in [the over_react redux todo demo app](https://github.com/Workiva/over_react/pull/439), I discovered that if a JS component makes use of [`event.persist()`](https://reactjs.org/docs/events.html#event-pooling), a runtime exception occurs if that JS component is wrapped with `forwardRef` as a result of the Dart event wrapper not having a method called `persist`.

This PR remedies that issue in a non-breaking way by adding a public - but very ominous field `$$jsPersistDoNotSetThisOrYouWillBeFired` - which gets set to a function by that analogous synthetic event factory in `react_client.dart`. This field will call the JS `event.persist()` function when called on the Dart side.  The use of this field was necessary since the pre-existing pattern was to add have an argument in the constructor that would do the same _(e.g. `SyntheticEvent._preventDefault` gets set via the `_preventDefault` argument in the constructor)_.  Following this pre-existing pattern would require us to make breaking changes to numerous public constructor signatures.

FYA: @sydneyjodon-wk @joebingham-wk @greglittlefield-wf @kealjones-wk

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -1451,6 +1451,29 @@ class SyntheticEvent {
   /// See: <https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation>
   final dynamic stopPropagation;
 
+  /// For use by react-dart internals only.
+  @protected
+  void Function() $$jsPersistDoNotSetThisOrYouWillBeFired;
+  bool _isPersistent = false;
+
+  /// Whether the event instance has been removed from the ReactJS event pool.
+  ///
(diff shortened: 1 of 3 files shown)


greglittlefield-wf:
@aaronlademann-wf Do you have steps to reproduce this? I'm still having trouble seeing how a JS component could call a the Dart synthetic event's method; seems like some other issue may be at play here.

PR #251, opened 2020-01-23
Improve forwardRef documentation, add displayName argument

aaronlademann-wf:
While grokking our `forwardRef` API, I found myself wanting some implementation examples which could not be found in the documentation comment.

I also noticed that consumers had no way to define a `displayName` for the wrapper HOC created by `forwardRef`.

These changes address both issues by adding extensive examples in the documentation comment for `forwardRef` that mirror the analogous [ReactJS forwardRef documentation](https://reactjs.org/docs/forwarding-refs.html), and by adding a `displayName` argument to the function which then gets set on the JS side via a call to `defineProperty`.

---

FYA: @greglittlefield-wf @joebingham-wk @sydneyjodon-wk


lib/react_client/react_interop.dart
```diff
@@ -127,17 +128,156 @@ class JsRef {
...
+///
+/// /// This makes it so that the `ref` set on the Dart `IconButton` component
+/// /// gets passed all the way to the root node rendered by the JS `MaterialUI.IconButton` component.
+/// final IconButton = forwardRefToJsComponent(MaterialUI.IconButton);
+///
+/// void main() {
+///   setClientConfiguration();
+///   final iconButtonRootNodeRef = react.createRef<Element>();
+///
+///   react_dom.render(IconButton({'ref': iconButtonRootNodeRef}, someIcon), querySelector('#idOfSomeNodeInTheDom'));
```
greglittlefield-wf: I don't think this is the best example, because this example would function the same without the `forwardRefToJsComponent` call. Could we show it with the `additionalRefPropKeys` arg?

lib/react_client/react_interop.dart
```diff
@@ -127,17 +128,156 @@ class JsRef {
...
+Map _convertNonDefaultRefPropKeysToJs(Map props, List<String> additionalRefPropKeys) {
+  var propsMapToReturn = props;
+  if (additionalRefPropKeys.isNotEmpty) {
+    var propsWithConvertedRefs = {...props};
+    for (var propKey in additionalRefPropKeys) {
+      var ref = propsWithConvertedRefs[propKey];
+
+      // The default 'ref' prop key value gets converted by default
+      // by ReactJsComponentFactoryProxy, so we don't need to do it here.
+      if (propKey != 'ref' && ref is Ref) {
```
greglittlefield-wf: Should we also have wrapping for callback refs here?

test/factory/js_factory_test.dart
```diff
@@ -2,9 +2,13 @@
 @TestOn('browser')
 library js_factory_test;
 
+import 'dart:html';
+import 'dart:js_util';
```
joebingham-wk: #nit unused import
```suggestion
```

lib/react_client/react_interop.dart
aaronlademann-wf: @greglittlefield-wf is this what you mean?

```dart
if (ref is Function) {
  propsWithConvertedRefs[propKey] = allowInterop((ReactComponent instance) {
    // Call as dynamic to perform dynamic dispatch, since we can't cast to _CallbackRef<dynamic>,
    // and since calling with non-null values will fail at runtime due to the _CallbackRef<Null> typing.
    return ref(instance?.dartComponent);
  });
} else if (propKey != 'ref' && ref is Ref) {
  propsWithConvertedRefs[propKey] = ref.jsRef;
}
```

greglittlefield-wf: Yup!

lib/react_client/react_interop.dart
```diff
@@ -127,15 +130,171 @@ class JsRef {
...
+///   setClientConfiguration();
+///   final textFieldInputNodeRef = react.createRef<TextInputElement>();
+///
+///   react_dom.render(TextInput({'inputRef': textFieldInputNodeRef}), querySelector('#idOfSomeNodeInTheDom'));
+///
+///   // Prints the value of the <input> rendered by the JS `MaterialUI.TextField` component.
+///   print(textFieldInputNodeRef.current.value);
+/// }
+/// ```
+ReactJsComponentFactoryProxy forwardRefToJs(
```
greglittlefield-wf: Taking a fresh look at what this functionality provides, I think this name is misleading, since it doesn't really provide similar functionality to forwardRef.

What about something like `withJsRefsConverted`? Also, `additionalRefPropKeys` should probably be renamed to not have `additional`, since this function doesn't really do anything without them

lib/react_client/react_interop.dart
greglittlefield-wf: Now that I think about it, I almost wonder if we should just make `additionalRefPropKeys` a constructor arg in ReactJsComponentFactoryProxy and have this logic live in `_generateJsProps`...

lib/react_client/react_interop.dart
aaronlademann-wf: I'm confused what you mean by _"doesn't really provide similar functionality to forwardRef"_.

This function returns the return value of a call to `forwardRef`.

lib/react_client/react_interop.dart
aaronlademann-wf: @greglittlefield-wf re: the conversion of ref callbacks... `ReactJsComponentFactoryProxy` calls `_generateJsProps` with `convertCallbackRefValue` set to false.  Should it be true?

lib/react_client/react_interop.dart
greglittlefield-wf: Right, but it doesn't allow you to forward a ref onto a different component/prop. It's really just a HOC that's using `forwardRef`.

For example, the last snippet in this section: https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components
> ```jsx
> function logProps(Component) {
>   class LogProps extends React.Component {
>     componentDidUpdate(prevProps) {
>       console.log('old props:', prevProps);
>       console.log('new props:', this.props);
>     }
> 
>     render() {
>       const {forwardedRef, ...rest} = this.props;
> 
>       // Assign the custom prop "forwardedRef" as a ref
>       return <Component ref={forwardedRef} {...rest} />;
>     }
>   }
> 
>   // Note the second param "ref" provided by React.forwardRef.
>   // We can pass it along to LogProps as a regular prop, e.g. "forwardedRef"
>   // And it can then be attached to the Component.
>   return React.forwardRef((props, ref) => {
>     return <LogProps {...props} forwardedRef={ref} />;
>   });
> }
> ```

`logProps` isn't `logPropsForwardRef`, it just uses it so that the HOC passes the ref through.

lib/react_client/react_interop.dart
greglittlefield-wf: @aaronlademann-wf No, I think the naming on that is just confusing. That flag controls whether callback refs should be wrapped with a version that looks for the Dart component and tries to pass that in instead, and so should only be necessary for Dart component factories

lib/src/react_client/factory_util.dart
```diff
@@ -63,24 +72,32 @@ void convertRefValue(Map args) {
...
-      return (ref as dynamic)(instance);
-    });
+void convertRefValue2(
+  Map args, {
+  bool convertCallbackRefValue = true,
+  List<String> additionalRefPropKeys = const [],
+}) {
+  final refKeys = ['ref', ...additionalRefPropKeys];
+
+  for (var refKey in refKeys) {
```
joebingham-wk: #nit `refKey` could be final
```suggestion
  for (final refKey in refKeys) {
```

lib/src/react_client/factory_util.dart
```diff
@@ -63,24 +72,32 @@ void convertRefValue(Map args) {
...
+    var ref = args[refKey];
```
joebingham-wk: #nit ref could be final
```suggestion
    final ref = args[refKey];
```

lib/src/react_client/factory_util.dart
```diff
@@ -63,24 +72,32 @@ void convertRefValue(Map args) {
...
+      //
+      // Use _CallbackRef<Null> to check arity, since parameters could be non-dynamic, and thus
+      // would fail the `is _CallbackRef<dynamic>` check.
+      // See https://github.com/dart-lang/sdk/issues/34593 for more information on arity checks.
+    } else if (ref is CallbackRef<Null> && convertCallbackRefValue) {
+      args[refKey] = allowInterop((dynamic instance) {
+        // Call as dynamic to perform dynamic dispatch, since we can't cast to _CallbackRef<dynamic>,
+        // and since calling with non-null values will fail at runtime due to the _CallbackRef<Null> typing.
+        if (instance is ReactComponent && instance.dartComponent != null)
+          return (ref as dynamic)(instance.dartComponent);
```
joebingham-wk: #nit [prefer using curly braces for multiline if statement](https://dart.dev/guides/language/effective-dart/style#do-use-curly-braces-for-all-flow-control-structures) (probably just got wrapped during formatting)

lib/src/react_client/factory_util.dart
```diff
@@ -126,11 +143,14 @@ JsMap generateJsProps(Map props,
...
     bool convertCallbackRefValue = true,
+    List<String> additionalRefPropKeys = const [],
     bool wrapWithJsify = true}) {
   final propsForJs = JsBackedMap.from(props);
 
   if (shouldConvertEventHandlers) convertEventHandlers(propsForJs);
-  if (convertRefValue) convertRefValue2(propsForJs, convertCallbackRefValue: convertCallbackRefValue);
+  if (convertRefValue)
+    convertRefValue2(propsForJs,
+        convertCallbackRefValue: convertCallbackRefValue, additionalRefPropKeys: additionalRefPropKeys);
```
joebingham-wk: #nit [prefer using curly braces for multiline if statement](https://dart.dev/guides/language/effective-dart/style#do-use-curly-braces-for-all-flow-control-structures) (probably just got wrapped during formatting)

lib/react_client/react_interop.dart
aaronlademann-wf: @greglittlefield-wf I disagree that from an API clarity standpoint - `withJsRefsConverted` is more clear about the purpose of the function.

The purpose of the function is so that refs get forwarded.  All the HOC wrapping is semantics, IMO. The function does more than just convert refs... it also **forwards** them to the JS component.

greglittlefield-wf: Okay, so taking a step back...

At this point, this function is just a wrapper around `ReactJsComponentFactoryProxy`, and the `forwardRef` call really serves no purpose. 

With the following changes:
```diff
-  return forwardRef((props, ref) {
-    return jsFactoryProxy({
-      ...props,
-      'ref': ref,
-    }, props['children']);
-    // Convert DOM props for the HOC so the consumer API callbacks that involve events receive Dart `SyntheticEvent`s.
-  }, displayName: '${displayName ?? jsClassComponent.displayName}');
+  return jsFactoryProxy;
```
The behavior would be exactly the same.

So, why do we need this function at all, if it's just a wrapper around `ReactJsComponentFactoryProxy`? Consumers should just use that instead.

PR #252, opened 2020-01-24
Implement React.memo

aaronlademann-wf:
> __This is not a clean diff.  It will not be clean until https://github.com/cleandart/react-dart/pull/251 merges__

This PR implements [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) in the Dart API, along with examples and tests.

---

FYA: @greglittlefield-wf @sydneyjodon-wk @joebingham-wk


test/react_memo_test.dart
```diff
@@ -0,0 +1,163 @@
+import 'dart:async';
```
joebingham-wk: #nit unused import
```suggestion
```

test/react_memo_test.dart
```diff
@@ -0,0 +1,163 @@
...
+import 'dart:developer';
```
joebingham-wk: #nit unused import
```suggestion
```

test/react_memo_test.dart
```diff
@@ -0,0 +1,163 @@
...
+@TestOn('browser')
+import 'dart:html';
+import 'dart:js_util';
```
joebingham-wk: #nit unused import
```suggestion
```

PR #258, opened 2020-06-03
CPLAT-11281 Only include propTypes in DDC output

aaronlademann-wf:
> NOTE: We need to coordinate the release of these changes with [updates to the `logsPropError*` matchers in over_react_test](https://github.com/Workiva/over_react_test/pull/104) so that consumer tests asserting `propTypes` logging running in dart2js don't start to fail.

### Problem
`propTypes` are a dev-only tool that do nothing when `react_dom_with_react_dom_prod.js` is utilized. That production build of react is most commonly used alongside a production deploy of Dart code compiled with dart2js.

Currently, the `propTypes` maps remain in the dart2js output - adding unnecessary bulk to consumer module(s).

### Solution
Wrap the access to `Component2.propTypes` in an `assert()` so that they are included in JS compiled using DDC, but not dart2js.

### Boy Scoutin'
* Deprecated `registerComponent`.  This should have been deprecated when `Component` was - but we missed it.


test/util.dart
```diff
@@ -54,6 +54,12 @@ Map unmodifiableMap([Map map1, Map map2, Map map3, Map map4]) {
   return new Map.unmodifiable(merged);
 }
 
+bool isDDC() {
```
greglittlefield-wf, line 57: Same comment as in the other PR, this should be called `assertsEnabled` or something

aaronlademann-wf:
Just FYI - the dev SDK travis build is failing e'where - it is unrelated to the changes in this PR.

greglittlefield-wf (changes requested):

test/util.dart
aaronlademann-wf: Addressed in https://github.com/cleandart/react-dart/pull/258/commits/893bc4f052c20b519dcc28626e0895146c9596b1

greglittlefield-wf (approved):
+10

PR #259, opened 2020-07-14
Function Component defaultProps Updates

sydneyjodon-wk:
Updates for https://github.com/Workiva/over_react/pull/606


sydneyjodon-wk:
Decided not to implement `defaultProps` for function components

PR #261, opened 2020-07-20
CPLAT-10894 Rethrow and print registerComponent/registerComponent2 errors

smaifullerton-wk:
Recommended review: [Whitespace changes only](https://github.com/cleandart/react-dart/pull/261/files?diff=split&w=1)

## Problem
Currently, registerComponent calls that throw have their errors swallowed by React if they occur in the call stack of a component render (due to the render triggering lazy evaluation of a top-level factory variable).

This can commonly happen if:
- defaultProps throws
- propTypes throws
    - can happen when consumedProps throws

## Changes: 
- In `_registerComponent` and `_registerComponent2`, add a try-catch-rethrow that also logs the error, in case it's swallowed by React
- Use print instead of window.console.error so it shows up in test output
- In `_registerComponent2` only: add more specific try-catch-rethrows around defaultProps/propTypes accesses, since that's usually what throws 
- Add regression tests that use the `prints` matcher to verify the error is printed

## QA
- [ ] CI passes, unit test coverage should be sufficient unless @greglittlefield-wf feels otherwise

### DIFF
--- a/test/react_client_test.dart
+++ b/test/react_client_test.dart
@@ -176,7 +176,7 @@ main() {
 
   group('registerComponent', () {
     test('throws with printed error', () {
-      expect(() => react.registerComponent(() => ThrowsInDefaultPropsComponent()), throwsA(Error));
+      expect(() => react.registerComponent(() => ThrowsInDefaultPropsComponent()), throwsStateError);
       expect(() {
         try {
           react.registerComponent(() => ThrowsInDefaultPropsComponent());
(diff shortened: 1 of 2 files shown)


test/react_client_test.dart
```diff
@@ -173,6 +173,47 @@ main() {
...
+      expect(() {
+        try {
+          react.registerComponent2(() => ThrowsInPropTypesComponent2());
+        } catch (_) {}
+      }, prints(contains('Error when registering Component2 when getting propTypes')));
+    });
+
+    test('throws with generic error when something else throws', () {
+      expect(() => react.registerComponent2(() => DartComponent2Component(), bridgeFactory: (component) => throw Error),
+          throwsA(Error));
```
greglittlefield-wf, line 209: This is actually throwing the `Error` type instead of an actual error instance.

It works, but it's not really best practice and could confuse devs working on this code in the future.

Could we update this and other tests/`throw`s to not throw the types?

```suggestion
      expect(() => react.registerComponent2(() => DartComponent2Component(), bridgeFactory: (component) => throw StateError('bad bridge')),
          throwsStateError);
```

smaifullerton-wk:
@greglittlefield-wf thanks, ready

PR #262, opened 2020-07-21
Add chainRef/chainRefList utilities

greglittlefield-wf:
## Motivation
There are some cases in which you may want to set multiple refs on a component.

In particular, when using `forwardRef` or ref props, you may have a ref set by the consumer of the component in addition to an internal ref that need to be set on the same component.

## Solution
- Add utilities to "chain" multiple refs into one:
    - `chainRefs`
    - `chainRefList`
- Add tests

## Testing
- Verify tests pass (i.e., CI passes)

### DIFF
--- a/lib/react_client/react_interop.dart
+++ b/lib/react_client/react_interop.dart
@@ -91,6 +91,16 @@ class Ref<T> {
     }
     return jsCurrent;
   }
+
+  /// Used internally to combine refs https://github.com/facebook/react/issues/13029
+  @protected
+  set current(T value) {
+    if (value is Component) {
+      jsRef.current = value.jsThis;
+    } else {
+      jsRef.current = value;
+    }
+  }
 }
 
 /// A JS ref object returned by [React.createRef].
@@ -101,6 +111,10 @@ class Ref<T> {
 @anonymous
 class JsRef {
   external dynamic get current;
+
+  /// Used internally to combine refs https://github.com/facebook/react/issues/13029
+  @protected
+  external set current(dynamic value);
 }
 
 /// Automatically passes a [Ref] through a component to one of its children.
--- a/lib/react_client/react_interop.dart
+++ b/lib/react_client/react_interop.dart
@@ -82,8 +82,10 @@ class Ref<T> {
   T get current {
     final jsCurrent = jsRef.current;
 
-    if (jsCurrent is! Element) {
-      final dartCurrent = (jsCurrent as ReactComponent)?.dartComponent;
+    // Note: this ReactComponent check will pass for many types of JS objects,
+    // so don't assume for sure that it's a ReactComponent
+    if (jsCurrent is! Element && jsCurrent is ReactComponent) {
+      final dartCurrent = jsCurrent.dartComponent;
 
       if (dartCurrent != null) {
         return dartCurrent as T;
--- a/lib/src/react_client/chain_refs.dart
+++ b/lib/src/react_client/chain_refs.dart
@@ -1,0 +1,83 @@
+import 'package:react/react_client/react_interop.dart';
+
+/// Returns a ref that updates both [ref1] and [ref2], effectively
+/// allowing you to set multiple refs.
+///
+/// Useful when you're using ref forwarding and want to also set your own ref.
+///
+/// For more information on this problem, see https://github.com/facebook/react/issues/13029.
+///
+/// Inputs can be callback refs, [createRef]-based refs ([Ref]),
+/// or raw JS `createRef` refs ([JsRef]).
+///
+/// If either [ref1] or [ref2] is null, the other ref will be passed through.
+///
+/// If both are null, null is returned.
+///
+/// ```dart
+/// final Example = forwardRef((props, ref) {
+///   final localRef = useRef<FooComponent>();
+///   return Foo({
+///     ...props,
+///     'ref': mergeRefs(localRef, ref),
+///   });
+/// });
+/// ```
+dynamic chainRefs(dynamic ref1, dynamic ref2) {
+  return chainRefList([ref1, ref2]);
+}
+
+/// Like [chainRefs], but takes in a list of [refs].
+dynamic chainRefList(List<dynamic> refs) {
+  refs.forEach(_validateChainRefsArg);
+
+  final nonNullRefs = refs.where((ref) => ref != null).toList();
+
+  // Return null if there are no refs to chain
+  if (nonNullRefs.isEmpty) return null;
+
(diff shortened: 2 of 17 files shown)


lib/src/react_client/chain_refs.dart
```diff
@@ -0,0 +1,83 @@
...
+///   });
+/// });
+/// ```
+dynamic chainRefs(dynamic ref1, dynamic ref2) {
+  return chainRefList([ref1, ref2]);
+}
+
+/// Like [chainRefs], but takes in a list of [refs].
+dynamic chainRefList(List<dynamic> refs) {
+  refs.forEach(_validateChainRefsArg);
```
aaronlademann-wf, line 32: Any reason _(other than the lack of individual error messages)_ to not wrap this whole thing in an assert so the refs are not iterated at all in dart2js?

```dart
dynamic chainRefList(List<dynamic> refs) {
  assert(refs.every(_isChainedRefValid));

  // ...
}

bool _isChainedRefValid(dynamic ref) {
  if (ref is Function) return ref is! Function(Null);
  return ref is! String;
}
```

greglittlefield-wf: Good point! I'll update this

test/factory/common_factory_tests.dart
```diff
@@ -268,15 +269,11 @@ void refTests<T>(ReactComponentFactoryProxy factory, {void verifyRefValue(dynami
...
-  _typedCallbackRefTests<T>(factory);
-}
-
-void _typedCallbackRefTests<T>(react.ReactComponentFactoryProxy factory) {
-  if (T == dynamic) {
-    throw ArgumentError('Generic parameter T must be specified');
-  }
-
   group('has functional callback refs when they are typed as', () {
+    if (T == dynamic) {
```
smaifullerton-wk, line 273: I'm missing something here - what `T` is this referencing?

lib/react_client/react_interop.dart
```diff
@@ -91,6 +91,16 @@ class Ref<T> {
...
+
+  /// Used internally to combine refs https://github.com/facebook/react/issues/13029
+  @protected
+  set current(T value) {
+    if (value is Component) {
+      jsRef.current = value.jsThis;
+    } else {
+      jsRef.current = value;
+    }
+  }
```
sydneyjodon-wk, line 103: Not sure if this is a big deal or not, but in `5.4.0-wip` [this setter](https://github.com/cleandart/react-dart/blob/5.4.0-wip/lib/react_client/react_interop.dart#L123-L126) isn't just used internally. Just mentioning in case there's some bad outcome of it no longer being internal when the other branch merges.

test/factory/common_factory_tests.dart
greglittlefield-wf: `T` is the generic parameter to `refTests`, which represents the expected type of the ref

lib/react_client/react_interop.dart
greglittlefield-wf: Mm good to know. 

I think we should try and keep it `@protected` if it's not used outside of over_react, so that people don't misuse it

greglittlefield-wf: We might have to follow up after this and fix usages of `.current` (assuming this merges first), since the implementation is different.

lib/src/react_client/chain_refs.dart
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+
+  return _chainedRef;
+}
+
+void _validateChainRefsArg(dynamic ref) {
+  assert(ref is! Function || ref is Function(Null), 'callback refs must take a single argument');
+  assert(ref is! String, 'String refs cannot be chained');
+}
```
sydneyjodon-wk, line 87: Can we assert here that `ref` is either a `Ref` or a callback ref? Or could `ref` be other types? It just seems weird that if I pass in something like an integer in the list of refs there are no failures.

test/factory/common_factory_tests.dart
```diff
@@ -299,6 +296,106 @@ void _typedCallbackRefTests<T>(react.ReactComponentFactoryProxy factory) {
...
+        {'name': 'ref3', 'value': anything},
+        {'name': 'ref4', 'value': anything},
+      ]);
+      verifyRefValue(calls[0]['value']);
+      verifyRefValue(calls[1]['value']);
+      verifyRefValue(calls[2]['value']);
+      verifyRefValue(calls[3]['value']);
+    });
+
+    // Other cases tested in mergeRefList's own tests
```
sydneyjodon-wk, line 397: #nit
```suggestion
    // Other cases tested in chainRefList's own tests
```

test/react_client/chain_refs_test.dart
```diff
@@ -0,0 +1,63 @@
...
+
+      group('asserts that inputs arg are not', () {
+        test('strings', () {
+          expect(() => chainRefList([testRef, 'bad ref']), throwsA(isA<AssertionError>()));
+        });
+
+        test('unsupported function types', () {
+          expect(() => chainRefList([testRef, () {}]), throwsA(isA<AssertionError>()));
+        });
+      }, tags: 'no-dart2js');
```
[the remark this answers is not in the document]
greglittlefield-wf: No, by default `assert` functions are compiled out in dart2js, so we use them in places where we want to check in dev mode but avoid running the code in prod

lib/src/react_client/chain_refs.dart
greglittlefield-wf: Unfortunately we can't because type-checking on `JsRef` will return `true` for all cases, at least last time I checked. 

I'll try adding more checks to see if that works now, and if not, I'll add a comment with more info

greglittlefield-wf: Addressed in ce82545 and also added better test coverage for JS refs, which I discovered weren't really being tested fully

lib/react_client/react_interop.dart
```diff
@@ -82,8 +82,10 @@ class Ref<T> {
   T get current {
     final jsCurrent = jsRef.current;
 
-    if (jsCurrent is! Element) {
-      final dartCurrent = (jsCurrent as ReactComponent)?.dartComponent;
+    // Note: this ReactComponent check will pass for many types of JS objects,
+    // so don't assume for sure that it's a ReactComponent
+    if (jsCurrent is! Element && jsCurrent is ReactComponent) {
```
greglittlefield-wf, line 87: I added this `is ReactComponent` check since using the Dart wrapper with a non-component value threw (which may not much in practice, but is a total valid use of `createRef`).

PR #263, opened 2020-07-21
Proxy event.persist() when wrapping synthetic JS events

aaronlademann-wf:
I'm resurrecting #249, as I've run into a more concrete use-case for `event.persist()` than the one that prompted #249.

Since component updates are not guaranteed to be synchronous in React 16+, the optional callback for `setState` may be called asynchronously on the JS side of things. This means that doing something like passing a `SyntheticEvent` to a consumer callback within that callback will result in ReactJS spitting out a big ugly warning about the event having been recycled.

__Without persist()__
```dart
void handleClick(SyntheticMouseEvent event) {
  setState({}, () {
    // Ruh roh Scoobs... this could be recycled when they access it.
    props.someConsumerCallbackThatNeedsTheEvent(event);
  });
}
```

__With persist()__
```dart
void handleClick(SyntheticMouseEvent event) {
  event.persist();
  setState({}, () {
    // All good... event is persisted and accessible for the consumer.
    props.someConsumerCallbackThatNeedsTheEvent(event);
  });
}
```

### DIFF
--- a/lib/react.dart
+++ b/lib/react.dart
@@ -1451,6 +1451,61 @@ class SyntheticEvent {
   /// See: <https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation>
   final dynamic stopPropagation;
 
+  /// For use by react-dart internals only.
+  @protected
+  void Function() $$jsPersistDoNotSetThisOrYouWillBeFired;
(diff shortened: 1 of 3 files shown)


greglittlefield-wf:
QA +1

PR #268, opened 2020-07-28
Update memo signature

smaifullerton-wk:
Updates the `memo` signature to accept a `ReactDartFunctionComponentFactoryProxy` directly, since only function components can be memoized. 

Also removes the `displayName` property as it is not part of React's `memo`, and should be set by the function component config instead.


lib/react_client/react_interop.dart
```diff
@@ -288,28 +289,16 @@ ReactJsComponentFactoryProxy forwardRef(
 /// > Do not rely on it to “prevent” a render, as this can lead to bugs.
 ///
 /// See: <https://reactjs.org/docs/react-api.html#reactmemo>.
-ReactJsComponentFactoryProxy memo(
```
greglittlefield-wf, line 291: The doc comment also needs to be updated

PR #269, opened 2020-07-31
Ensure `forwardRef`'s `ReactJsComponentFactoryProxy` always uses List children

joebingham-wk:
## Motivation
While adding onto the `forwardRef` API in OverReact, it was discovered that consuming `props.children` in the `forwardRef` HOC threw a type error because the children weren't guaranteed to be a List as expected.

## Changes
- Utilize the `alwaysReturnChildrenAsList` flag on `ReactJsComponentFactoryProxy` to ensure children are always wrapped in a List
- Add tests


test/factory/common_factory_tests.dart
```diff
@@ -320,6 +320,72 @@ void refTests<T>(ReactComponentFactoryProxy factory, {void verifyRefValue(dynami
...
 
+  group('forwardRef wraps children in a list when the child a is:', () {
+    final simpleFactory = factory({
+      'id': 'test',
+    });
+
+    var ForwardRefTestComponent = forwardRef((props, ref) {
+      // Extra type checking since JS refs being passed through
+      // aren't caught by built-in type checking.
+      expect(ref, isA<Ref>());
```
greglittlefield-wf, line 331: These `expect`s get called in the component zone, not the test zone, so if they throw they won't fail the test or show up in the console output. I think they actually throw right now with "expect may only be called within tests", but the error is swallowed.

Can we move these `expect`s out of the render call-stack? (e.g., by storing ref/children in closure variables or something).

PR #271, opened 2020-08-03
Run function/forwardRef components in component zone, run children tests

greglittlefield-wf:
## Motivation
- Render logic for function and `forwardRef` components weren't being run in the component zone, which was inconsistent with class-based components.
- Existing children tests that verify children are always `List`s in Dart weren't being run for function and `forwardRef` components

## Solution
- Add component zone logic and tests to `commonFactoryTests`
- Run `commonFactoryTests` for `forwardRef` components
- Update children tests to run for function and `forwardRef` components

### Testing
- Tests pass
- Revert changes in https://github.com/cleandart/react-dart/pull/269 and verify that children tests fail

### DIFF
--- a/test/factory/common_factory_tests.dart
+++ b/test/factory/common_factory_tests.dart
@@ -22,7 +22,11 @@ import 'package:react/react_client/react_interop.dart';
 
 import '../util.dart';
 
-void commonFactoryTests(ReactComponentFactoryProxy factory, {bool isFunctionComponent = false}) {
+/// Runs common tests for [factory].
+///
+/// [dartComponentVersion] should be specified for all components with Dart render code in order to
+/// properly test `props.children`, forwardRef compatibility, etc.
+void commonFactoryTests(ReactComponentFactoryProxy factory, {bool isFunctionComponent = false, String dartComponentVersion}) {
   _childKeyWarningTests(
     factory,
     renderWithUniqueOwnerName: _renderWithUniqueOwnerName,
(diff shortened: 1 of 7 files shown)


test/factory/common_factory_tests.dart
```diff
@@ -103,17 +112,43 @@ void commonFactoryTests(ReactComponentFactoryProxy factory, {bool isFunctionComp
         shouldAlwaysBeList: isDartComponent2(factory({})));
   });
 
-  // Skipped for Function Components because they dont have instance members ... they are functions.
```
greglittlefield-wf, line 106: Needed to get creative here to work around this restriction; ended up using the `onDartRender` callback to capture the actual children passed to render via props

test/factory/common_factory_tests.dart
```diff
@@ -284,10 +319,9 @@ void refTests<T>(ReactComponentFactoryProxy factory, {void verifyRefValue(dynami
   });
 
   test('forwardRef function passes a ref through a component to one of its children', () {
+    dynamic actualRef;
     var ForwardRefTestComponent = forwardRef((props, ref) {
-      // Extra type checking since JS refs being passed through
-      // aren't caught by built-in type checking.
-      expect(ref, isA<Ref>());
```
greglittlefield-wf, line 290: Needed to move this since `expect`s don't work in the default component zone

test/factory/common_factory_tests.dart
```diff
@@ -64,41 +66,41 @@ void commonFactoryTests(ReactComponentFactoryProxy factory, {bool isFunctionComp
 
         test('$childrenCount', () {
           final expectedChildren = new List.generate(childrenCount, (i) => i + 1);
-          final arguments = <dynamic>[{...props}]..add(expectedChildren);
+          final arguments = <dynamic>[props, ...expectedChildren];
```
greglittlefield-wf: Uhh this was actually an bug in the test I didn't notice until I went to convert it to use ui-as-code syntax; it should have been `..addAll(expectedChildren)`

PR #273, opened 2020-08-07
Make JsBackedMap.values MSIE 11 compatible

aaronlademann-wf:
`Object.values` is not supported in MSIE 11. We still have consumers that need to support the browser, so our `JsBackedMap.values` implementation needs to support it.

Currently, if you access the `values` getter ([as we do within `forwardRef`](https://github.com/cleandart/react-dart/blob/5.5.0-wip/lib/react_client/react_interop.dart#L238)), MSIE 11 will choke. This fixes that.

@greglittlefield-wf

### DIFF
--- a/lib/react_client/js_backed_map.dart
+++ b/lib/react_client/js_backed_map.dart
@@ -38,7 +38,8 @@ class JsBackedMap extends MapBase<dynamic, dynamic> {
 
   // Private helpers with narrower typing than we want to expose, for use in other methods
   List<dynamic> get _keys => _Object.keys(jsObject);
-  List<dynamic> get _values => _Object.values(jsObject);
+  // Use keys to access the value instead oof `Object.values` since MSIE 11 doesn't support it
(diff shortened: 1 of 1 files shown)


lib/react_client/js_backed_map.dart
```diff
@@ -38,7 +38,8 @@ class JsBackedMap extends MapBase<dynamic, dynamic> {
 
   // Private helpers with narrower typing than we want to expose, for use in other methods
   List<dynamic> get _keys => _Object.keys(jsObject);
-  List<dynamic> get _values => _Object.values(jsObject);
+  // Use keys to access the value instead oof `Object.values` since MSIE 11 doesn't support it
+  List<dynamic> get _values => _keys.map((key) => this[key]).toList();
```
greglittlefield-wf, line 42: I'd like to see us eventually polyfill `Object.values` on the JS side so that this can be potentially more optimized in browsers that support it, but this is totally fine for now

PR #276, opened 2020-09-01
useImperativeHandle should not throw if ref is null

aaronlademann-wf:
I came across this RTE recently that occurs during mount as a result of us accessing `jsRef` on the `Ref` when passing through to the  `React.useImperativeHandle` JS interop layer.

Instead of throwing, these changes make it so that the creation of the imperative handle is simply a no-op unless the `Ref` provided is not null.

### DIFF
--- a/lib/hooks.dart
+++ b/lib/hooks.dart
@@ -529,8 +529,10 @@ void useLayoutEffect(dynamic Function() sideEffect, [List<Object> dependencies])
 /// ```
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#useimperativehandle>.
-void useImperativeHandle(Ref ref, dynamic Function() createHandle, [List<dynamic> dependencies]) =>
-    React.useImperativeHandle(ref.jsRef, allowInterop(createHandle), dependencies);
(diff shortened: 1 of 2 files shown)


lib/hooks.dart
```diff
@@ -529,8 +529,10 @@ void useLayoutEffect(dynamic Function() sideEffect, [List<Object> dependencies])
 /// ```
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#useimperativehandle>.
-void useImperativeHandle(Ref ref, dynamic Function() createHandle, [List<dynamic> dependencies]) =>
-    React.useImperativeHandle(ref.jsRef, allowInterop(createHandle), dependencies);
+void useImperativeHandle(Ref ref, dynamic Function() createHandle, [List<dynamic> dependencies]) {
+  if (ref == null) return;
+  React.useImperativeHandle(ref.jsRef, allowInterop(createHandle), dependencies);
```
sydneyjodon-wk, line 534: Just out of curiousity, would there be a benefit to passing in `ref?.jsRef` and letting React's `useImperativeHandle` handle the `null` input? Or is it better to handle it on our end and no-op as soon as possible?

aaronlademann-wf: I would vote to short-circuit before making an unnecessary interop call.

PR #277, opened 2020-09-04
CPLAT-12469 Update to React 17

joebingham-wk:
- Updated to React 17
- Removed deprecated APIs that are no longer in use:
  - `setReactConfiguration` - [sourcegraph](https://sourcegraph.wk-dev.wdesk.org/search?q=lang:dart+setReactConfiguration+count:1000&patternType=regexp)
  - `setReactDOMConfiguration` - [sourcegraph](https://sourcegraph.wk-dev.wdesk.org/search?q=lang:dart+setReactDOMConfiguration+count:1000&patternType=regexp)
  - `ReactDartComponentFactoryProxy.reactComponentFactory` and `ReactDartComponentFactoryProxy2.reactComponentFactory` - [sourcegraph](https://sourcegraph.wk-dev.wdesk.org/search?q=lang:dart+%5C.reactComponentFactory+count:1000&patternType=regexp)
  - `ReactJsContextComponentFactoryProxy.factory`, `ReactJsComponentFactoryProxy.factory`, and `ReactDomComponentFactoryProxy.factory` - [sourcegraph](https://sourcegraph.wk-dev.wdesk.org/search?q=lang:dart+%5C.factory+count:1000&patternType=regexp)

## QA Instructions
- [ ] verify CI passes
- [ ] verify that [over_react CI](https://github.com/Workiva/over_react/pull/637) is passing
- [ ] verify examples work as expected

Please review: @joebingham-wk @aaronlademann-wf @greglittlefield-wf


joebingham-wk:
@kealjones-wk

js_src/_dart_helpers.js
```diff
@@ -131,10 +131,21 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
...
+      // TODO: figure out how to get the `displayName` here...
+      React.PropTypes.checkPropTypes(jsConfig.propTypes, this.props, 'prop', this._getDisplayName());
       var result = dartInteropStatics.handleRender(this.dartComponent, this.props, this.state, this.context);
       if (typeof result === 'undefined') result = null;
       return result;
     }
+    // Hacky workaround to get the displayName for `checkPropTypes` call.
+    _getDisplayName() {
+      return ReactDartComponent2.displayName;
+    }
```
kealjones-wk, line 148: It should be noted that I opted for this route to prevent breakages in the `PropTypes` api, specifically the `resetWarningCache()` react has opted to build in their own "prop-types" rather than using the external package. Currently react does not provide a way to reset the warning cache of itself, which is why i opted to use the external package. It is also unclear what the future hold for prop-types in general so the external package is a safer bet for long term support as the React team feels that Typescript on its own provides most of the prop type checks statically, unfortunately dart types do not support things such as Literal Types, Tuples, Unions, Optional/Required fields (i mean they kinda do but not as nicely), etc...

sydneyjodon-wk:
Dart stable is failing from master, but `2.9.2` is passing.

sydneyjodon-wk (approved):

js_src/_dart_helpers.js
```diff
@@ -131,6 +131,12 @@ function _createReactDartComponentClass2(dartInteropStatics, componentStatics, j
       return typeof derivedState !== 'undefined' ? derivedState : null;
     }
     render() {
+      // Must call checkPropTypes manually because React moved away from using the `prop-types` package.
+      // See: https://github.com/facebook/react/pull/18127
+      // React now uses its own internal cache of errors for PropTypes which broke `PropTypes.resetWarningCache()`.
+      // Solution was to use `PropTypes.checkPropTypes` directly which makes `PropTypes.resetWarningCache()` work.
+      // Solution from: https://github.com/facebook/react/issues/18251#issuecomment-609024557
+      React.PropTypes.checkPropTypes(jsConfig.propTypes, this.props, 'prop', ReactDartComponent2.displayName);
```
greglittlefield-wf, line 139: Two things:
1. can we wrap this in a compile-time check so that this call isn't present in the dev bundle?
```suggestion
      if (process.env.NODE_ENV !== 'production') {
        React.PropTypes.checkPropTypes(jsConfig.propTypes, this.props, 'prop', ReactDartComponent2.displayName);
      }
```
2. Can we update the propTypes tests to verify that the correct displayName is being passed? RIght now we capture all of the information passed into prop validators but [only test some of it](https://github.com/cleandart/react-dart/blob/1d3fea5da67fa46b2e95747e684c1296376fc6e2/test/lifecycle_test.dart#L166)

greglittlefield-wf (changes requested):

js_src/_dart_helpers.js
sydneyjodon-wk: It seems like the displayName is being tested [here](https://github.com/cleandart/react-dart/blob/1d3fea5da67fa46b2e95747e684c1296376fc6e2/test/lifecycle_test.dart#L198)? Do you want me to add that check in both places?

greglittlefield-wf: Oh good catch! It looks like that test would pass if the name is `ReactDartComponent2`, though, and we wouldn't be able to tell if the wrong name is being passed in.

Now that it's wrapped in that `assertsEnabled()` check, though (which wasn't present when the original test was written), I think that case shouldn't happen, and we could update it like so:
```diff
-'componentName': anyOf('PropTypesTestComponent', 'ReactDartComponent2'),
+'componentName': 'PropTypesTestComponent',
```

and then we'd know for sure that the test name is being passed through.

Would you mind updating that test so that we get full coverage on this code?

sydneyjodon-wk: Updated in 2978197!

greglittlefield-wf (approved):
+10

PR #282, opened 2020-11-05
CPLAT-12688 Add synthetic Event Utils to master

joebingham-wk:
To prepare for React 17, there needs to be APIs that abstract the creation of SyntheticEvents so that we can remove dependence on the event class constructors. 

### Changes
- Add a `createSynthetic{eventClass}` function for every synthetic event type that allows consumers to create an empty instance, set specific properties on an instance, or merge an instance with new properties
- Add a simple test for every API to ensure that the getters / setters and merge logic is working as expected

### QA
- [ ] Review that tests correctly test getting and setting
- [ ] Verify consumer examples of using the APIs work as expected (passing CI should be sufficient)
  - [dataset-service-ui](https://github.com/Workiva/dataset-service-ui/pull/120)
  - [graph_ui](https://github.com/Workiva/graph_ui/pull/1804)


lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,324 @@
...
+SyntheticAnimationEvent getEmptyAnimationEvent() => buildAnimationEvent();
+SyntheticUIEvent getEmptyUIEvent() => buildUIEvent();
+SyntheticWheelEvent getEmptyWheelEvent() => buildWheelEvent();
+
+/// Returns a newly constructed [SyntheticEvent] instance.
+///
+/// In addition to creating empty instances (by invoking without any parameters) or completely custom instances
+/// by using [properties], this method will merge previously existing [baseEvent]s with the [properties] provided.
+/// [properties] takes precedence, and therefore can be used to override specific fields on the [baseEvent].
+SyntheticEvent buildSyntheticEvent({SyntheticEvent baseEvent, Map<String, dynamic> properties = const {}}) {
```
greglittlefield-wf, line 27: #nit I think "create" would be better in this context; "build" might imply there's a builder object involved."

So, I'd prefer `createSyntheticEvent` instead of `buildSyntheticEvent`.

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,324 @@
+// ignore_for_file: deprecated_member_use_from_same_package
+
+import '../../react.dart';
+
+/// Returns an empty [SyntheticEvent] with all fields set to `null`.
+///
+/// This utility can be useful when a test just needs access to an event object,
+/// but no fields need to be set.
+SyntheticEvent getEmptySyntheticEvent() => buildSyntheticEvent();
```
greglittlefield-wf, line 9: I don't think these "empty" methods provide enough value to justify having them; for example, why not just call `buildSyntheticEvent" directly?

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,324 @@
...
+    properties['eventPhase'] ?? baseEvent?.eventPhase,
+    properties['isTrusted'] ?? baseEvent?.isTrusted,
+    properties['nativeEvent'] ?? baseEvent?.nativeEvent,
+    properties['target'] ?? baseEvent?.target,
+    properties['timeStamp'] ?? baseEvent?.timeStamp ?? 0,
+    properties['type'] ?? baseEvent?.type ?? 'empty event',
+  );
+}
+
+SyntheticClipboardEvent buildClipboardEvent({SyntheticClipboardEvent baseEvent, Map<String, dynamic> properties = const {}}) {
```
greglittlefield-wf, line 44: I think the rest of these should also have "synthetic" in the name, for consistency and to make it clear that they're returning synthetic events and not native ones

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,324 @@
...
+    properties['type'] ?? baseEvent?.type ?? 'empty event',
+    properties['deltaX'] ?? baseEvent?.deltaX,
+    properties['deltaMode'] ?? baseEvent?.deltaMode,
+    properties['deltaY'] ?? baseEvent?.deltaY,
+    properties['deltaZ'] ?? baseEvent?.deltaZ,
+  );
+}
+
+extension SyntheticEventTypeHelpers on SyntheticEvent {
+  bool get isClipboardEvent => this.tryCast<SyntheticClipboardEvent>() != null;
```
greglittlefield-wf, line 306: #nit tryCast isn't necessary here
```suggestion
  bool get isClipboardEvent => this is SyntheticClipboardEvent;
```

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,324 @@
...
+SyntheticWheelEvent getEmptyWheelEvent() => buildWheelEvent();
+
+/// Returns a newly constructed [SyntheticEvent] instance.
+///
+/// In addition to creating empty instances (by invoking without any parameters) or completely custom instances
+/// by using [properties], this method will merge previously existing [baseEvent]s with the [properties] provided.
+/// [properties] takes precedence, and therefore can be used to override specific fields on the [baseEvent].
+SyntheticEvent buildSyntheticEvent({SyntheticEvent baseEvent, Map<String, dynamic> properties = const {}}) {
+  return SyntheticEvent(
+    properties['bubbles'] ?? baseEvent?.bubbles,
```
greglittlefield-wf, line 29: Since we already have to enumerate out all of the properties, I think it'd be helpful to have these as named arguments so that consumers get static analysis when they're setting them and can't accidentally misspell key names or provide a value with the wrong type.

So, instead of:
```dart
        SyntheticEvent event = buildSyntheticEvent(properties: {
          'currentTarget': element,
          'target': element,
        });
```
it would be:
```dart
        SyntheticEvent event = buildSyntheticEvent(
          currentTarget: element,
          target: element,
        );
```

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,324 @@
...
+    properties['cancelable'] ?? baseEvent?.cancelable,
+    properties['currentTarget'] ?? baseEvent?.currentTarget,
+    properties['defaultPrevented'] ?? baseEvent?.defaultPrevented,
```
greglittlefield-wf, line 32: Hmm I feel like we should add non-nullable defaults for these boolean properties while we have the chance to do so, to help prevent null errors in the future.

lib/src/react_client/event_helpers.dart
joebingham-wk: That's definitely fair. I went back and forth myself and my thinking was that from what I saw, this was actually the most common use-case and the intent was to create an empty event. So in application, I felt like this would be the main API and read the best. In contrast `buildSyntheticEvent` to me implied that there's some work to be done like wrapping a pre-existing event or adding specific fields.

I see what you mean though. Like I mentioned, I went back and forth and landed on keeping them so I'd still personally prefer that. Do you feel like they'll ultimately cause more confusion or maintenance? If that's the case I can remove them!

lib/src/react_client/event_helpers.dart
joebingham-wk: [But proxies, Greg, proxy all the things!](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition/blob/uinverse/src/main/java/com/seriouscompany/business/java/fizzbuzz/packagenamingpackage/impl/Main.java)

Jk 🤦

lib/src/react_client/event_helpers.dart
joebingham-wk: That's a really good point, I will add that!

joebingham-wk:
For CI, Dart 2.10.x fails on the formatting step - my understanding is that this was known and ignored that for now. Let me know if there is something to do there though!

lib/src/react_client/event_helpers.dart
greglittlefield-wf: I'd prefer to remove them so that we're not providing two ways to do the same thing, and keeping the APIs as simple as possible. Also, if they're using the `createSyntheticEvent` APIs with no arguments, and then later on determine they need to customize a field, all they have to do is add an argument as opposed to having to switch methods.

lib/react.dart
```diff
@@ -1584,6 +1588,7 @@ class SyntheticEvent {
   /// See: <https://developer.mozilla.org/en-US/docs/Web/API/Event/type>
   final String type;
 
+  @Deprecated(_syntheticEventDeprecationMessage)
```
greglittlefield-wf, line 1591: It'd be helpful to say which APIs specifically to switch to, so that they can just autocomplete/auto-import them. Right now, they'd have to dig through the source code to find them

> Use `createSyntheticEvent` instead.

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,701 @@
...
+  dynamic stopPropagation,
+  num eventPhase,
+  bool isTrusted,
+  dynamic nativeEvent,
+  dynamic target,
+  num timeStamp,
+  String type,
+}) {
+  return SyntheticEvent(
+    bubbles ?? baseEvent?.bubbles ?? true,
```
greglittlefield-wf, line 104: #nit I think this should default to `false`, because most artificially-created synthetic events will never be dispatched through React's event system and thus won't bubble.

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,701 @@
...
+    cancelable ?? baseEvent?.cancelable ?? true,
+    currentTarget ?? baseEvent?.currentTarget,
+    defaultPrevented ?? baseEvent?.defaultPrevented ?? false,
+    preventDefault ?? baseEvent?.preventDefault ?? () {},
+    stopPropagation ?? baseEvent?.stopPropagation ?? () {},
+    eventPhase ?? baseEvent?.eventPhase,
+    isTrusted ?? baseEvent?.isTrusted ?? true,
```
greglittlefield-wf, line 111: I think this should default to `false`

https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted
> The isTrusted read-only property of the Event interface is a Boolean that is true when the event was generated by a user action, and false when the event was created or modified by a script or dispatched via EventTarget.dispatchEvent().

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,701 @@
...
+    nativeEvent ?? baseEvent?.nativeEvent,
+    target ?? baseEvent?.target,
+    timeStamp ?? baseEvent?.timeStamp ?? 0,
+    type ?? baseEvent?.type ?? 'empty event',
+    altKey ?? baseEvent?.altKey,
```
greglittlefield-wf, line 201: #nit can we default altKey/ctrlKey/metaKey/shiftKey to false as well?

lib/src/react_client/event_helpers.dart
```diff
@@ -0,0 +1,701 @@
...
+/// by using the named parameters, this function will merge previously existing [baseEvent]s with the named parameters provided.
+/// The named parameters takes precedence, and therefore can be used to override specific fields on the [baseEvent].
+SyntheticEvent createSyntheticEvent({
+  SyntheticEvent baseEvent,
+  bool bubbles,
+  bool cancelable,
+  dynamic currentTarget,
+  bool defaultPrevented,
+  dynamic preventDefault,
+  dynamic stopPropagation,
```
greglittlefield-wf, line 95: These should be typed
```suggestion
  void Function() preventDefault,
  void Function() stopPropagation,
```

greglittlefield-wf (changes requested):
Just a couple more small things; this is looking great!

lib/react.dart
joebingham-wk: True, good point!

lib/src/react_client/event_helpers.dart
joebingham-wk: Ahh yep, thank you!

lib/src/react_client/event_helpers.dart
joebingham-wk: I'll definitely tighten it - Naturally I originally had it `dynamic` to match the constructor typing. Was there a reason it was originally typed as `dynamic`?

greglittlefield-wf: Yeah, I think it was just an oversight when it was originally written

greglittlefield-wf (approved):

test/react_client/event_helpers_test.dart
```diff
@@ -0,0 +1,999 @@
...
+          isTrusted: false,
+          nativeEvent: 'updated string',
+          target: newElement,
+          timeStamp: 200,
+          type: 'updated non-default',
+          clipboardData: 'new data',
+        );
+
+        testSyntheticEventBaseAfterMerge(newEvent);
+        expect(baseEvent.clipboardData, 'initial data');
```
sydneyjodon-wk, line 193:
```suggestion
        expect(newEvent.clipboardData, 'new data');
```

test/react_client/event_helpers_test.dart
```diff
@@ -0,0 +1,999 @@
...
+          currentTarget: newElement,
+          defaultPrevented: false,
+          preventDefault: () => mergeTestCounter++,
+          stopPropagation: () => mergeTestCounter++,
+          eventPhase: 2,
+          isTrusted: false,
+          nativeEvent: 'updated string',
+          target: newElement,
+          timeStamp: 200,
+          type: 'updated non-default',
```
sydneyjodon-wk, line 135: I might have missed them, but are there any tests that verify that if a named parameter is not set in `newEvent`, it will default to `baseEvent`'s property? (i.e. if `type` wasn't set here, then `newEvent.type` would be `'non-default'`?)

sydneyjodon-wk, line 135: Also related question: if we specifically wanted to set `type` to `null` here, would `newEvent.type` still end up being `'non-default'` because of the `baseEvent`?

sydneyjodon-wk (commented):
Just a couple small questions.

test/react_client/event_helpers_test.dart
sydneyjodon-wk: Actually looking at it again, I think `type` was a bad example since it has a default value when null. Maybe something like `target` though - that could end up being null - though I'm not sure if it's a common case to want to override something in `baseEvent` with null?

joebingham-wk: Ahh nope, good catch. I added those!

joebingham-wk: You're correct that yep, a consumer couldn't pass `null` in to unset a value. I also thought that it wasn't a common use case and could usually be worked around by not passing in a `baseEvent` and instead just pass along only the values that need to be set. In the future if it did become common, I think a change could be just to add a new parameter to the APIs that allows consumers to not get any defaults at all

sydneyjodon-wk (approved):
+1 on 25718c7 and 02e3b49

QA in progress..

sydneyjodon-wk:
QA+1 Everything looks great!
> - [x] Review that tests correctly test getting and setting
> - [x] Verify consumer examples of using the APIs work as expected (passing CI should be sufficient)
>   - [x] [dataset-service-ui](https://github.com/Workiva/dataset-service-ui/pull/120)
>   - [x] [graph_ui](https://github.com/Workiva/graph_ui/pull/1804)

joebingham-wk:
Verified CI failed due to that same Dart 2.10.x formatting step!

PR #283, opened 2020-11-09
Release React 5.7.0

joebingham-wk:
The latest version of react includes:
- #282 Added `SyntheticEvent` utils that remove the need to use the class constructors as well as assist in type checking event types


aaronlademann-wf (approved):
+1

joebingham-wk:
The Dart 2.10.x step failed (as usual), but the other version runs passed

PR #220, opened 2019-10-10
Fix dataTransfer RTE take 2

greglittlefield-wf:
Work around issue in #154 that necessitated revert of #151

PR #284, opened 2020-11-09
CPLAT-12665 Remove Synthetic Event Wrappers

joebingham-wk:
### Motivation
In order to make updating to React 17 easy, we need to change the internals of the publicly exposed event helper utils to construct anonymous interop object as opposed to using the wrapper constructors.

__NOTE:__ There was some follow up work discovered while working the ticket. These are:
- Some fields are missing from our event interop classes. These are described in [this ticket](https://jira.atl.workiva.net/browse/CPLAT-12758)
- `SyntheticFormEvent` is not a class in React.dart. It still works as a wrapper, but it seems like an opportunity to clean up some dead code since the interface is empty anyway. That discussion can be found [here](https://github.com/cleandart/react-dart/pull/284#discussion_r524407985).

### Changes
- change the `create{x}Event` utilities to create maps cast as the synthetic event type
- update the `SyntheticEventTypeHelpers` helpers to use duck typing instead of `is`. 
- Add a helper for `SyntheticMouseEvent.dataTransfer` that takes the interoped field and wraps it in a Dart wrapper, running the conversion logic that was used for the original synthetic event Dart classes

### QA
- [ ] verify test coverage still properly reflects how the utilities should be used. Tests for the `SyntheticEventTypeHelpers` helpers were added
- [ ]  verify upgrading between the APIs is as seamless as possible. Below are two test cases that showed what the usage looks like with React.dart stable as well as with the React 17 WIP branches.
  - data-service-ui:
    - [Stable](https://github.com/Workiva/dataset-service-ui/pull/120)
    - [React 17](https://github.com/Workiva/dataset-service-ui/pull/121)
  - graph_ui
    - [Stable](https://github.com/Workiva/graph_ui/pull/1804)
    - [React 17](https://github.com/Workiva/graph_ui/pull/1811)


lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -50,26 +64,35 @@ class SyntheticKeyboardEvent extends SyntheticEvent {
...
+  /// Use `createSyntheticFormEvent` instead.
+  external factory SyntheticFormEvent._();
+}
 
 /// A JS object that looks like a [DataTransfer] but isn't one.
-///
-/// See `syntheticDataTransferFactory` for more info.
 @JS()
-@anonymous
 class NonNativeDataTransfer {
```
joebingham-wk, line 96: I excluded this because it seemed like a special case since it isn't actually an event, but let me know if it should be treated like one and receive a `createNonNativeDataTransfer` API and have it's constructor overridden.

greglittlefield-wf: Crap, I forgot about this... 

tl;dr sometimes the `dataTransfer` property we get from React is a native `DataTransfer` object, and sometimes it's not, and mistyping it either way has caused issues in the past.

So, below, I don't think we can type `SyntheticMouseEvent.dataTransfer` as `NonNativeDataTransfer`.

See https://github.com/cleandart/react-dart/pull/151, https://github.com/cleandart/react-dart/pull/154, and https://github.com/cleandart/react-dart/pull/220 for more background info

I don't think we ever discovered or looked into when and why the non-native datatransfer object got in there... I talked to Evan and he couldn't remember exactly where he was hitting it:
> evan.weible
> I think it was when I was messing around with drag-and-drop file uploads
> …but I can’t remember much more than that 😕 
> 
> greg.littlefield
> Oooh nice, nice. Do you remember if it was a thing that happened when you were interacting with the app vs hitting it in tests?
>
> evan.weible
> I think when interacting with the app
> but really not sure

Some theories:
- It's from React internals passing in a non-native event somewhere
    - I couldn't find evidence of this; React seems to just pass it through
- It's from test utilities that pass in a JS object stub of a datatransfer object: https://sourcegraph.wk-dev.wdesk.org/search?q=%5B%27%22%5DdataTransfer%5B%27%22%5D+lang:dart++count:1000&patternType=regexp
- The property is being polyfilled in browsers or replaced by an extension

I hate to go down this rabbit hole, but now is probably the time to get to the bottom of this, since we no longer have a layer ([syntheticDataTransferFactory](https://github.com/cleandart/react-dart/pull/284/files#diff-00a14801d8125575e2b4e7ad2eb7b74ae9c57162f484a8d4e7f792416382a9b0L127-L128)) in which we can handle both.

....Oooh, unless we replaced the `dataTransfer` getter with an extension getter that continues to return our `SyntheticDataTransfer` object...

That might be a better solution for now, since I'd imagine chasing it down the root issue is going to be a PITA.

---

Sorry, that was a lot haha. Thoughts?

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -8,8 +8,12 @@ import 'dart:html';
 import 'package:js/js.dart';
 
 @JS()
-@anonymous
 class SyntheticEvent {
```
greglittlefield-wf, line 11: I'm fine doing this in a separate PR, but at some point we should add doc comments on these classes and fields (there are some we can transfer over from the old classes).

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -25,17 +29,27 @@ class SyntheticEvent {
   external void preventDefault();
 
   external void persist();
+
+  external bool isPersistent();
```
greglittlefield-wf, line 33: These should be deprecated

lib/src/react_client/synthetic_event_wrappers.dart
joebingham-wk: Oof that's a fun one haha. I agree it's probably best to use an extension to keep the current functionality. Part of me does really want to dig in and figure out what's happening, and I agree it's a good time.

Part of my logic for procrastinating it a moment more is that with all the event changes, WSD is already in a weird state. I'd want to use WSD to test the fix for `dataTransfer` though, so if that were part of this ticket it'd make getting WSD fixed a dependency to this. Whereas we can do that a little more piecemeal if we do this change, then fix WSD, and then come back to `dataTransfer`. How would you feel about using the extension for now and discussing during a standup or planning to see if it's worth digging into the real cause before the 6.0.0 release? Or would you rather we just do it now if we ever do?

joebingham-wk: @greglittlefield-wf forgot to tag ya haha

lib/src/react_client/event_helpers.dart
```diff
@@ -564,60 +734,271 @@ SyntheticWheelEvent createSyntheticWheelEvent({
...
-  bool get isClipboardEvent => this is SyntheticClipboardEvent;
+  /// Uses Duck Typing to attempt to detect if the event instance is a [SyntheticClipboardEvent].
+  ///
+  /// __NOTE:__ A field unique to this class must be non-null in order for this to return true, even
+  /// if the event instance was instantiated as a [SyntheticClipboardEvent]. For this class, this means
+  /// that one of the following fields must be non-null:
+  /// - clipboardData
+  bool get isClipboardEvent {
+    final typedThis = this as SyntheticClipboardEvent;
+    if (typedThis.clipboardData != null) return true;
```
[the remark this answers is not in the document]
joebingham-wk: That's so much better haha. Thank you!

lib/src/react_client/synthetic_event_wrappers.dart
joebingham-wk: I can add it to this one, that's easy!

lib/src/react_client/synthetic_event_wrappers.dart
greglittlefield-wf: I'd personally be fine with the extension for the 6.0.0 release, since there's not really any benefit to actually fixing the issue as opposed to continuing to use the `SyntheticDataTransfer`

joebingham-wk: But Greg, isn't your inner engineer just dying to know how that's getting mutated? 😉 

That's fair though, it makes sense we don't need to dig into it until we have reason to believe it's causing an issue. So I will make that change and be good to go!

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -4,153 +4,634 @@
...
+
+  /// An [EventTarget] representing a secondary target for this event.
+  ///
+  /// In some cases (such as when tabbing in or out a page), this property may be set to `null` for security reasons.
+  ///
+  /// See <https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget>
   external EventTarget get relatedTarget;
 }
 
+/// A [SyntheticEvent] wrapper that represents a form event.
```
joebingham-wk, line 282: This doesn't exist in React.js and I feel like it should be removed. I started to remove it, but because it's used in several places it felt like that could be its own ticket. Thoughts?

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -4,153 +4,634 @@
...
+  ///
+  /// See <https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/ctrlKey>
   external bool get ctrlKey;
-  external dynamic get dataTransfer;
+
+  /// The data that is transferred during a drag and drop interaction.
+  ///
+  /// See <https://developer.mozilla.org/en-US/docs/Web/API/DragEvent/dataTransfer>
+  @Deprecated('This field can return inconsistent results between dart2js and ddc. Use `wrappedDataTransfer` instead.')
+  external get dataTransfer;
```
joebingham-wk, line 365: This property actually doesn't exist on `SyntheticMouseEvent`s in React.js. Rather, it's attached to a `SyntheticDragEvent`. Drag events look the same as mouse events, just also having this property. So actually, our mouse event here is a drag event. This also feels like something we should fix in react 6.0.0, but wasn't sure if it should be in this ticket or a followup

lib/src/react_client/synthetic_event_wrappers.dart
joebingham-wk: I added a helper in `lib/src/react_client/event_helpers.dart` and deprecated the `dataTransfer` field on the interop class to try to achieve this.. let me know if this wasn't what you were thinking though.

If that is the best path, I could go back into `react.dart` and add the helper there as well so we can still forward fix

lib/src/react_client/event_helpers.dart
```diff
@@ -758,247 +760,49 @@ SyntheticWheelEvent createSyntheticWheelEvent({
...
-    if (typedThis.deltaMode != null) return true;
-    if (typedThis.deltaY != null) return true;
-    if (typedThis.deltaZ != null) return true;
-
-    return false;
-  }
+extension DataTransferHelper on SyntheticMouseEvent {
+  /// Wraps [dataTransfer] in a Dart object, detecting if it is a DOM [DataTransfer] object or a
+  /// JS object that looks like it.
+  SyntheticDataTransfer get wrappedDataTransfer => syntheticDataTransferFactory(dataTransfer);
```
greglittlefield-wf, line 807: Can we make this `dataTransfer` instead of `wrappedDataTransfer` (and remove dataTransfer from the JS class) so that it's backwards-compatible?
```suggestion
  SyntheticDataTransfer get dataTransfer => syntheticDataTransferFactory(getProperty(this, 'dataTransfer'));
```

Otherwise, current usages of `dataTransfer`, expecting it to be a SyntheticDataTransfer, would break.

lib/src/react_client/synthetic_event_wrappers.dart
greglittlefield-wf: I just added a comment on the extension

greglittlefield-wf (commented):
Just a quick pass before I sign off for the day

lib/src/react_client/synthetic_event_wrappers.dart
joebingham-wk: Doope, my interop skills are still lacking and I forget about those methods 🤦 Thank you!!

lib/src/react_client/event_helpers.dart
```diff
@@ -564,60 +736,74 @@ SyntheticWheelEvent createSyntheticWheelEvent({
...
+  /// Uses Duck Typing to detect if the event instance is a [SyntheticFocusEvent].
+  bool get isFocusEvent => hasProperty(this, 'relatedTarget') && !hasProperty(this, 'button');
 
-  /// Returns whether this is a [SyntheticFocusEvent].
-  bool get isFocusEvent => this is SyntheticFocusEvent;
+  /// Uses Duck Typing to detect if the event instance is a [SyntheticFormEvent].
+  ///
+  /// __NOTE:__ This getter is here for completeness, but because the interface for form events
+  /// is the same as that of a [SyntheticEvent] (the base for all other synthetic event types),
+  /// via Duck Typing every [SyntheticEvent] is considered a [SyntheticFormEvent].
```
greglittlefield-wf, line 779: Hmm if this is the case, I feel like we should just omit it so that consumers don't use it without reading the full doc comment and assume it does something other than always return true.

Or, another thought, could we use the set of event types to determine what type of event it is?
From https://reactjs.org/docs/events.html#form-events:
```
onChange onInput onInvalid onReset onSubmit 
```


But, taking a look on SourceGraph, it's only used in one test, and could be replaced with a not-null-check, so I'd lean toward removing it.

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -4,153 +4,621 @@
...
+/// See <https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent>
 @JS()
-@anonymous
 class SyntheticKeyboardEvent extends SyntheticEvent {
+  /// [SyntheticKeyboardEvent]s cannot be manually instantiated.
+  ///
+  /// Use `createSyntheticKeyboardEvent` instead.
+  external factory SyntheticKeyboardEvent._();
+
+  /// Returns `true` if the `Alt` (`Option` or `⌥` on OS X) key was active when the key event was generated.
```
greglittlefield-wf, line 184: #nit Getters should be documented with noun phrases ("returns" should be used only for methods), and prefer using "this" to reference the current instance

- https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases
- https://dart.dev/guides/language/effective-dart/documentation#prefer-using-this-instead-of-the-to-refer-to-a-members-instance

```suggestion
  /// Whether the `Alt` (`Option` or `⌥` on OS X) key was active when this event was generated.
```

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -4,153 +4,621 @@
...
+  ///
+  /// See <https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/altKey>
   external bool get altKey;
+
+  /// Returns a [String] representing the character value of the key.
```
greglittlefield-wf, line 189: #nit Same as the above for the usage of "Returns", and also, specifying/referencing the type aren't necessary because that information is already available via the type annotation.

- https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases
- https://dart.dev/guides/language/effective-dart/documentation#avoid-redundancy-with-the-surrounding-context
- https://dart.dev/guides/language/effective-dart/documentation#prefer-brevity
```suggestion
  /// The character value of the key.
```

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -4,153 +4,621 @@
...
+  /// by a script or dispatched via [Event.dispatchEvent].
+  ///
+  /// __Read Only__
+  ///
+  /// See: <https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted>
   external bool get isTrusted;
-  external get nativeEvent;
-  external get target;
+
+  /// Native browser event that [SyntheticEvent] wraps around.
```
greglittlefield-wf, line 74: #nit Referencing  SyntheticEvent isn't helpful here; instead, reference "this"

- https://dart.dev/guides/language/effective-dart/documentation#avoid-redundancy-with-the-surrounding-context
- https://dart.dev/guides/language/effective-dart/documentation#prefer-using-this-instead-of-the-to-refer-to-a-members-instance
```suggestion
  /// The native browser event this wraps.
```

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -4,153 +4,621 @@
...
+/// See <https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent>
 @JS()
-@anonymous
 class SyntheticTouchEvent extends SyntheticEvent {
+  /// [SyntheticTouchEvent]s cannot be manually instantiated.
+  ///
+  /// Use `createSyntheticTouchEvent` instead.
+  external factory SyntheticTouchEvent._();
+
+  /// A value indicating whether or not the alt key was down when the touch event was fired.
```
greglittlefield-wf, line 472: #nit Same as above, but also "or not" is redundant, and can just be
```suggestion
  /// Whether the alt key was down when this event was fired.
```

lib/src/react_client/event_helpers.dart
```diff
@@ -564,60 +736,74 @@ SyntheticWheelEvent createSyntheticWheelEvent({
...
+  bool get isWheelEvent => hasProperty(this, 'deltaX');
+}
 
-  /// Returns whether this is a [SyntheticWheelEvent].
-  bool get isWheelEvent => this is SyntheticWheelEvent;
+extension DataTransferHelper on SyntheticMouseEvent {
+  /// The data that is transferred during a drag and drop interaction.
+  ///
+  /// See <https://developer.mozilla.org/en-US/docs/Web/API/DragEvent/dataTransfer>
+  SyntheticDataTransfer get dataTransfer => syntheticDataTransferFactory(getProperty(this, 'dataTransfer'));
```
greglittlefield-wf, line 808: Sorry, I was the one that suggested an extension, but I just thought: can we make this a normal getter? Would you mind trying that out?

test/react_client_test.dart
```diff
@@ -138,24 +138,25 @@ main() {
...
 
       test('for a DOM element', () {
         var component = react.div(props);
         var jsProps = unconvertJsProps(component);
-        for (final key in eventPropKeyToEventFactory.keys) {
+        for (final key in knownEventKeys) {
           expect(jsProps[key], isNotNull, reason: 'JS event handler prop should not be null');
-          expect(jsProps[key], same(originalHandlers[key]), reason: 'JS event handler prop was not unconverted');
+          expect(jsProps[key], anyOf(same(originalHandlers[key]), same(allowInterop(originalHandlers[key]))),
+              reason: 'JS event handler prop was not unconverted');
```
greglittlefield-wf, line 152: I think this message is outdated now
```suggestion
              reason: 'JS event handler be the original or original wrapped in allowInterop');
```

test/react_client/event_helpers_test.dart
```diff
@@ -1502,5 +1649,530 @@ main() {
...
 }
+
+// ignore: avoid_implementing_value_types
+class MockKeyboardEvent extends Mock implements KeyboardEvent {}
+
+// ignore: avoid_implementing_value_types
+class MockMouseEvent extends Mock implements MouseEvent {}
+
+enum SyntheticEventType {
+  SyntheticClipboardEvent,
```
greglittlefield-wf, line 2166: #nit enum values should be lowercase (I see why they aren't here, but I think it'd be more readable if they matched the convention)

greglittlefield-wf (commented):
A couple more small comments and then some nits around documentation style that apply to a bunch of the comments. 

I'd be happy to lend a hand in updating those doc comments, so just let me know!

lib/src/react_client/synthetic_event_wrappers.dart
greglittlefield-wf: +1 to its own ticket

greglittlefield-wf:
### Public API Changes

:scream: Error

Semver Audit Service experienced an error during its report.

For details, view the service logs for the correlation ID: [a7c47899-415a-4f75-93a6-65f16a9825ef](https://splunk.workiva.net/en-US/app/search/search?q=search%20index%3D%22kinesis-wk-dev%22%20%20%22service.name%22%3Dsemver-audit%20%22context.correlationId%22%3Da7c47899-415a-4f75-93a6-65f16a9825ef)



Showing results for 8063199ce92cda53f944dd5688c86308fdd2b55a
> Powered by [semver-audit-service](https://github.com/Workiva/semver-audit-service). Please report any problems by [filing an issue](https://github.com/Workiva/semver-audit-service/issues).
> Reported by the dart semver audit client 2.2.0
[Browse public API](https://semver-audit.wk-dev.wdesk.org/cleandart/react-dart/8063199ce92cda53f944dd5688c86308fdd2b55a).

greglittlefield-wf:
### Public API Changes

:scream: Error

Semver Audit Service experienced an error during its report.

For details, view the service logs for the correlation ID: [f5842dea-28ad-4267-8426-a8dde801a2ff](https://splunk.workiva.net/en-US/app/search/search?q=search%20index%3D%22kinesis-wk-dev%22%20%20%22service.name%22%3Dsemver-audit%20%22context.correlationId%22%3Df5842dea-28ad-4267-8426-a8dde801a2ff)



Showing results for 8063199ce92cda53f944dd5688c86308fdd2b55a
> Powered by [semver-audit-service](https://github.com/Workiva/semver-audit-service). Please report any problems by [filing an issue](https://github.com/Workiva/semver-audit-service/issues).
> Reported by the dart semver audit client 2.2.0
[Browse public API](https://semver-audit.wk-dev.wdesk.org/cleandart/react-dart/8063199ce92cda53f944dd5688c86308fdd2b55a).

greglittlefield-wf:
### Public API Changes

:scream: Error

Semver Audit Service experienced an error during its report.

For details, view the service logs for the correlation ID: [d04b976b-e960-46d3-9d1b-9530869c6fad](https://splunk.workiva.net/en-US/app/search/search?q=search%20index%3D%22kinesis-wk-dev%22%20%20%22service.name%22%3Dsemver-audit%20%22context.correlationId%22%3Dd04b976b-e960-46d3-9d1b-9530869c6fad)



Showing results for 8063199ce92cda53f944dd5688c86308fdd2b55a
> Powered by [semver-audit-service](https://github.com/Workiva/semver-audit-service). Please report any problems by [filing an issue](https://github.com/Workiva/semver-audit-service/issues).
> Reported by the dart semver audit client 2.2.0
[Browse public API](https://semver-audit.wk-dev.wdesk.org/cleandart/react-dart/8063199ce92cda53f944dd5688c86308fdd2b55a).

greglittlefield-wf:
### Public API Changes

 Recommendation: **:bangbang: Major version bump (fyi @Workiva/semver-audit-group )**


```diff
@@ line 301: package:react/src/react_client/event_factory.dart @@
-  SyntheticAnimationEvent syntheticAnimationEventFactory(SyntheticAnimationEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 216: package:react/src/react_client/event_factory.dart @@
-  SyntheticMouseEvent syntheticMouseEventFactory(SyntheticMouseEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 17: package:react/src/react_client/event_factory.dart @@
-  SyntheticClipboardEvent syntheticClipboardEventFactory(SyntheticClipboardEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 346: package:react/src/react_client/event_factory.dart @@
-  SyntheticWheelEvent syntheticWheelEventFactory(SyntheticWheelEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 37: package:react/src/react_client/event_factory.dart @@
-  SyntheticCompositionEvent syntheticCompositionEventFactory(SyntheticCompositionEvent e)
// Removing a top-level function is a major change.
```
<details>
<summary>Click to see 28 more API Changes</summary><p>

---
```diff
@@ line 118: package:react/src/react_client/event_factory.dart @@
-  SyntheticDataTransfer syntheticDataTransferFactory(Object dt)
// Removing a top-level function is a major change.
```

```diff
@@ line 107: package:react/src/react_client/event_factory.dart @@
-  SyntheticFormEvent syntheticFormEventFactory(SyntheticFormEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 278: package:react/src/react_client/event_factory.dart @@
-  SyntheticTransitionEvent syntheticTransitionEventFactory(SyntheticTransitionEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 57: package:react/src/react_client/event_factory.dart @@
-  SyntheticKeyboardEvent syntheticKeyboardEventFactory(SyntheticKeyboardEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 254: package:react/src/react_client/event_helpers.dart @@
-  SyntheticMouseEvent createSyntheticMouseEvent({SyntheticMouseEvent baseEvent, bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, void Function() preventDefault, void Function() stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, num button, num buttons, num clientX, num clientY, bool ctrlKey, SyntheticDataTransfer dataTransfer, bool metaKey, num pageX, num pageY, dynamic relatedTarget, num screenX, num screenY, bool shiftKey})
+  SyntheticMouseEvent createSyntheticMouseEvent({SyntheticMouseEvent baseEvent, bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, void Function() preventDefault, void Function() stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, num button, num buttons, num clientX, num clientY, bool ctrlKey, dynamic dataTransfer, bool metaKey, num pageX, num pageY, dynamic relatedTarget, num screenX, num screenY, bool shiftKey})
// `type` of `dataTransfer` has changed.
// Changing a parameter signature is a major change.
```

```diff
@@ line 9: package:react/src/react_client/event_factory.dart @@
-  SyntheticEvent syntheticEventFactory(SyntheticEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 324: package:react/src/react_client/event_factory.dart @@
-  SyntheticUIEvent syntheticUIEventFactory(SyntheticUIEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 87: package:react/src/react_client/event_factory.dart @@
-  SyntheticFocusEvent syntheticFocusEventFactory(SyntheticFocusEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 186: package:react/src/react_client/event_factory.dart @@
-  SyntheticPointerEvent syntheticPointerEventFactory(SyntheticPointerEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 251: package:react/src/react_client/event_factory.dart @@
-  SyntheticTouchEvent syntheticTouchEventFactory(SyntheticTouchEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 1884: package:react/react.dart @@
class SyntheticTransitionEvent extends SyntheticEvent
-     SyntheticTransitionEvent SyntheticTransitionEvent(dynamic bubbles, dynamic cancelable, dynamic currentTarget, dynamic _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, dynamic eventPhase, dynamic isTrusted, dynamic nativeEvent, dynamic target, dynamic timeStamp, dynamic type, String propertyName, num elapsedTime, String pseudoElement)
//    Removing a constructor is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

```

```diff
@@ line 1756: package:react/react.dart @@
class SyntheticMouseEvent extends SyntheticEvent
-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     dynamic get relatedTarget
+     EventTarget get relatedTarget
//    `type` of `relatedTarget` has changed from `dynamic` to `EventTarget`.
//    Changing a class field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     SyntheticMouseEvent SyntheticMouseEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, num button, num buttons, num clientX, num clientY, bool ctrlKey, SyntheticDataTransfer dataTransfer, bool metaKey, num pageX, num pageY, dynamic relatedTarget, num screenX, num screenY, bool shiftKey)
//    Removing a constructor is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticDataTransfer get dataTransfer
//    Removing a field is a major change.

```

```diff
@@ line 1447: package:react/react.dart @@
class SyntheticEvent
-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticEvent SyntheticEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1806: package:react/react.dart @@
class SyntheticPointerEvent extends SyntheticEvent
-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticPointerEvent SyntheticPointerEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, num pointerId, num width, num height, num pressure, num tangentialPressure, num tiltX, num tiltY, num twist, String pointerType, bool isPrimary)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

```

```diff
@@ line 1912: package:react/react.dart @@
class SyntheticAnimationEvent extends SyntheticEvent
-     SyntheticAnimationEvent SyntheticAnimationEvent(dynamic bubbles, dynamic cancelable, dynamic currentTarget, dynamic _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, dynamic eventPhase, dynamic isTrusted, dynamic nativeEvent, dynamic target, dynamic timeStamp, dynamic type, String animationName, num elapsedTime, String pseudoElement)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

```

```diff
@@ line 1610: package:react/react.dart @@
class SyntheticClipboardEvent extends SyntheticEvent
-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     SyntheticClipboardEvent SyntheticClipboardEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, dynamic clipboardData)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1634: package:react/react.dart @@
class SyntheticKeyboardEvent extends SyntheticEvent
-     SyntheticKeyboardEvent SyntheticKeyboardEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, String char, num charCode, bool ctrlKey, String locale, num location, String key, num keyCode, bool metaKey, bool repeat, bool shiftKey)
//    Removing a constructor is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1940: package:react/react.dart @@
class SyntheticUIEvent extends SyntheticEvent
-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticUIEvent SyntheticUIEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, num detail, dynamic view)
//    Removing a constructor is a major change.

```

```diff
@@ line 1678: package:react/react.dart @@
class SyntheticCompositionEvent extends SyntheticEvent
-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticCompositionEvent SyntheticCompositionEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, String data)
//    Removing a constructor is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1848: package:react/react.dart @@
class SyntheticTouchEvent extends SyntheticEvent
-     dynamic get touches
+     TouchList get touches
//    `type` of `touches` has changed from `dynamic` to `TouchList`.
//    Changing a class field is a major change.

-     dynamic get targetTouches
+     TouchList get targetTouches
//    `type` of `targetTouches` has changed from `dynamic` to `TouchList`.
//    Changing a class field is a major change.

-     SyntheticTouchEvent SyntheticTouchEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, dynamic changedTouches, bool ctrlKey, bool metaKey, bool shiftKey, dynamic targetTouches, dynamic touches)
//    Removing a constructor is a major change.

-     dynamic get changedTouches
+     TouchList get changedTouches
//    `type` of `changedTouches` has changed from `dynamic` to `TouchList`.
//    Changing a class field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 1726: package:react/react.dart @@
class SyntheticFormEvent extends SyntheticEvent
-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     SyntheticFormEvent SyntheticFormEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 1702: package:react/react.dart @@
class SyntheticFocusEvent extends SyntheticEvent
-     dynamic get relatedTarget
+     EventTarget get relatedTarget
//    `type` of `relatedTarget` has changed from `dynamic` to `EventTarget`.
//    Changing a class field is a major change.

-     SyntheticFocusEvent SyntheticFocusEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, dynamic relatedTarget)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 1966: package:react/react.dart @@
class SyntheticWheelEvent extends SyntheticEvent
-     SyntheticWheelEvent SyntheticWheelEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, num deltaX, num deltaMode, num deltaY, num deltaZ)
//    Removing a constructor is a major change.

-     bool get isPersistent
+     bool isPersistent()
//    Removed getter for `isPersistent`.
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 91: package:react/src/react_client/event_helpers.dart @@
+  SyntheticFormEvent fakeSyntheticFormEvent(Element element, String type)
// Adding a top-level function is a minor change.
```

```diff
@@ line 50: package:react/src/react_client/event_helpers.dart @@
+  SyntheticMouseEvent wrapNativeMouseEvent(MouseEvent nativeEvent)
// Adding a top-level function is a minor change.
```

```diff
@@ line 14: package:react/src/react_client/event_helpers.dart @@
+  SyntheticKeyboardEvent wrapNativeKeyboardEvent(KeyboardEvent nativeEvent)
// Adding a top-level function is a minor change.
```

```diff
@@ line 804: package:react/src/react_client/event_helpers.dart @@
+  extension DataTransferHelper on SyntheticMouseEvent
// Adding a class is a minor change.


```

```diff
@@ line 292: package:react/src/react_client/synthetic_event_wrappers.dart @@
+  class NonNativeDataTransfer
// Adding a class is a minor change.


```
---

</p></details>

Showing results for 8063199ce92cda53f944dd5688c86308fdd2b55a
> Powered by [semver-audit-service](https://github.com/Workiva/semver-audit-service). Please report any problems by [filing an issue](https://github.com/Workiva/semver-audit-service/issues).
> Reported by the dart semver audit client 2.2.0
[Browse public API](https://semver-audit.wk-dev.wdesk.org/cleandart/react-dart/8063199ce92cda53f944dd5688c86308fdd2b55a).

greglittlefield-wf:
Some feedback from the semver report:

- `NonNativeDataTransfer` is publicly exported; I think we should keep that private for now since it's an implementation detail and there's not currently a use-case for exposing it.

- `isPersistant`
    > ```diff
    > -     bool get isPersistent
    > +     bool isPersistent()
    > ```
    Unfortunately the old API differed from the React version, so we'll want to keep this as a getter (we should be able to just do one that always returns true, either via a non-external getter or an extension).

- Changing `relatedTarget` from `dynamic` to `EventTarget`: while this is the correct typing, changing it would result in breakages in consumer code where they're currently performing dynamic calls to methods that exist on EventTarget subclasses (which [seems to be pretty common](https://sourcegraph.wk-dev.wdesk.org/search?q=lang%3Adart+%5C.relatedTarget+count%3A1000&patternType=regexp)). For example:
    ```dart
    if (event.relatedTarget is Element) {
      event.relatedTarget.focus();
    }
    ```
    This currently only works because the call to `.focus()` is on a dynamic object; `event.relatedTarget` doesn't get type-promoted inside the if-check, so after these changes you'd get an error: "The method 'focus' isn't defined for the type 'EventTarget'"

    While we should probably update the typing eventually, I think we should keep it as `dynamic` for now to make the upgrade easier.

- TouchEvent updates: I'm a little worried about the `TouchList` types being incompatible with simulated events that pass in `List`s (e.g., [this case](https://sourcegraph.wk-dev.wdesk.org/github.com/Workiva/wdesk_sdk/-/blob/test/unit/browser/segregated_tests/memory_leakers/truss_1/rich_app_shell/components/side_panel_test.dart#L334:14)) because `TouchList` can't easily be instantiated in Dart; what about `List<Touch>` instead?

lib/src/react_client/event_helpers.dart
joebingham-wk: I can remove it! 

With removing it, we'd want to go back and deprecate it in the stable line, right?

lib/src/react_client/event_helpers.dart
joebingham-wk: Yep! I gave it a go and it behaves weirdly. Even though I define the getter on the interop class, which I'd expect would override the `external`, calling `event.dataTransfer` appears to access the external getter anyway. It only happens when the getter is on the class itself and not an extension though. I stepped through in the debugger and when the getter is in the interop class, it acts like a simple getter and there isn't anything to step into (it just grabs the property off the JS object). As an extension though, I can step through to the `syntheticDataTransferFactory`. 

So that's weird. Let me know if you want to step through it together and we definitely can!

lib/src/react_client/synthetic_data_transfer.dart
```diff
@@ -104,9 +104,8 @@ SyntheticDataTransfer syntheticDataTransferFactory(Object dt) {
   }
 
   // Copy these lists and ensure they're typed properly.
-  // todo use .cast() in Dart 2
```
joebingham-wk: I gave addressing this a go while I was here 🤷‍♂️  It seemed simple, but I don't fully grasp the why behind the change so let me know if it's actually more complicated or incorrect

lib/src/react_client/event_helpers.dart
joebingham-wk: Removed in 067e282!

lib/src/react_client/synthetic_event_wrappers.dart
joebingham-wk: :noice: I had originally just kept the original MDN web API doc comments as it seemed to make sense to just keep it matching with what was already out there, but those tweaks were easy. Addressed in 2bc5ffa (plus a couple in 215d9ec).

lib/src/react_client/synthetic_event_wrappers.dart
```diff
@@ -148,7 +148,7 @@ class SyntheticEvent {
   ///
   /// > See: [persist]
   @Deprecated('The modern event system does not use pooling. This always returns true.')
-  external bool isPersistent();
+  bool get isPersistent => true;
```
joebingham-wk, line 151: Hmmm related to this change, it looks like this needs to be external: https://wf-skynet-hrd.appspot.com/apps/test/smithy/2552158/1/graph_ui/unit-tests#summary

Trying it as an extension instead :loading:

lib/src/react_client/event_helpers.dart
joebingham-wk: [This](https://github.com/cleandart/react-dart/pull/284#discussion_r526340434) might be a related issue as well

lib/src/react_client/synthetic_event_wrappers.dart
joebingham-wk: Yep, the extension worked: https://wf-skynet-hrd.appspot.com/apps/test/smithy/2552158/3

lib/src/react_client/event_helpers.dart
greglittlefield-wf: > With removing it, we'd want to go back and deprecate it in the stable line, right?

Ooh, nice catch, I didn't think about that. Yup!

lib/src/react_client/event_helpers.dart
greglittlefield-wf: Makes sense, sounds like this is just one of the unwritten rules of JS interop. Thanks for looking into it!!

lib/src/react_client/synthetic_data_transfer.dart
greglittlefield-wf: The reason this code is in place is that we have lists `rawFiles` and `rawTypes` which are `List<dynamic>` (or null), but they need to be converted to `List<File>` and `List<String>`.

The idea of the comment is that in Dart 2 we could start using `cast` like this:
```dart
final files = rawFiles.cast<File>() ?? <File>[];
// or this
final files = (rawFiles ?? []).cast<File>();
// or this lol
final List<File> files = rawFiles?.cast() ?? [];
```

If we had implicit casts disabled in this repo, then the type parameters can be elided if you inline the expression in the constructor argument:

```dart
return new SyntheticDataTransfer(dropEffect, effectAllowed, rawFiles?.cast() ?? [], types?.cast() ?? []);
```
but we don't have em disabled so nevermind haha.

Another option, not using `cast`, would be this:
```dart
final files = <File>[...?rawFiles];
final types = <String>[...?rawTypes];
```
I think that one is my favorite :)

greglittlefield-wf:
### Public API Changes

 Recommendation: **:bangbang: Major version bump (fyi @Workiva/semver-audit-group )**


```diff
@@ line 301: package:react/src/react_client/event_factory.dart @@
-  SyntheticAnimationEvent syntheticAnimationEventFactory(SyntheticAnimationEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 216: package:react/src/react_client/event_factory.dart @@
-  SyntheticMouseEvent syntheticMouseEventFactory(SyntheticMouseEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 17: package:react/src/react_client/event_factory.dart @@
-  SyntheticClipboardEvent syntheticClipboardEventFactory(SyntheticClipboardEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 346: package:react/src/react_client/event_factory.dart @@
-  SyntheticWheelEvent syntheticWheelEventFactory(SyntheticWheelEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 37: package:react/src/react_client/event_factory.dart @@
-  SyntheticCompositionEvent syntheticCompositionEventFactory(SyntheticCompositionEvent e)
// Removing a top-level function is a major change.
```
<details>
<summary>Click to see 28 more API Changes</summary><p>

---
```diff
@@ line 118: package:react/src/react_client/event_factory.dart @@
-  SyntheticDataTransfer syntheticDataTransferFactory(Object dt)
// Removing a top-level function is a major change.
```

```diff
@@ line 107: package:react/src/react_client/event_factory.dart @@
-  SyntheticFormEvent syntheticFormEventFactory(SyntheticFormEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 278: package:react/src/react_client/event_factory.dart @@
-  SyntheticTransitionEvent syntheticTransitionEventFactory(SyntheticTransitionEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 57: package:react/src/react_client/event_factory.dart @@
-  SyntheticKeyboardEvent syntheticKeyboardEventFactory(SyntheticKeyboardEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 254: package:react/src/react_client/event_helpers.dart @@
-  SyntheticMouseEvent createSyntheticMouseEvent({SyntheticMouseEvent baseEvent, bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, void Function() preventDefault, void Function() stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, num button, num buttons, num clientX, num clientY, bool ctrlKey, SyntheticDataTransfer dataTransfer, bool metaKey, num pageX, num pageY, dynamic relatedTarget, num screenX, num screenY, bool shiftKey})
+  SyntheticMouseEvent createSyntheticMouseEvent({SyntheticMouseEvent baseEvent, bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, void Function() preventDefault, void Function() stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, num button, num buttons, num clientX, num clientY, bool ctrlKey, dynamic dataTransfer, bool metaKey, num pageX, num pageY, dynamic relatedTarget, num screenX, num screenY, bool shiftKey})
// `type` of `dataTransfer` has changed.
// Changing a parameter signature is a major change.
```

```diff
@@ line 9: package:react/src/react_client/event_factory.dart @@
-  SyntheticEvent syntheticEventFactory(SyntheticEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 324: package:react/src/react_client/event_factory.dart @@
-  SyntheticUIEvent syntheticUIEventFactory(SyntheticUIEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 87: package:react/src/react_client/event_factory.dart @@
-  SyntheticFocusEvent syntheticFocusEventFactory(SyntheticFocusEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 186: package:react/src/react_client/event_factory.dart @@
-  SyntheticPointerEvent syntheticPointerEventFactory(SyntheticPointerEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 251: package:react/src/react_client/event_factory.dart @@
-  SyntheticTouchEvent syntheticTouchEventFactory(SyntheticTouchEvent e)
// Removing a top-level function is a major change.
```

```diff
@@ line 1884: package:react/react.dart @@
class SyntheticTransitionEvent extends SyntheticEvent
-     SyntheticTransitionEvent SyntheticTransitionEvent(dynamic bubbles, dynamic cancelable, dynamic currentTarget, dynamic _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, dynamic eventPhase, dynamic isTrusted, dynamic nativeEvent, dynamic target, dynamic timeStamp, dynamic type, String propertyName, num elapsedTime, String pseudoElement)
//    Removing a constructor is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

```

```diff
@@ line 1756: package:react/react.dart @@
class SyntheticMouseEvent extends SyntheticEvent
-     bool get isPersistent
//    Removing a field is a major change.

-     dynamic get relatedTarget
+     EventTarget get relatedTarget
//    `type` of `relatedTarget` has changed from `dynamic` to `EventTarget`.
//    Changing a class field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     SyntheticMouseEvent SyntheticMouseEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, num button, num buttons, num clientX, num clientY, bool ctrlKey, SyntheticDataTransfer dataTransfer, bool metaKey, num pageX, num pageY, dynamic relatedTarget, num screenX, num screenY, bool shiftKey)
//    Removing a constructor is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticDataTransfer get dataTransfer
//    Removing a field is a major change.

```

```diff
@@ line 1447: package:react/react.dart @@
class SyntheticEvent
-     bool get isPersistent
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticEvent SyntheticEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1806: package:react/react.dart @@
class SyntheticPointerEvent extends SyntheticEvent
-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticPointerEvent SyntheticPointerEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, num pointerId, num width, num height, num pressure, num tangentialPressure, num tiltX, num tiltY, num twist, String pointerType, bool isPrimary)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

```

```diff
@@ line 1912: package:react/react.dart @@
class SyntheticAnimationEvent extends SyntheticEvent
-     SyntheticAnimationEvent SyntheticAnimationEvent(dynamic bubbles, dynamic cancelable, dynamic currentTarget, dynamic _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, dynamic eventPhase, dynamic isTrusted, dynamic nativeEvent, dynamic target, dynamic timeStamp, dynamic type, String animationName, num elapsedTime, String pseudoElement)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

```

```diff
@@ line 1610: package:react/react.dart @@
class SyntheticClipboardEvent extends SyntheticEvent
-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

-     SyntheticClipboardEvent SyntheticClipboardEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, dynamic clipboardData)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1634: package:react/react.dart @@
class SyntheticKeyboardEvent extends SyntheticEvent
-     SyntheticKeyboardEvent SyntheticKeyboardEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, String char, num charCode, bool ctrlKey, String locale, num location, String key, num keyCode, bool metaKey, bool repeat, bool shiftKey)
//    Removing a constructor is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1940: package:react/react.dart @@
class SyntheticUIEvent extends SyntheticEvent
-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticUIEvent SyntheticUIEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool _defaultPrevented, dynamic _preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, num detail, dynamic view)
//    Removing a constructor is a major change.

```

```diff
@@ line 1678: package:react/react.dart @@
class SyntheticCompositionEvent extends SyntheticEvent
-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

-     SyntheticCompositionEvent SyntheticCompositionEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, String data)
//    Removing a constructor is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

```

```diff
@@ line 1848: package:react/react.dart @@
class SyntheticTouchEvent extends SyntheticEvent
-     dynamic get touches
+     List<Touch> get touches
//    `type` of `touches` has changed from `dynamic` to `List<Touch>`.
//    Changing a class field is a major change.

-     dynamic get targetTouches
+     List<Touch> get targetTouches
//    `type` of `targetTouches` has changed from `dynamic` to `List<Touch>`.
//    Changing a class field is a major change.

-     SyntheticTouchEvent SyntheticTouchEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, bool altKey, dynamic changedTouches, bool ctrlKey, bool metaKey, bool shiftKey, dynamic targetTouches, dynamic touches)
//    Removing a constructor is a major change.

-     dynamic get changedTouches
+     List<Touch> get changedTouches
//    `type` of `changedTouches` has changed from `dynamic` to `List<Touch>`.
//    Changing a class field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 1726: package:react/react.dart @@
class SyntheticFormEvent extends SyntheticEvent
-     bool get isPersistent
//    Removing a field is a major change.

-     SyntheticFormEvent SyntheticFormEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 1702: package:react/react.dart @@
class SyntheticFocusEvent extends SyntheticEvent
-     SyntheticFocusEvent SyntheticFocusEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, dynamic relatedTarget)
//    Removing a constructor is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 587: package:react/src/react_client/event_helpers.dart @@
extension SyntheticEventTypeHelpers on SyntheticEvent
-     bool get isFormEvent
//    Removing a field is a major change.

+     bool get isPersistent
//    Adding a field is a minor change.

```

```diff
@@ line 1966: package:react/react.dart @@
class SyntheticWheelEvent extends SyntheticEvent
-     SyntheticWheelEvent SyntheticWheelEvent(bool bubbles, bool cancelable, dynamic currentTarget, bool defaultPrevented, dynamic preventDefault, dynamic stopPropagation, num eventPhase, bool isTrusted, dynamic nativeEvent, dynamic target, num timeStamp, String type, num deltaX, num deltaMode, num deltaY, num deltaZ)
//    Removing a constructor is a major change.

-     bool get isPersistent
//    Removing a field is a major change.

-     void Function() $$jsPersistDoNotSetThisOrYouWillBeFired
//    Removing a field is a major change.

-     dynamic get stopPropagation
+     void stopPropagation()
//    Removed getter for `stopPropagation`.
//    Removing a field is a major change.

```

```diff
@@ line 91: package:react/src/react_client/event_helpers.dart @@
+  SyntheticFormEvent fakeSyntheticFormEvent(Element element, String type)
// Adding a top-level function is a minor change.
```

```diff
@@ line 50: package:react/src/react_client/event_helpers.dart @@
+  SyntheticMouseEvent wrapNativeMouseEvent(MouseEvent nativeEvent)
// Adding a top-level function is a minor change.
```

```diff
@@ line 14: package:react/src/react_client/event_helpers.dart @@
+  SyntheticKeyboardEvent wrapNativeKeyboardEvent(KeyboardEvent nativeEvent)
// Adding a top-level function is a minor change.
```

```diff
@@ line 803: package:react/src/react_client/event_helpers.dart @@
+  extension DataTransferHelper on SyntheticMouseEvent
// Adding a class is a minor change.


```
---

</p></details>

Showing results for 9aa536f65a57faefdf8996dc77c9bbf41f905b25
> Powered by [semver-audit-service](https://github.com/Workiva/semver-audit-service). Please report any problems by [filing an issue](https://github.com/Workiva/semver-audit-service/issues).
> Reported by the dart semver audit client 2.2.0
[Browse public API](https://semver-audit.wk-dev.wdesk.org/cleandart/react-dart/9aa536f65a57faefdf8996dc77c9bbf41f905b25).

greglittlefield-wf (approved):
+10

We just need to make sure to re-export these new extensions from over_react so that people don't have to import react-dart: https://github.com/Workiva/graph_ui/pull/1811/files#diff-24bb7e81736e0c04f430abab63b23e46d779cefd80f10a76d1ee7be3d3b7a5a5R17

lib/src/react_client/synthetic_data_transfer.dart
joebingham-wk: Oof, that's clean, thank you for the context and examples! I updated that in 7d1bc70!

greglittlefield-wf (approved):
+10

PR #287, opened 2020-11-23
Deprecate Helper Method

joebingham-wk:
A helper method was added in stable that we realized wasn't necessary in the 6.0.0 line. This PR just deprecates that method!


greglittlefield-wf (approved):
+10

PR #288, opened 2020-11-24
Update SyntheticEvent duck typing

sydneyjodon-wk:
Type checking `SyntheticEvents` returns `false` incorrectly when simulated events do not have expected properties.

Changes:
- Update test utils to test event types for simulated events.
- Update event duck type checks to also check `event.type`.


lib/src/react_client/event_helpers.dart
```diff
@@ -760,44 +760,47 @@ SyntheticWheelEvent createSyntheticWheelEvent({
...
+
   /// Whether the event instance has been removed from the ReactJS event pool.
   ///
   /// > See: [persist]
   @Deprecated('The modern event system does not use pooling. This always returns true.')
   bool get isPersistent => true;
 
   /// Uses Duck Typing to detect if the event instance is a [SyntheticClipboardEvent].
-  bool get isClipboardEvent => hasProperty(this, 'clipboardData');
+  bool get isClipboardEvent => hasProperty(this, 'clipboardData') || _checkEventType(['copy', 'paste', 'cut']);
```
greglittlefield-wf, line 772: These all should be `const`s
```suggestion
  bool get isClipboardEvent => hasProperty(this, 'clipboardData') || _checkEventType(const ['copy', 'paste', 'cut']);
```

lib/src/react_client/event_helpers.dart
```diff
@@ -760,44 +760,47 @@ SyntheticWheelEvent createSyntheticWheelEvent({
 }
 
 extension SyntheticEventTypeHelpers on SyntheticEvent {
+  bool _checkEventType(List<String> types) => hasProperty(this, 'type') && types.any((t) => this.type.contains(t));
```
greglittlefield-wf, line 763: #nit it's unlikely, but `this.type` could still technically be `null` here, so perhaps we could combine the `hasProperty` check with a null-check?
```suggestion
  bool _checkEventType(List<String> types) => type != null && types.any((t) => this.type.contains(t));
```

test/react_test_utils_test.dart
```diff
@@ -91,58 +93,88 @@ testUtils({isComponent2: false, dynamic eventComponent, dynamic sampleComponent,
...
+      }
     }
 
     group('event', () {
-      group('animationEnd', () => testEvent(Simulate.animationEnd, 'animationEnd'));
-      group('animationIteration', () => testEvent(Simulate.animationIteration, 'animationIteration'));
-      group('animationStart', () => testEvent(Simulate.animationStart, 'animationStart'));
-      group('blur', () => testEvent(Simulate.blur, 'blur'));
+      group('animationEnd',
+          () => testEvent(Simulate.animationEnd, 'animationEnd', (e) => expect(e.isAnimationEvent, isTrue)));
```
greglittlefield-wf, line 111: I just realized that these all test the `true`-cases for events, but not the `false` cases.

Could we update these tests to do something similar to event_helpers_test's `commonFalseTests`?

greglittlefield-wf (changes requested):

test/react_test_utils_test.dart
sydneyjodon-wk: Oh good catch! I'll update those!

test/react_test_utils_test.dart
```diff
@@ -91,58 +96,101 @@ testUtils({isComponent2: false, dynamic eventComponent, dynamic sampleComponent,
...
-      group('submit', () => testEvent(Simulate.submit, 'submit'));
-      group('touchCancel', () => testEvent(Simulate.touchCancel, 'touchCancel'));
-      group('touchEnd', () => testEvent(Simulate.touchEnd, 'touchEnd'));
-      group('touchMove', () => testEvent(Simulate.touchMove, 'touchMove'));
-      group('touchStart', () => testEvent(Simulate.touchStart, 'touchStart'));
-      group('transitionEnd', () => testEvent(Simulate.transitionEnd, 'transitionEnd'));
-      group('wheel', () => testEvent(Simulate.wheel, 'wheel'));
+      void Function(SyntheticEvent) _expectEventType<T>() {
+        return (SyntheticEvent event) {
+          expect(event.isClipboardEvent, T == SyntheticClipboardEvent);
```
greglittlefield-wf, line 115: I like this solution, but I'm not sure that type-equality between reified type parameters is guaranteed by the language, so I'm hesitant to rely on it without asserting that exactly one of these conditions is `true`.

For instance, all of them could be false if the type isn't exactly equal, or they could all be true if JS interop types end up being the same in dart2js or something weird and they're equal to each other.

Instead of generic parameters, could we use enums from `event_helpers_test.dart` instead?

greglittlefield-wf (changes requested):

test/react_test_utils_test.dart
sydneyjodon-wk: Updated in 03454a1

sydneyjodon-wk:
This [ticket](https://jira.atl.workiva.net/browse/CPLAT-12602) was created to address remaining CI failures.

greglittlefield-wf (approved):
+10

aaronlademann-wf (approved):
+1

PR #290, opened 2020-11-25
Format on Dart 2.10.4, only check formatting on stable runs

greglittlefield-wf:
## Motivation
CI was failing on every commit because the code was formatted under Dart 2.9, and CI was only skipping formatting on `dev` runs and not the Dart 2.10 `stable` run.

Previously, stable runs were broken due to build issues, but now that that's fixed, we can format on stable.

## Solution
- Format on Dart 2.10.4 (stable)
- Only check formatting on stable runs

## QA steps
- Verify all Travis builds are green

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: dart
 
 dart:
+  # We need 2.9.2 this to verify the Chrome MemoryInfo workaround: https://github.com/cleandart/react-dart/pull/280
+  # TODO remove the workaround as well as this config once SDK lower bound is >=2.9.3
   - 2.9.2
   - stable
   - dev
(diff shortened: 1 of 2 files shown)


test/lifecycle_test/util.dart
```diff
@@ -40,7 +40,11 @@ mixin LifecycleTestHelper on Component {
       'context': context,
     });
 
-    var lifecycleCallback = props == null ? staticProps == null ? null : staticProps[memberName] : props[memberName];
+    var lifecycleCallback = props == null
+        ? staticProps == null
+            ? null
+            : staticProps[memberName]
+        : props[memberName];
```
greglittlefield-wf, line 47: This is just a formatting change

.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: dart
 
 dart:
+  # We need 2.9.2 this to verify the Chrome MemoryInfo workaround: https://github.com/cleandart/react-dart/pull/280
```
joebingham-wk, line 4: #nit typo
```suggestion
  # We need 2.9.2 to verify the Chrome MemoryInfo workaround: https://github.com/cleandart/react-dart/pull/280
```

joebingham-wk (approved):
+10 - change looks good and CI is all green 🎉 🎉 🎉

joebingham-wk (approved):
+10 refresh

PR #294, opened 2020-12-18
Add null handling to synthetic event duck type checking

aaronlademann-wf:
### Problem
When [we updated the synthetic event system to use duck typing instead of Dart `is` type checking](https://github.com/cleandart/react-dart/pull/288), we created a possible null exception that presents as seen below when the event is null: 

```
Error: Cannot use 'in' operator to search for 'key' in null
```

This is a regression since before we made the switch, a check like this would have simply returned false:

```dart
final someSyntheticEvent = null;
final isSyntheticMouseEvent = someSyntheticEvent is SyntheticMouseEvent; // false
```

but now, it throws the exception when the logic is updated to:

```dart
final someSyntheticEvent = null;
final isSyntheticMouseEvent = someSyntheticEvent.isMouseEvent; // throws RTE
```

### Solution

1. Add null checks before calling `hasProperty` and `_checkEventType`.
2. Add tests

### DIFF
--- a/lib/src/react_client/event_helpers.dart
+++ b/lib/src/react_client/event_helpers.dart
@@ -760,7 +760,8 @@ SyntheticWheelEvent createSyntheticWheelEvent({
 }
 
 extension SyntheticEventTypeHelpers on SyntheticEvent {
-  bool _checkEventType(List<String> types) => type != null && types.any((t) => type.contains(t));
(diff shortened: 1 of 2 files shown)


sydneyjodon-wk (approved):
+1

joebingham-wk (approved):
+10

Code looks great! I verified the added test fails (with a very gross, cryptic error) without the added changes

greglittlefield-wf (approved):

PR #350, opened 2022-12-05
FED-202 Return `jsUndefined` instead of `null` when children prop is empty

sydneyjodon-wk:
Some JS components, like `Tab` expect no children to be passed to them.

However, when rendered from Dart, Tab's propTypes warn about children not being supported, even when no children are provided:
```
react.js:6902 Warning: Failed prop type: The prop `children` is not supported. Please remove it.
    at Tab (http://localhost:8080/packages/react_material_ui/react-material-ui-development.umd.js:46607:19)
```

This is because, when there are no children, `ReactJsComponentFactoryProxy` passes `null` into createElement, and the `unsupportedProp` validation checks warns when the prop is undefined.

## Changes

- Return `jsUndefined` instead of `null` from `generateChildren` when the children prop is empty
- Add regression test

## QA Instructions

- [ ] Verify CI passes
- [ ] Verify good test coverage

### DIFF
--- a/test/factory/js_factory_test.js
+++ b/test/factory/js_factory_test.js
@@ -8,12 +8,4 @@ window._JsFooFunction = React.forwardRef((props, ref) => (
   React.createElement("div", {...props, ref: ref})
 ));
 
-window._JsNoChildren = class JsNoChildrenComponent extends React.Component {
-  render() {
-    if(this.props.children !== undefined) {
-      throw Error('children prop must be undefined');
-    }
-
-    return React.createElement("div", this.props);
-  }
-};
+window.hasUndefinedChildren = (reactElement) => reactElement.props.children === undefined;
--- a/test/factory/js_factory_test.js
+++ b/test/factory/js_factory_test.js
@@ -7,3 +7,13 @@ window._JsFoo = class JsFooComponent extends React.Component {
 window._JsFooFunction = React.forwardRef((props, ref) => (
   React.createElement("div", {...props, ref: ref})
 ));
+
+window._JsNoChildren = class JsNoChildrenComponent extends React.Component {
+  render() {
+    if(this.props.children !== undefined) {
+      throw Error('children prop must be undefined');
+    }
+
+    return React.createElement("div", this.props);
+  }
+};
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Security Insights

## Action Items
---

test/factory/js_factory_test.js
```diff
@@ -7,3 +7,13 @@ window._JsFoo = class JsFooComponent extends React.Component {
 window._JsFooFunction = React.forwardRef((props, ref) => (
   React.createElement("div", {...props, ref: ref})
 ));
+
+window._JsNoChildren = class JsNoChildrenComponent extends React.Component {
+  render() {
+    if(this.props.children !== undefined) {
+      throw Error('children prop must be undefined');
+    }
```
sydneyjodon-wk, line 15: I don't know if there's a more straightforward way to test that `props.children` is specifically `undefined` and not `null`

aaronlademann-wf (approved):
+1

test/factory/js_factory_test.js
greglittlefield-wf: I thought maybe we could change [this expectation](https://github.com/Workiva/react-dart/blob/master/test/factory/common_factory_tests.dart#L55):
```diff
-expect(getChildren(instance), shouldAlwaysBeList ? [] : isNull);
+// We want to make sure this is undefined and not null, so we have to use `same`
+// which uses `identical`, since in Dart null and undefined are otherwise interchangeable.
+expect(getChildren(instance), shouldAlwaysBeList ? [] : same(jsUndefined));
```

But it looks like that doesn't even work, since the `identical` check gets compiled to `==` instead of `===` , and `identical(jsNull, jsUndefined)` is `true` 😕.

So, checking it from the JS seems like the way to go to me!

Related to my comment above, we could also potentially check whether it's undefined on the ReactElement without rendering it, if the rendering/throwing causes issues or complicates things.
```tsx
window.hasUndefinedChildren = (reactElement) => reactElement.props.children === undefined;
```
```dart
 expect(hasUndefinedChildren(JsNoChildren({})), isTrue);
```

sydneyjodon-wk: Oh that's a lot nicer! Thank you!

rmconsole-wf (approved):
+1 from RM

PR #357, opened 2023-04-19
FED-1254 add Suspense component

kealjones-wk:
Some components are super "lazy", let's add a way to help them out.


aviary3-wk:
## Security Insights

## Action Items
---

lib/react_client/component_factory.dart
```diff
@@ -381,10 +380,10 @@ ReactClass _wrapForwardRefFunctionComponent(DartForwardRefFunctionComponent dart
   // and throws if it gets `undefined`. `jsNull` is an interop variable that holds a JS `null` value
   // to force `null` as the return value if user returns a Dart `null`.
   // See: https://github.com/dart-lang/sdk/issues/27485
-  jsFunctionComponent(JsMap props, dynamic ref) =>
-      componentZone.run(() => dartFunctionComponent(JsBackedMap.backedBy(props), ref) ?? jsNull);
 
-  final interopFunction = allowInterop(jsFunctionComponent);
+  final interopFunction = allowInterop((JsMap props, dynamic ref) =>
+      componentZone.run(() => dartFunctionComponent(JsBackedMap.backedBy(props), ref) ?? jsNull));
+
```
kealjones-wk, line 386: what the heck? i didn't change these...

example/lazy/index.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
```
joebingham-wk, line 1: This isn't linked up to the root `index.html`, so you have to manually navigate to it! I think we should add a link since the other stuff is all linked from there

lib/react.dart
```diff
@@ -65,6 +65,11 @@ typedef ReactDartFunctionComponentFactoryProxy FunctionComponentRegistrar(DartFu
 /// See: <https://reactjs.org/docs/fragments.html>
 var Fragment = ReactJsComponentFactoryProxy(React.Fragment);
 
+/// Suspense lets you display a fallback UI until its children have finished loading.
+///
+/// See: <https://react.dev/reference/react/Suspense>
+var Suspense = ReactJsComponentFactoryProxy(React.Suspense);
```
joebingham-wk, line 71: I know it's not going public yet, but when it does, I still think there should be an extended doc comment with consumption guidelines talking about using `Lazy`, or not if it's not public, so folks can easily find how to convert the JS paradigm to Dart! Let me know (whenever you circle back here) if there was a reason that was included when you addressed the first round feedback!

joebingham-wk (changes requested):

example/lazy/index.html
kealjones-wk: I was actually not showing it quite yet just because `lazy` isn't really like "public" yet... But i guess i could be convinced since it demos suspense... I honestly even considered not including this but i used to for manual testing... and thought it would be a shame to lose it lol

lib/react.dart
kealjones-wk: There is a long and lovely doc comment in the over_react version already: https://github.com/Workiva/over_react/pull/820/files#diff-0b321057dde5988d314e11179540acd2207feda3eba1af26f6ac81bd04f603a6R24-R55

eat your heart out

joebingham-wk: > eat your heart out

I would if I felt you addressed my feedback 😜 I won't push it more after this thought, but IMO if we aren't going to put the same level of effort into react-dart as we do OverReact, in terms of comments and API guidelines, then they shouldn't be separate packages. I think the API as is in react-dart is pretty confusing, which is lame for an OSS package

example/lazy/index.html
joebingham-wk: It's as much a `Suspense` demo as it is a lazy one, and `Suspense` is public 🙂 I think it also just makes our life harder  as well, so I'd still vote for it, but it's probably not worth blocking the PR over!

joebingham-wk (approved):
Tests, interop, and demo look good!!

+10

lib/react.dart
kealjones-wk: bro, same. lol I think react-dart is a waste of time, especially as dart evolves (@greglittlefield-wf probably feels very different) but PERSONALLY it seems like a somewhat unnecessary step between us and the JS, but also there is so much like history there that removing the react-dart dependency would probably break so much shit it would never work lol.

joebingham-wk (dismissed):
+10 refresh, looking awesome!!

lib/react.dart
```diff
@@ -65,6 +65,38 @@ typedef ReactDartFunctionComponentFactoryProxy FunctionComponentRegistrar(DartFu
...
 var Fragment = ReactJsComponentFactoryProxy(React.Fragment);
 
+/// [Suspense] lets you display a fallback UI until its children have finished loading.
+///
+/// Like [react.Fragment], [Suspense] does not render any visible UI.
+/// It lets you specify a loading indicator in case some components in
+/// the tree below it are not yet ready to render.
+/// [Suspense] currently works with:
+/// - Components that use React.lazy
+///   - (dynamic imports, not currently implemented in dart)
```
joebingham-wk, line 75: Should this be a nested indent?

```suggestion
/// - (dynamic imports, not currently implemented in dart)
```

kealjones-wk: that was intentional as it is a clarification on the point above it.

joebingham-wk (approved):
+10

kealjones-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

lib/react.dart
greglittlefield-wf: The separation of react-dart and over_react is mostly: 
a) tech debt (these started as separate packages, mainly because we didn't own react-dart)
b) organizationally/architecturally, react-dart works well (for the most part) as a separate layer from over_react, handling most of the JS interop, and over_react mostly being the typed props and factory layer on top of that. Even if we were to combine them into a single repo and make most of the react-dart APIs private, we'd probably maintain most of this separation.

The effort required to maintain the two of them separately hasn't been enough at this point to justify the effort to consolidating them. If we find having them be separate is too painful, I'm happy to discuss!

>  but IMO if we aren't going to put the same level of effort into react-dart as we do OverReact, in terms of comments and API guidelines, then they shouldn't be separate packages. I think the API as is in react-dart is pretty confusing, which is lame for an OSS package

That's a fair point. I think the reality here is that we don't expect people to use react-dart directly most of the time, and instead to use over_react. Perhaps to help address this while these packages are separate, we could just be more forthright about how much effort we put into this project's developer experience, and recommend using over_react instead of using this project directly.

PR #366, opened 2023-08-21
FED-1569 Prepare for null safety: fix lints and implicit casts

greglittlefield-wf:
## Motivation
A while back @aaronlademann-wf started [a branch](https://github.com/Workiva/react-dart/compare/use-workiva-analysis-options) to update react-dart's to use workiva_analysis_options and updated lints, and clean things up.

I then based off that branch, made some adjustments, and enabled implicit casts to improve type safety and make things easier when migrating to null safety. I then used that as a base for a null safety branch.

As a result, all of the null safety changes are dependent on these unmerged lint and implicit cast changes.

To get this repo in a better state, and to enable the merge of null safety changes, let's land these lint and implicit cast changes.

## Changes

The original commits also ended up getting pretty messy, with some larger commits and quite a few merge commits in between. The changes also were based on workiva_analysis_options v1 instead of v2.

So, I ended up redoing/rebasing most of these changes, splitting them up into very granular and easy-to-review commits, and rebasing the null safety branch on top of those. Shoutout to @aaronlademann-wf for your original work on this! ❤️ 

For review, I _**highly recommend**_ reviewing the changes commit-by-commit since there are so many different lint fixes that touch a lot of code. There's a lot of commits, but most of them are either small or `dart fix --apply` ones that are very easy to review.

Summary of changes:
- Upgrade lint configuration to workiva_analysis_options v2, and customize some lints
- Fix lints
    - The vast majority of these are individual commits using `dart fix --apply`
- Fix implicit casts
    - This was done primarily in one commit, along with related adjustments make variables `final` and remove LHS typing where possible
    - The vast majority of these changes occurred in `example` and `test` directories; changes to `lib` were minimal, so the risk of regression here should be pretty low so long as the `lib` changes are reviewed thoroughly.
    - I couldn't disable implicit casts an analysis_options since we'd still get a lot of warnings as a result of `ReactComponentFactoryProxy.call` returning `dynamic` as opposed to `ReactElement`. I'd like to change that typing, but since that's technically a major change, let's save that for another PR.
- Stop running CI on Dart 3 until we're null-safe (it was failing on this PR)

## Testing
- CI passes
- Consumer tests [in over_react](https://github.com/Workiva/over_react/compare/master...test_consume_workiva/react-dart_366_over_react) pass


aviary-wf:
## Security Insights

## Action Items
---

.github/workflows/dart_ci.yml
greglittlefield-wf, line 1: # Reminder to review commits one-by-one!

This should make review much easier. See PR description for more info

lib/react.dart
```diff
@@ -2116,207 +2116,3 @@ dynamic view = validateJsApiThenReturn(() => ReactDomComponentFactoryProxy('view
 
 /// The SVG `<vkern>` [VkernElement].
 dynamic vkern = validateJsApiThenReturn(() => ReactDomComponentFactoryProxy('vkern'));
-
-/// Create React DOM `Component`s by calling the specified [creator].
-_createDOMComponents(creator) {
-  a = creator('a');
-  abbr = creator('abbr');
-  address = creator('address');
```
aaronlademann-wf, line 2104: How long has this been hanging around unused?

lib/react.dart
```diff
@@ -683,7 +683,7 @@ abstract class Component2 implements Component {
   /// name of the proxying JS component defined in _dart_helpers.js).
   @override
   String get displayName {
-    var value;
+    String value;
```
aaronlademann-wf, line 686: I'm assuming there's no reason to worry about this type being tightened?

lib/react_client/js_interop_helpers.dart
```diff
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'dart:js_util';
 
 import 'package:js/js.dart';
+import 'package:react/react_client/react_interop.dart' show forwardRef2;
```
aaronlademann-wf, line 10: Note for posterity: this was added for a reference in a doc comment.

lib/src/react_client/dart_interop_statics.dart
```diff
@@ -27,14 +28,15 @@ final ReactDartInteropStatics dartInteropStatics = (() {
           jsThis.setState(newObject());
         }
 
-        getRef(name) {
+        // ignore: omit_local_variable_types, prefer_function_declarations_over_variables
+        final RefMethod getRef = (name) {
```
aaronlademann-wf: #nit could use a comment explaining why this isn't a function declaration

lib/src/react_client/factory_util.dart
```diff
@@ -102,7 +102,7 @@ dynamic generateChildren(List childrenArgs, {bool shouldAlwaysBeList = false}) {
 
   if (children == null) {
     children = shouldAlwaysBeList ? childrenArgs.map(listifyChildren).toList() : childrenArgs;
-    markChildrenValidated(children);
+    markChildrenValidated(children as List);
```
aaronlademann-wf, line 104: Should this still get cast `as List` if `shouldAlwaysBeList` is false?

lib/src/react_client/synthetic_data_transfer.dart
```diff
@@ -104,8 +104,8 @@ SyntheticDataTransfer syntheticDataTransferFactory(Object dt) {
   }
 
   // Copy these lists and ensure they're typed properly.
-  final files = <File>[...?rawFiles];
-  final types = <String>[...?rawTypes];
+  final files = List<File>.from(rawFiles ?? []);
+  final types = List<String>.from(rawTypes ?? []);
```
aaronlademann-wf, line 108: Shouldn't this be

```suggestion
  final files = List.from(rawFiles ?? []).cast<File>();
  final types = List.from(rawTypes ?? []).cast<String>();
```

aaronlademann-wf (commented):
Mostly some questions...

Other than that, I can plus one / ten this after that...

lib/react.dart
greglittlefield-wf: Looks like almost 3 years 💀 https://github.com/Workiva/react-dart/commit/1ef1196679239187f45d277de2123994785ffdfd#diff-971a483a502cde928eaed632c11bd797a2a1e6c112fb9152dbe374291d7c8dbaL2804

lib/src/react_client/factory_util.dart
greglittlefield-wf: On the line above, the value chlidren is reassigned to is always a `List`, and markChildrenValidated takes in a `List`, but the static type of children is still `dynamic` so we need a cast here. 

In null safety, with Dart's type inference improvements, `children` actually has a static type of `List`, and this becomes an unnecessary cast warning and ends up getting removed.

lib/src/react_client/synthetic_data_transfer.dart
greglittlefield-wf: I think in this case, either would be acceptable. 

`List<File>.from(rawFiles ?? [])` creates a new list and casts objects to the target type as they're copied.

`List.from(rawFiles ?? []).cast<File>()` creates a new `List<dynamic>` and then returns a view on that list that casts items as they're accessed.

Casting them on copy feels better to me in this case, though, so that any bad typing is revealed always, when the event is dispatched, as opposed to only when these properties are accessed.

lib/src/react_client/dart_interop_statics.dart
greglittlefield-wf: Added in [60aec37](https://github.com/Workiva/react-dart/pull/366/commits/60aec372c4860242e91ab34d2c5ec2171ce54239)

aaronlademann-wf (approved):
+1

aaronlademann-wf:
QA +1

- [x] CI Passing
- [x] [over_react consumer CI passing](https://github.com/Workiva/over_react/actions/runs/5931710840)

@Workiva/release-management-pp

sydneyjodon-wk (approved):
+1 on Aaron's commit

rmconsole-wf (approved):
+1 from RM

PR #372, opened 2023-10-02
FEDX-414 Add APIs and update deprecations in preparation for 7.0.0 release

greglittlefield-wf:
## Motivation
react-dart 7.0.0 will contains some breaking changes to APIs that are not yet deprecated in the 6.x line.

The 7.0.0 WIP branch also relies some API additions that aren't present in 6.x, so we'll want to backport them to ensure consumers can transition to these new APIs before setting their lower bounds to 7.0.0.

## Solution
- Deprecate APIs to be removed in 7.0.0
- Update existing outdated deprecations for APIs that say they'll be removed in 7.0.0, but won't yet be removed
    - These were mostly `Component` and its associated APIs, which we don't have a timeline for removing
- Backport APIs `htmlMain` and `useRefInit`
- Add draft changelog entry for 7.0.0.
    - I had this written up but not committed anywhere, and figured I'd add it now to help provide context during review of this PR. We can refine this entry later as part of the PRs for the 7.0.0 changes.

## QA steps
- Verify CI passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,45 @@
+## 7.0.0-wip
+
+- Migrate to null safety
+
+#### Deprecated API removals
+- forwardRef (use forwardRef2 instead)
+- memo (use memo2 instead)
+- main (use htmlMain instead)
+- Ref class constructors: default and `useRefInit` (use useRef/createRef instead)
+- ReducerHook and StateHook class constructors (use hook functions instead).
+- APIs that have been no-ops since react-dart 6.0.0
+    - SyntheticEvent members `persist` and `isPersistent`
+    - unconvertJsEventHandler 
+- APIs that were never intended for public use:
+    - JsPropValidator
+    - dartInteropStatics
+    - ComponentStatics(2)
+    - createReactDartComponentClass(2)
+    - JsComponentConfig(2)
+    - ReactDartInteropStatics
+    - InteropContextValue
+    - markChildrenValidated
+
+#### Other API breakages
+- ReducerHook and StateHook have no public constructors and can no longer be extended
+- Ref.fromJs is now a factory constructor, meaning the Ref class can no longer be extended
+- ReactComponentFactoryProxy.call and .build return type changed from dynamic to ReactElement
+    - This matches the type returned from `build` for all subclasses, which is what’s returned by call, and reflects the type returned at runtime
+    - Has potential to cause some static analysis issues, but for the most part should not affect anything since ReactElement is typically treated as an opaque type
+        - Needs consumer tests
+    - Top-level component factories are typed as ReactDomComponentFactoryProxy instead of being `dynamic`: react.div
+- All PropValidatorInfo arguments are required
+- Changes to public but internal code that should not affect consumers:
+    - ReactDartComponentInternal
+        - Constructor now takes a required argument, props is final
+    - initComponentInternal arguments are typed to reflect runtime assumptions
+- ReactComponentFactoryProxy no longer `implements Function`
+    - This should not be a breakage, since as of Dart 2.0 inheriting from Function has had no effect
+
+#### Potential behavior breakages
+- Component and Component2 members `props`/`state`/`jsThis` are late, will now throw instead of being null if accessed before initialized (e.g., in a constructor, final class field, or static lifecycle method).
+
 ## [6.2.1](https://github.com/Workiva/react-dart/compare/6.2.0...6.2.1)
 - [#366] Fix lints and eliminate most implicit casts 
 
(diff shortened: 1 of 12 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

CHANGELOG.md
```diff
@@ -1,3 +1,45 @@
...
+- All PropValidatorInfo arguments are required
+- Changes to public but internal code that should not affect consumers:
+    - ReactDartComponentInternal
+        - Constructor now takes a required argument, props is final
+    - initComponentInternal arguments are typed to reflect runtime assumptions
+- ReactComponentFactoryProxy no longer `implements Function`
+    - This should not be a breakage, since as of Dart 2.0 inheriting from Function has had no effect
+
+#### Potential behavior breakages
+- Component and Component2 members `props`/`state`/`jsThis` are late, will now throw instead of being null if accessed before initialized (e.g., in a constructor, final class field, or static lifecycle method).
```
aaronlademann-wf, line 41: This is pretty much guaranteed to break `Component` subclasses where lifecycle methods reference those fields, correct?

lib/react.dart
```diff
@@ -308,8 +306,7 @@ abstract class Component {
   /// >
   /// > This was never designed for public consumption, and there will be no replacement implementation in `Component2`.
   /// >
-  /// > Will be removed in `7.0.0` along with `Component`.
-  @Deprecated('7.0.0')
+  /// > Will be removed when `Component` is removed in a future major release.
```
aaronlademann-wf, line 309: is this removal intentional?

lib/react.dart
```diff
@@ -444,19 +439,20 @@ abstract class Component {
   /// > Due to the release of getSnapshotBeforeUpdate in ReactJS 16,
   /// > componentWillUpdate is no longer the method used to check the state
   /// > and props before a re-render. Both the Component class and
-  /// > componentWillUpdate will be removed in the react.dart 7.0.0 release.
+  /// > componentWillUpdate will be removed alongside Component.
   /// > Use Component2 and Component2.getSnapshotBeforeUpdate instead.
   /// >
-  /// > This will be completely removed when the JS side of it is slated for removal (ReactJS 18 / react.dart 7.0.0)
+  /// > This will be completely removed alongside the Component class.
   void componentWillUpdate(Map nextProps, Map nextState) {}
```
aaronlademann-wf, line 446: I'm a little unclear why certain methods are deprecated, and others are not.

aaronlademann-wf (approved):
Couple questions, but I'm +1 here

lib/react.dart
greglittlefield-wf: I think we deprecated only the ones related to the legacy context, and for the rest deprecating `Component` itself seemed sufficient.

lib/react.dart
greglittlefield-wf: Good catch, nope; I'll re-add it

CHANGELOG.md
greglittlefield-wf: They'll be initialized for most lifecycle methods, except for "static" ones like `defaultProps`, `getDerivedStateFromProps`, and `getDerivedStateFromError` where you shouldn't ever reference props/state.

Besides those "static" lifecycle methods, they'll be initialized by the time we get to lifecycle methods.

The sequence during component mount is:
- The Dart component class is instantiated
   - non-`late` class fields are initialized
   - Default constructor is called
- `props` are initialized
- `getInitialState` or `initialState` is called
- `state` is initialized
- All non-static lifecycle methods (`componentWillMount`, `componentDidMount`, `render`, etc.) are called


For example:
```dart
class FooComponent extends Component2 {
  FooComponent() {
    // props would have always been null here, but in 7.0.0 accessing it throws
    print(props);
  }
  render() => 'foo';
}
```
```dart
class FooComponent extends Component2 {
  // props would have always been null when this is initialized, but in 7.0.0 accessing it throws
  final something = (props ?? {})['something'];
  render() => 'foo';
}
```
```dart
class FooComponent extends Component2 {
  getDerivedStateFromProps(Map nextProps, Map prevState)  {
    // props would have always been null here, but in 7.0.0 accessing it throws
    print(props);
  }
}
```

Basically, this shouldn't ever happen unless you're doing something wrong. And if you were doing it wrong, you'd probably get a null exception when trying to read off of them. I think I originally had a note to that effect in here, but scrapped it since I couldn't get wording I was happy with. I'll take a stab at re-adding it.

aaronlademann-wf (approved):
+1

aaronlademann-wf (approved):
+1 / +10

- [x] Passing CI

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #373, opened 2023-10-02
FED-1728 Migrate to null safety

greglittlefield-wf:
## Motivation
This package needs to be upgraded to null safety in order for it to be used in other null-safe packages, and is a prerequisite for being consumed in Dart 3.

This PR contains the bulk of the changes that will make up the react-dart 7.0.0 major release.

A separate PR will be made into the v7_wip branch to perform other planned deprecated API removals. We can do a more complete audit of public API changes either at that point, or in the PR to merge the `v7_wip` branch.

## Changes
- Migrate to null safety
- Additional changes that were prerequisites for null safety
    - Fix `null` key handling in JsBackedMap, required due to null-safe typing of getProperty/setProperty when implicit-casts are disabled, add tests
    - Update mocks to work in null safety by using mockito's builder, update CI and introduce dart_dev for formatting that excludes generated files
- Add new `assert`s that callback ref args aren’t non-nullable, add tests
- Major API changes
   - Update types to help fix implicit casts, as well as generally improve typing
       - Update `ReactComponentFactoryProxy.call` return type from `dynamic` to `ReactElement`
       - Update typing of top-level   DOM factories from `dynamic` to `ReactDomComponentFactoryProxy`
   - API removals (more will be coming in a later PR)
     - Deprecated APIs with replacements: `forwardRef`, `memo`, `main`
     - All public ReducerHook and StateHook constructors
     - All public Ref constructors except for `.fromJs`
- Cleanup
    - Fully disable implicit casts, fix remaining implicit casts not fixed in https://github.com/Workiva/react-dart/pull/366
    - Remove ddc_emulated_function_name_bug - not needed, unnecessary to migrate
    - Tests: sync copy of RefTestCaseCollection w/ copy in over_react null safety branch

## Testing
- Smoke test examples
- CI passes
    - Verify newly-added test files are run in CI
- Verify CI passes on over_react test branches
    - 4.x (current release line) - https://github.com/Workiva/over_react/pull/845
    - null safety WIP branch - https://github.com/Workiva/over_react/pull/828
       - Failures for VM tests (over_react builder tests) are expected and can be ignored; so long as dart2js and DDC tests pass, we're good


aviary3-wk:
## Security Insights

## Action Items
---

example/test/unmount_test.dart
```diff
@@ -23,7 +23,7 @@ void main() {
   print('What');
   final mountedNode = querySelector('#content');
 
-  querySelector('#mount').onClick.listen((_) => react_dom.render(simpleComponent({}), mountedNode));
+  querySelector('#mount')!.onClick.listen((_) => react_dom.render(simpleComponent({}), mountedNode));
```
aaronlademann-wf, line 26: Is there a reason this couldn't just be 

```dart
querySelector('#mount')?.onClick?.listen(//...
```

lib/hooks.dart
```diff
@@ -79,7 +65,10 @@ class StateHook<T> {
 /// ```
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-state.html>.
-StateHook<T> useState<T>(T initialValue) => StateHook(initialValue);
+StateHook<T> useState<T>(T initialValue) {
+  final result = React.useState(initialValue);
+  return StateHook._(result[0] as T, result[1] as void Function(dynamic));
+}
```
aaronlademann-wf, line 71: Should this be

```suggestion
StateHook<T> useState<T>(T initialValue) {
  final result = React.useState(initialValue);
  return StateHook._(result[0] as T, result[1] as void Function(T));
}
```

Or is this not possible because `T` is a dart type, and the function type casting here is representative of a value returned from JS only? If the latter, that might be worth a comment in the code.

lib/hooks.dart
```diff
@@ -142,7 +134,7 @@ StateHook<T> useStateLazy<T>(T Function() init) => StateHook.lazy(init);
 /// ```
 ///
 /// See: <https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects>.
-void useEffect(dynamic Function() sideEffect, [List<Object> dependencies]) {
+void useEffect(dynamic Function() sideEffect, [List<Object?>? dependencies]) {
```
aaronlademann-wf, line 137: Should `dependencies` be defaulted to `const []`?

I ask because I believe `useEffect` lints when `dependencies` is `undefined`.

lib/hooks.dart
```diff
@@ -468,7 +449,7 @@ Ref<T> useRefInit<T>(T initialValue) => Ref.useRefInit(initialValue);
 /// ```
 ///
 /// Learn more: <https://reactjs.org/docs/hooks-reference.html#usememo>.
-T useMemo<T>(T Function() createFunction, [List<dynamic> dependencies]) =>
+T useMemo<T>(T Function() createFunction, [List<dynamic>? dependencies]) =>
```
aaronlademann-wf, line 452: Same thing here - a call to `useMemo` with no dependencies... does nothing. Should dependencies default to `const []`?

lib/react.dart
```diff
@@ -122,7 +120,7 @@ abstract class Component {
   /// [doesn't work for overriding fields](https://github.com/dart-lang/sdk/issues/27452).
   ///
   /// TODO: Switch back to a plain field once this issue is fixed.
-  Map _props;
+  late Map _props;
```
aaronlademann-wf, line 123: IMO, the use of `late` here deserves a comment - even if it isn't part of the documentation comment.

lib/react.dart
```diff
@@ -192,15 +190,13 @@ abstract class Component {
       'Only supported in the deprecated Component, and not Component2. Use createRef or a ref callback instead.')
   set ref(RefMethod value) => _ref = value;
 
-  dynamic _jsRedraw;
+  late Function _jsRedraw;
```
aaronlademann-wf, line 193: Is `Function` as specific as we can be?

I know dynamic call-sites are no bueno for dart2js compile times - so its great that we're getting rid of that here... but are functions with no known arg arity something to avoid if we can help it?

lib/react.dart
```diff
@@ -214,31 +210,29 @@ abstract class Component {
 
   /// Allows the [ReactJS `displayName` property](https://reactjs.org/docs/react-component.html#displayname)
   /// to be set for debugging purposes.
-  String get displayName => runtimeType.toString();
+  String? get displayName => runtimeType.toString();
```
aaronlademann-wf, line 213: TIL Dart's `toString` can return null?

aaronlademann-wf (commented):
Didn't make it very far... will pick this back up tomorrow.

example/test/unmount_test.dart
greglittlefield-wf: Since we always expect the node to be there, I think it throwing if this is null is better behavior than it silently failing.

lib/hooks.dart
greglittlefield-wf: Yeah unfortunately the runtime type of the function is always going to be `Function(dynamic)` since it comes from the JS, so that would be an invalid cast. I can def add a comment 👍

lib/hooks.dart
greglittlefield-wf: I don't think it does lint in that case; you can omit useEffect's dependencies if you want the effect to run after every render (or if you don't care if it runs after every render). 

See docs here: https://react.dev/reference/react/useEffect#useeffect:~:text=optional%20dependencies

lib/hooks.dart
greglittlefield-wf: Yeah this one is a little weird; useMemo is documented as dependencies being required, but the TypeScript definition for it allows `undefined` for some reason, and that got (understandably) translated to an optional arg here.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/90a2a0544f5b47692e7782d7f5c5ebc36da3c0c5/types/react/index.d.ts#L1158-L1165

I think we could make a new, versioned API of this with required `dependencies`, but I don't think defaulting to `const []` in this implementation is a good idea, since it could result in unintentionally stale values if people forget the dependencies.

Also, the over_react analyzer plugin does warn when these aren't specified: https://github.com/Workiva/over_react/blob/8cd3f6cffadfac9bb703af666e904a6c9e965d2a/tools/analyzer_plugin/test/integration/diagnostics/exhaustive_deps_test_cases.dart#L1679

I don't feel super strongly about introducing a new `useMemo` just to make it required, and am fine with it being optional for now, so I'd vote to address that later down the road.

lib/react.dart
greglittlefield-wf: Good catch, yeah I can definitely make this more specific as `void Function()`

lib/react.dart
greglittlefield-wf: Oh it can't; this is actually nullable so that in the Component2 override can return null:
https://github.com/Workiva/react-dart/blob/dbcf43ad127a791f45277bea18fec2825fefa0fa/lib/react.dart#L677-L685

I can add a comment about this.

greglittlefield-wf: Addressed in https://github.com/Workiva/react-dart/pull/373/commits/ee90271ade553281b32779002dbb03325e20db4d

lib/react.dart
greglittlefield-wf: Addressed in https://github.com/Workiva/react-dart/pull/373/commits/ee90271ade553281b32779002dbb03325e20db4d

lib/react.dart
greglittlefield-wf: Addressed in https://github.com/Workiva/react-dart/pull/373/commits/ee90271ade553281b32779002dbb03325e20db4d

lib/hooks.dart
greglittlefield-wf: Oh looks like I was wrong there; `void Function(dynamic)` is technically a subtype of `void Function(T)`, so we could have done that. Looks like we actually do that in `ReducerHook`.

However, in this case we do `dynamic` instead of `T`, because this function can be called with either the value or an updater function. So the typing is really `void Function(T|(T Function(T)))`, which in Dart's type system is only expressible as `void Function(dynamic)`.

I added some comments and a shared typedef for this to help make it clearer: https://github.com/Workiva/react-dart/pull/373/commits/7ab1084bf79ac8049b2921b64aa1088113837060

lib/hooks.dart
aaronlademann-wf: Hmmm... wonder where I got that from... nvm :)

CHANGELOG.md
```diff
@@ -37,8 +36,39 @@
 - ReactComponentFactoryProxy no longer `implements Function`
     - This should not be a breakage, since as of Dart 2.0 inheriting from Function has had no effect
 
-#### Potential behavior breakages
-- Component and Component2 members `props`/`state`/`jsThis` are late, will now throw instead of being null if accessed before initialized (e.g., in a constructor, final class field, or static lifecycle method).
+#### Behavior breakages unlikely to cause issues
+- Component and Component2 members `props`/`state`/`jsThis` are late, will now throw instead of being null if accessed before initialized.
```
aaronlademann-wf, line 40: #nit might be worth linking to the Dart documentation for `late` here?

lib/react.dart
```diff
@@ -453,7 +464,7 @@ abstract class Component {
   /// > This will be completely removed alongside the Component class.
   @Deprecated('This legacy, unstable context API is only supported in the deprecated Component, and not Component2.'
       ' Instead, use Component2.context, Context.Consumer, or useContext.')
-  void componentWillUpdateWithContext(Map nextProps, Map nextState, Map nextContext) {}
+  void componentWillUpdateWithContext(Map nextProps, Map nextState, Map? nextContext) {}
```
aaronlademann-wf, line 467: Is it expected that `nextContext` is nullable here, but not in `shouldComponentUpdateWithContext`?

lib/react.dart
```diff
@@ -769,7 +780,7 @@ abstract class Component2 implements Component {
   /// > [Consider recommended alternative solutions first!](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions)
   ///
   /// See: <https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops>
-  Map getDerivedStateFromProps(Map nextProps, Map prevState) => null;
+  Map? getDerivedStateFromProps(Map nextProps, Map prevState) => null;
```
aaronlademann-wf, line 783: Is it possible for `prevState` to be null on first mount? Or is this lifecycle method not called then?

lib/react_client/bridge.dart
```diff
@@ -42,11 +42,11 @@ abstract class Component2Bridge {
...
   /// __For internal/advanced use only.__
-  static Component2Bridge forComponent(Component2 component) => bridgeForComponent[component];
+  static Component2Bridge? forComponent(Component2 component) => bridgeForComponent[component];
 
-  void setState(Component2 component, Map newState, SetStateCallback callback);
-  void setStateWithUpdater(Component2 component, StateUpdaterCallback stateUpdater, SetStateCallback callback);
-  void forceUpdate(Component2 component, SetStateCallback callback);
+  void setState(Component2 component, Map? newState, SetStateCallback? callback);
+  void setStateWithUpdater(Component2 component, StateUpdaterCallback stateUpdater, SetStateCallback? callback);
+  void forceUpdate(Component2 component, SetStateCallback? callback);
```
aaronlademann-wf, line 49: Should the `callback` args in these three be optional?

aaronlademann-wf: Same question for the impl's below.

lib/react_client/component_factory.dart
```diff
@@ -100,7 +98,7 @@ class ReactDartComponentFactoryProxy<TComponent extends Component> extends React
 
   /// Returns a JavaScript version of the specified [props], preprocessed for consumption by ReactJS and prepared for
   /// consumption by the `react` library internals.
-  static InteropProps generateExtendedJsProps(Map props, dynamic children, {Map defaultProps}) {
+  static InteropProps generateExtendedJsProps(Map props, dynamic children, {Map? defaultProps}) {
     if (children == null) {
       children = [];
```
aaronlademann-wf, line 103: Should this be `const []` to avoid unnecessary redraws, or am I misunderstanding where/how this would be used by the JS pieces under the hood?

lib/react_client/react_interop.dart
```diff
@@ -35,21 +32,21 @@ typedef JsPropValidator = dynamic Function(
...
   @Deprecated('Use createElement instead. To be removed in 7.0.0.')
   external static ReactJsComponentFactory createFactory(type);
-  external static ReactElement createElement(dynamic type, props, [dynamic children]);
+  external static ReactElement createElement(dynamic type, props, [Object? children]);
   external static JsRef createRef();
   external static ReactClass forwardRef(Function(JsMap props, dynamic ref) wrapperFunction);
   external static ReactClass memo(
     dynamic wrapperFunction, [
-    bool Function(JsMap prevProps, JsMap nextProps) areEqual,
+    bool Function(JsMap prevProps, JsMap nextProps)? areEqual,
```
aaronlademann-wf, line 49: Does this mean that someone could pass null as the second optional argument? Would that serve a purpose?

lib/react_client/react_interop.dart
```diff
@@ -59,16 +56,16 @@ abstract class React {
...
-  external static dynamic useMemo(dynamic Function() createFunction, [List<dynamic> dependencies]);
-  external static void useLayoutEffect(dynamic Function() sideEffect, [List<Object> dependencies]);
-  external static void useImperativeHandle(dynamic ref, dynamic Function() createHandle, [List<dynamic> dependencies]);
+  external static JsRef useRef([Object? initialValue]);
+  external static dynamic useMemo(dynamic Function() createFunction, [List<Object?>? dependencies]);
+  external static void useLayoutEffect(dynamic Function() sideEffect, [List<Object?>? dependencies]);
+  external static void useImperativeHandle(dynamic ref, dynamic Function() createHandle, [List<Object?>? dependencies]);
   // NOTE: The use of generics on the `useDebugValue` interop will break the hook.
-  external static dynamic useDebugValue(dynamic value, [Function format]);
+  external static dynamic useDebugValue(dynamic value, [Function? format]);
```
aaronlademann-wf, line 68: Do we know the arity of this function?

lib/react_client/react_interop.dart
```diff
@@ -460,20 +422,20 @@ abstract class ReactDartComponentVersion {
...
+    Function? componentWillReceiveProps,
+    Function? shouldComponentUpdate,
+    Function? componentWillUpdate,
+    Function? componentDidUpdate,
+    Function? componentWillUnmount,
+    Function? getChildContext,
+    Map<String, dynamic>? childContextTypes,
+    Function? getDefaultProps,
+    Function? getInitialState,
+    Function? render,
```
aaronlademann-wf, line 438: Is it worth adding function arity to the function types here?

lib/react_client/react_interop.dart
```diff
@@ -624,17 +586,17 @@ class InteropProps implements JsMap {
   @Deprecated('Only used with the deprecated Component base class and not Component2,'
       ' which stores props as properties on this object.')
   external ReactDartComponentInternal get internal;
-  external dynamic get key;
-  external dynamic get ref;
+  external dynamic /*?*/ get key;
+  external dynamic /*?*/ get ref;
```
aaronlademann-wf, line 590: Should these be `String` / `Object` to match the `InteropProps` argument types?

test/nullable_callback_detection/nullable_callback_detection_sound_test.dart
```diff
@@ -0,0 +1,30 @@
...
+library react.test.unsound_null_safety_test;
+
+import 'package:react/src/react_client/factory_util.dart';
+import 'package:test/test.dart';
+
+import 'null_safe_refs.dart' as null_safe_refs;
+import 'sound_null_safety_detection.dart';
+
+main() {
+  // This test ths the counterpart to nullable_callback_detection_unsound_test.dart,
```
aaronlademann-wf, line 11: #nit 

```suggestion
  // This tests the counterpart to nullable_callback_detection_unsound_test.dart,
```

test/react_client/js_interop_helpers_test.dart
```diff
@@ -53,12 +53,16 @@ class Foo {
...
+Object jsifyAndAllowInteropAsObject(Object value) => jsifyAndAllowInterop(value) as Object;
+
 main() {
   group('jsifyAndAllowInterop', () {
     test('converts a List', () {
       final list = [1, 2, 3, 4, 5, 6, 7, 8];
-      final array = jsifyAndAllowInterop(list);
+      final array = jsifyAndAllowInteropAsObject(list);
       expect(array is List, isTrue);
+      array as List;
```
aaronlademann-wf, line 65: Whoa - is this new? Cast once and be done for the remaining scope?

aaronlademann-wf (commented):
This is great work!!!

lib/react_client/bridge.dart
greglittlefield-wf: They could potentially be made optional, but I think they're required to ensure they're always explicitly specified and not lost along the way. Also, making them optional would be a tricky change to make since this class is extended in over_react.

 so I think we should keep them as-is unless there's a strong reason to make them optional

lib/react_client/component_factory.dart
greglittlefield-wf: Mm good question. I think using `const []` here would avoid unnecessary redraws, but only for legacy `Component` components that have a custom `shouldComponentUpdate` or custom `memo` equality function that doesn't do shallow equality checking of `children` items.

That feels like a pretty low amount of impact, so I'm reluctant to update this behavior in case consumers are incorrectly relying on components rerendering in those cases.

lib/react_client/react_interop.dart
greglittlefield-wf: In Dart null safety, any optional argument without a default needs to be nullable, and as a result, yeah it's possible/valid for someone to call this with a `null` second argument.

lib/react_client/react_interop.dart
greglittlefield-wf: Yeah, it looks like without generics though (which apparently breaks things; see comment on the previous line), we'd have to type it as `Function(Never)? format` to get the typing correct, which is pretty gross, so I'm inclined to leave it as-is.

lib/react_client/react_interop.dart
greglittlefield-wf: This API is deprecated only has a couple remaining usages, so I don't think it's worth it.

lib/react_client/react_interop.dart
greglittlefield-wf: Mm good catch. Since that would be a breaking change, I think I'd rather just update the types of `InteropProps`'s `key/`ref`arguments (which I can't find any usages of FWIW) to `dynamic` instead. 

Or just leave them as-is 🤷.

test/react_client/js_interop_helpers_test.dart
greglittlefield-wf: Yup! There's a bunch of tiny little type promotion improvements like this added alongside null safety; it's so nice!

lib/react.dart
greglittlefield-wf: Good catch. It seems like they'll always be non-null, but for this deprecated API frankly I don't want to go through the trouble of making these non-nullable in a safe way, so I'm just gonna make `shouldComponentUpdateWithContext` nullable to match.

greglittlefield-wf: Addressed in https://github.com/Workiva/react-dart/pull/373/commits/0616cd087f5433c117f09ceb83fcecf19e6665a1

CHANGELOG.md
greglittlefield-wf: Good idea. Added in https://github.com/Workiva/react-dart/pull/373/commits/cb5cfd60ee956f41d8a89144e19d704ce7c160f8

test/nullable_callback_detection/nullable_callback_detection_sound_test.dart
greglittlefield-wf: Oh I think this was supposed to be "This test is"; addressed in https://github.com/Workiva/react-dart/pull/373/commits/cb5cfd60ee956f41d8a89144e19d704ce7c160f8

lib/react.dart
greglittlefield-wf: Good question, it looks like it isn't possible for it to be null. 

On first mount, `prevState` is the value of `initialState`

This behavior is validated here in tests: https://github.com/Workiva/react-dart/blob/ae23239d2ff798baa5ff4a13a4682760d6ed1db5/test/lifecycle_test.dart#L1005

Since intialState is non-nullable and defaults to an empty map, and the underlying component's `state` is always initialized upon mount:
https://github.com/Workiva/react-dart/blob/0616cd087f5433c117f09ceb83fcecf19e6665a1/lib/src/react_client/dart_interop_statics.dart#L219
then it will always be non-null.

This also lines up with TypeScript's typing, showing this arg as the same type as the state generic parameter `S` (and not `S | undefined`): https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f800de4ffd291820a9e444e6b6cd3ac9b4a16e53/types/react/index.d.ts#L710

aaronlademann-wf (approved):
+1

aaronlademann-wf:
+10

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #379, opened 2023-11-02
FED-1850 Add Dart 2.19 to GHA matrix

greglittlefield-wf:
## Motivation
We're only running GitHub Actions in 2.18, and not 2.19 (the latest minor of Dart 2).

## Solution
Add 2.19.6 to the GitHub Actions matrix.

## Testing
CI passes, including new 2.19.6 run.

### DIFF
--- a/.github/workflows/dart_ci.yml
+++ b/.github/workflows/dart_ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         # Can't run on `stable` (Dart 3) until we're fully null-safe.
-        sdk: [2.18.7]
+        sdk: [2.18.7, 2.19.6]
     steps:
       - uses: actions/checkout@v2
       - uses: dart-lang/setup-dart@v1


.github/workflows/dart_ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         # Can't run on `stable` (Dart 3) until we're fully null-safe.
-        sdk: [2.18.7]
+        sdk: [2.18.7, 2.19.6]
```
kealjones-wk, line 19: Is there a reason to keep 2.18.7?

kealjones-wk (approved):
+1 

but one question

.github/workflows/dart_ci.yml
greglittlefield-wf: Mm good catch. I don't think there is.

Though, we do currently format under 2.18.7, so we'd have to update that to run on 2.19. To avoid potential merge conflicts with the v7_wip branch, I'd vote to wait to remove it and switch formatting over until after that's merged.

kealjones-wk: sure

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #381, opened 2023-11-09
FED-1881 Fix regression in SyntheticEvent mock class type-checking

greglittlefield-wf:
## Motivation
When testing these changes, we ran into a case where a consumer was mocking `type`, and relying on type-checking utilities to respect that.
```dart
class MockSyntheticEvent extends Mock implements SyntheticEvent {}
```
```dart
final mockEvent = MockSyntheticEvent();
when(mockEvent.type).then('click');
print(mockEvent.isMouseEvent);
```

Previously, this would print `true`, but in v7_wip, it prints `false`.

## Solution
- Update event type checking to match old behavior.
- Add regression test and additional test

## Testing
- Verify that regression test fails in the commit before the fix was introduced
    - CI link: https://github.com/Workiva/react-dart/actions/runs/6817966382/job/18542598172
    - Expected failing test: `Synthetic event helpers SyntheticEventTypeHelpers checks types correctly for Mock objects with `type` mocked`
- Verify that regression test pass after the fix was introduced (latest CI passes)

### DIFF
--- a/lib/src/react_client/event_helpers.dart
+++ b/lib/src/react_client/event_helpers.dart
@@ -760,11 +760,25 @@ SyntheticWheelEvent createSyntheticWheelEvent({
 }
 
 extension SyntheticEventTypeHelpers on SyntheticEvent {
-  // Use getProperty(this, 'type') since, although statically we may be dealing with a SyntheticEvent,
-  // this could be a non-event JS object cast to SyntheticEvent with a null `type`.
-  // This is unlikely, but is possible, and before the null safety migration this method
-  // gracefully returned false instead of throwing.
-  bool _checkEventType(List<String> types) => getProperty(this, 'type') != null && types.any((t) => type.contains(t));
+  // Access `type` in a try-catch since, although statically we may be dealing with a SyntheticEvent,
+  // this could be an object with a `null` `type` which would cause a type error since `type` is non-nullable.
+  //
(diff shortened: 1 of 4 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/react_client/event_helpers.dart
```diff
@@ -760,11 +760,25 @@ SyntheticWheelEvent createSyntheticWheelEvent({
...
+  // `type` getter and not behave as expected.
+  bool _checkEventType(List<String> types) {
+    String? type;
+    try {
+      // This is typed as null statically, but if it's not, it will probably throw (depending on the compiler).
+      type = this.type;
+    } catch (_) {}
+
+    return type != null && types.any(type.contains);
+  }
```
greglittlefield-wf, line 780: For reference, the current implementation in master: https://github.com/Workiva/react-dart/blob/f4f729cb683c92ab42135bfff93d11eebb4db2ed/lib/src/react_client/event_helpers.dart#L763

robbecker-wf (approved):

robbecker-wf:
QA+1 CI passes with updated/new tests that cover the functionality

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #384, opened 2024-03-01
FED-2298 Add ReactNode typedef

aaronlademann-wf:
### Motivation
In React, a "React node" is a value that can be returned from a component's render or used as children, and contains more values than just `ReactElement`.

TypeScript captures this type as a `ReactNode` type 
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v17/index.d.ts#L214
- https://github.com/facebook/react/blob/b3003047101b4c7a643788a8faf576f7e370fb45/packages/shared/ReactTypes.js#L10

Now that Dart has type aliases _(in SDK 2.13 and up)_, can name this type so that consumers can be more expressive in how they type their code _(and also to help them avoid using `ReactElement`, which in some cases is overly restrictive)_.

### Changes
* Add `ReactNode` typedef, which is an alias for `Object?`.
* Use this new typedef where appropriate


aviary-wf:
## Security Insights

## Action Items
---

test/forward_ref_test.dart
```diff
@@ -30,13 +30,17 @@ main() {
 
     group('sets name on the rendered component as expected', () {
       test('unless the displayName argument is not passed to forwardRef2', () {
-        final ForwardRefTestComponent = forwardRef2((props, ref) {});
+        final ForwardRefTestComponent = forwardRef2((props, ref) {
+          return null;
+        });
```
aaronlademann-wf, line 35: Another nice thing about switching from a `dynamic` return type to `Object?` is that for function components, you get an analysis warning if you forget to return something.

greglittlefield-wf (dismissed):
+1, QAing now...

greglittlefield-wf (approved):
- Changes look good
- Code analyzes and builds without error in over_react and other private consumers

+10

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #398, opened 2024-06-07
FED-2812 Catch MouseEvent.dataTransfer null exception

aaronlademann-wf:
### The Problem

In Dart SDK versions with null-safety enabled, [the interop](https://github.com/dart-lang/sdk/blob/master/sdk/lib/html/dart2js/html_dart2js.dart#L22421-L22422) for `MouseEvent.dataTransfer` is non-nullable despite the [JS spec for `dataTransfer`](https://developer.mozilla.org/en-US/docs/Web/API/DragEvent/dataTransfer) clearly indicating that it can be null.

This can cause the SDK to throw when a `MouseEvent` is manually constructed _(e.g. [in a test utility](https://github.com/workiva/over_react_test/blob/dd90164b94e18faac363ced0dc0508ca780183b0/lib/src/over_react_test/dom_util.dart#L63-L69))_, and the `dataTransfer` getter is then accessed when running with `sound_null_safety` enabled:

```
Unexpected null value encountered in Dart web platform libraries.
```

> [More info about these native non-null assertions can be found here if you're really curious](https://github.com/dart-lang/sdk/blob/master/sdk/lib/html/doc/NATIVE_NULL_ASSERTIONS.md)

### The Solution

While we cannot fix the native `MouseEvent.dataTransfer` issue directly, we _can_ monkey patch this flawed non-null assertion for the `SyntheticMouseEvent.dataTransfer` property that all of our consumer's UI / tests will be accessing as they migrate to null safety.

This PR adds an extension method that catches the Dart SDK's exception in a new `safeDataTransfer` getter, and updates the `wrapNativeMouseEvent` implementation to make `SyntheticMouseEvent.dataTransfer` access the new `safeDataTransfer` getter instead of the native `MouseEvent.dataTransfer` directly.

### QA

1. [ ] Verify that the test added in this PR fails without the new extension method (See failing tests here:  https://github.com/Workiva/react-dart/pull/397)
1. [ ] Verify that the same test is passing in this PR.

### DIFF
--- a/lib/src/react_client/event_helpers.dart
+++ b/lib/src/react_client/event_helpers.dart
@@ -4,6 +4,7 @@ library react_client.event_helpers;
 import 'dart:html';
 
 import 'package:js/js_util.dart';
+import 'package:meta/meta.dart';
 import 'package:react/react_client/js_interop_helpers.dart';
 import 'package:react/src/react_client/synthetic_data_transfer.dart';
 import 'package:react/src/react_client/synthetic_event_wrappers.dart';
@@ -72,7 +73,7 @@ SyntheticMouseEvent wrapNativeMouseEvent(MouseEvent nativeEvent) {
     'clientX': nativeEvent.client.x,
     'clientY': nativeEvent.client.y,
     'ctrlKey': nativeEvent.ctrlKey,
-    'dataTransfer': nativeEvent.dataTransfer,
+    'dataTransfer': nativeEvent.safeDataTransfer,
     'metaKey': nativeEvent.metaKey,
     'pageX': nativeEvent.page.x,
     'pageY': nativeEvent.page.y,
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/react_client/event_helpers.dart
```diff
@@ -823,3 +824,28 @@ extension DataTransferHelper on SyntheticMouseEvent {
...
+  /// `SyntheticMouseEvent`s do not throw the null exception.
+  @internal
+  DataTransfer? get safeDataTransfer {
+    DataTransfer? dataTransfer;
+    try {
+      dataTransfer = this.dataTransfer;
+    } catch (_) {}
+    return dataTransfer;
+  }
+}
```
KealJones, line 851: #nit no end newline?

KealJones (approved):
+1

KealJones (approved):
+10

aaronlademann-wf:
@Workiva/release-management-pp

rmconsole-wf (approved):
+1 from RM

PR #406, opened 2024-09-09
FED-3207 Add react lazy

kealjones-wk:
Adding in a React.lazy implementation 

This is part one of [FED-3114](https://github.com/Workiva/over_react/pull/941)

### DIFF
--- a/lib/react_client/react_interop.dart
+++ b/lib/react_client/react_interop.dart
@@ -19,6 +19,8 @@ import 'package:react/react_client/component_factory.dart' show ReactDartWrapped
 import 'package:react/src/react_client/dart2_interop_workaround_bindings.dart';
 import 'package:react/src/typedefs.dart';
 
+import '../src/js_interop_util.dart';
+
 typedef ReactJsComponentFactory = ReactElement Function(dynamic props, dynamic children);
 
 // ----------------------------------------------------------------------------
@@ -42,6 +44,7 @@ abstract class React {
     dynamic wrapperFunction, [
     bool Function(JsMap prevProps, JsMap nextProps)? areEqual,
   ]);
+  external static ReactClass lazy(Promise Function() loadFunction);
 
   external static bool isValidElement(dynamic object);
 
@@ -269,6 +272,43 @@ ReactComponentFactoryProxy memo2(ReactComponentFactoryProxy factory,
           return areEqual(dartPrevProps, dartNextProps);
         });
   final hoc = React.memo(factory.type, _areEqual);
+  setProperty(hoc, 'dartComponentVersion', ReactDartComponentVersion.component2);
+
+  return ReactDartWrappedComponentFactoryProxy(hoc);
+}
+
+
+/// Defer loading a component's code until it is rendered for the first time.
+///
+/// The `lazy` function is used to create lazy components in React Dart. Lazy components are loaded only when they are rendered for the first time, which can help improve the initial loading time of your application.
+///
+/// To use the `lazy` function, you need to wrap the lazy component with a `Suspense` widget. The `Suspense` widget allows you to specify what should be displayed while the lazy component is loading, such as a loading spinner or a placeholder.
+///
+/// Example usage:
+/// ```dart
+/// import 'package:react/react_client.dart' show lazy, Suspense;
+///
+/// final lazyComponent = lazy(() => import('./path/to/lazy_component.dart'));
+///
+/// // Wrap the lazy component with Suspense
+/// final app = Suspense(
+///   {
+///     fallback: 'Loading...',
+///   },
+///   lazyComponent(),
+/// );
+/// ```
+///
+/// Note: The `lazy` function is part of the `react_client` library in React Dart.
+/// Defer loading a component’s code until it is rendered for the first time.
+///
+/// Lazy components need to be wrapped with `Suspense` to render.
+/// `Suspense` also allows you to specify what should be displayed while the lazy component is loading.
+ReactComponentFactoryProxy lazy(Future<ReactComponentFactoryProxy> Function() loadFunction) {
+  final hoc = React.lazy(allowInterop(() => futureToPromise((() async {
+        final factory = await loadFunction();
+        return jsify({'default': factory.type});
+      })())));
   setProperty(hoc, 'dartComponentVersion', ReactDartComponentVersion.component2);
 
   return ReactDartWrappedComponentFactoryProxy(hoc);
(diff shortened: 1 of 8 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

test/react_lazy_test.dart
```diff
@@ -0,0 +1,26 @@
...
+
+import 'factory/common_factory_tests.dart';
+
+main() {
+  group('lazy', () {
+    group('- common factory behavior -', () {
+      final LazyTest = react.lazy(() async => react.registerFunctionComponent((props) {
+            props['onDartRender']?.call(props);
+            return react.div({...props});
+          }));
```
greglittlefield-wf, line 16: Can we also add some test coverage for lazy-wrapping a JS component (since it doesn't have the same factory proxy)?

greglittlefield-wf: I started spiked out what that would look like locally, since I wanted to be more specific than just "add more coverage please" lol, and wanted to make sure it was a reasonable ask if things got hairy.

In doing so, I realized there were some things like event and ref wrapping that weren't behaving correctly, which was a bit concerning. 

I started debugging and eventually realized that we could just add a wrapper in the react-dart layer, like you did [in over_react](https://github.com/Workiva/over_react/pull/941/files#diff-2a180371d85a0d65907ef08aee9aace73e38275f26b90e14fd457a7a0b8533d5R84)!

That fixed, like, all of the test failures I was seeing.

See this PR here with my proposed added tests, and the wrapper in react-dart: https://github.com/Workiva/react-dart/pull/407

Thoughts on putting that wrapper here instead? Without it, react-dart's `lazy` doesn't really work with JS components. I think it makes sense conceptually too, so that the concerns of the interop details stay in react-dart instead of bleeding into react-dart. And we'll get to run all of these additional shared react-dart tests on the implementations, too.

greglittlefield-wf (changes requested):

test/react_lazy_test.dart
```diff
@@ -0,0 +1,26 @@
...
+library react.react_lazy_test;
+
+import 'package:react/react.dart' as react;
+import 'package:react/react_client/react_interop.dart';
+import 'package:test/test.dart';
+
+import 'factory/common_factory_tests.dart';
+
+main() {
+  group('lazy', () {
```
greglittlefield-wf, line 11: Oh, one more piece of test coverage: could we test that if the future returned by lazy component errors, that that case is handled (to verify the Dart Future error rejects the JS promise correctly)?

When that happens, React should throw an error in the component that propagates to the error boundary.

react-dart doesn't have a canonical ErrorBoundary utility component like over_react does, but there's [this one](https://github.com/Workiva/react-dart/blob/91c9c2cf1eb1bc224f7078d2b6ca06cce8b2f259/test/js_builds/shared_tests.dart#L205-L223) that could be copied over or split out into a shared test utility file.

test/react_lazy_test.dart
kealjones-wk: makes way more sense to put it here, does that mean we can remove it from the over_react impl?

greglittlefield-wf: Yeah, that should make the over_react one unnecessary, and I think we should remove it if possible

lib/react_client/react_interop.dart
```diff
@@ -274,6 +277,52 @@ ReactComponentFactoryProxy memo2(ReactComponentFactoryProxy factory,
...
+///   },
+///   lazyComponent({}),
+/// );
+/// ```
+///
+/// Defer loading a component’s code until it is rendered for the first time.
+///
+/// Lazy components need to be wrapped with `Suspense` to render.
+/// `Suspense` also allows you to specify what should be displayed while the lazy component is loading.
+ReactComponentFactoryProxy lazy(Future<ReactComponentFactoryProxy> Function() load) {
```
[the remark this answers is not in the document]
kealjones-wk: i only put it in here since this is where forwardRef2 and memo2 and all that crap lives. But yeah fs i can use a seperate file.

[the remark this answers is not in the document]
greglittlefield-wf: Mm gotcha; yeah, I think ideally they wouldn't be in here, sorry about that. Thanks!

test/react_lazy_test.dart
kealjones-wk: done

lib/react_client/react_interop.dart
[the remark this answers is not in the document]
kealjones-wk: I put it in its own file and exported that from `react.dart` lemme know if you wanted it somewhere else.

greglittlefield-wf (approved):
LGTM!

This looks ready to merge, but before we do that, let's make sure that the over_react PR doesn't need any additional changes (mainly that [this thread](https://github.com/Workiva/over_react/pull/941#discussion_r1784943107) around the wrapper is addressed)

lib/src/react_client/lazy.dart
```diff
@@ -0,0 +1,67 @@
...
+          // have access to the original factory proxy outside of this async block.
+          final wrapper = forwardRef2((props, ref) {
+            final children = props['children'];
+            return factory.build(
+              {...props, 'ref': ref},
+              [
+                if (children != null && !(children is List && children.isEmpty)) children,
+              ],
+            );
+          }, displayName: 'Lazy(${getProperty(factory.type as Object, 'name') ?? 'Anonymous'})');
```
greglittlefield-wf, line 55: #nit might be nice to handle the displayName case too
```diff
diff --git a/lib/src/react_client/lazy.dart b/lib/src/react_client/lazy.dart
index b1958f0..0015fea 100644
--- a/lib/src/react_client/lazy.dart
+++ b/lib/src/react_client/lazy.dart
@@ -52,7 +52,7 @@ ReactComponentFactoryProxy lazy(Future<ReactComponentFactoryProxy> Function() lo
                 if (children != null && !(children is List && children.isEmpty)) children,
               ],
             );
-          }, displayName: 'Lazy(${getProperty(factory.type as Object, 'name') ?? 'Anonymous'})');
+          }, displayName: 'Lazy(${getComponentName(factory.type) ?? 'Anonymous'})');
           return jsify({'default': wrapper.type});
         }),
       ),
@@ -65,3 +65,16 @@ ReactComponentFactoryProxy lazy(Future<ReactComponentFactoryProxy> Function() lo
   setProperty(hoc, 'dartComponentVersion', ReactDartComponentVersion.component2);
   return ReactDartWrappedComponentFactoryProxy(hoc);
 }
+
+String? getComponentName(Object? type) {
+  if (type == null) return null;
+
+  if (type is String) return type;
+
+  final name = getProperty(type, 'name');
+  if (name is String) return name;
+
+  final displayName = getProperty(type, 'displayName');
+  if (displayName is String) return displayName;
+
+  return null;
```

greglittlefield-wf (approved):
+10

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #409, opened 2024-10-10
FED-3253 Officially support Dart 3

greglittlefield-wf:
Related over_react PR: https://github.com/Workiva/over_react/pull/958

## Motivation

We should support Dart 3.

react-dart can currently be consumed in Dart 3, despite not supporting it in its SDK constraints, due to the way [Dart 3 treats null-safe packages as compatible](https://dart.dev/resources/dart-3-migration#dart-3-backwards-compatibility). 

However, we don't officially support Dart 3, and our automated CI checks don't validate that it works properly in Dart 3.

Also, we still need to support Dart 2 for the time being, so ideally we'd be able to support both Dart 2 and 3 in the same react-dart version, and run CI against each Dart version.

## Changes
- Raise SDK constraints to explicitly support Dart 3
- Update CI to run against both Dart 2 and Dart 3
    - In Dart 3 runs, delete non-null-safe browser test cases, since they don't apply in Dart 3 and won't compile
    - Work around test hanging issue in Dart 3
    - Break SBOM job out of matrix, since it can only get run once
- dev_dependencies: widen build_web_compilers to allow versions needed by Dart 3

> [!CAUTION]
> There's a bug in Dart >=3.3.0 <3.5.0 that breaks the behavior in DDC of certain functions passed to components, such as refs and potentially JS callback props: https://github.com/dart-lang/sdk/issues/56897
>
> We can't prevent users on those versions from pulling in react-dart, since they can already resolve to [null-safe versions](https://dart.dev/resources/dart-3-migration#dart-3-backwards-compatibility) that have already been published.
>
> So, consumers will just have to avoid those versions of Dart, or use dart2js during development or tests. 
>
> If needed, we could potentially follow up with a warning that detects this bug and provides a helpful error.

### QA Checklist
- CI runs in both Dart 2 and 3, and all steps pass

### DIFF
--- a/.github/workflows/dart_ci.yml
+++ b/.github/workflows/dart_ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [2.19.6, 3.4.4, stable]
+        sdk: [2.19.6, 3.2.6, 3.4.4, stable]
     steps:
       - uses: actions/checkout@v2
       - id: setup-dart
--- a/.github/workflows/dart_ci.yml
+++ b/.github/workflows/dart_ci.yml
@@ -10,6 +10,23 @@ on:
       - '**'
 
 jobs:
+  # Run as a separate job outside the Dart SDK matrix below,
+  # since we can only emit a single SBOM.
+  create-sbom-release-asset:
+    name: Create SBOM Release Asset
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: dart-lang/setup-dart@v1
+        with:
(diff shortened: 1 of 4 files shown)


aviary-wf:
## Security Insights

## Action Items
---

robbecker-wf (dismissed):

robbecker-wf:
QA+1 CI passes. Verified that Dart 3 CI runs actually run on Dart 3 and analyzer 6.

robbecker-wf:
can ignore semgrep ❌

robbecker-wf (approved):
+10

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #413, opened 2025-01-27
FED-3282 Suppress react_dom render warnings

greglittlefield-wf:
## Motivation
In React 18, each react_dom.render call emits the following `console.error` message:
> Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot

Since we likely won't be adding support for `createRoot` until after we've switched to React 18, consumers won't be able to address these errors, and they'd be quite noisy in the meantime.

## Solution
- Suppress `console.error` messages warning about ReactDOM.render's deprecation
- Add regression test
- Remove / repurpose old `console.warn` wrapping for similar message suppression, which isn't needed in Dart 2.14+
   - Up SDK constraint to 2.14 (could probably be higher, but that's out of scope)
   - Remove old "legacy" test configs

## Testing instructions
- CI passes
- Regression test, added in 52f1a563d8889401c9960d0acfce0a4577bdbf79, fails before the fix was added


rmconsole-wf (approved):
+1 from RM

PR #414, opened 2025-03-25
FED-3283: React 18 js files

kealjones-wk:
- Upgraded React to 18 (most of that work was done in https://github.com/Workiva/react-dart/pull/365
  - a few of those changes were reverted to remove the addition of api's and keep documentation using reactDom.render.
- Migrate from yarn to pnpm
- Migrate from webpack to Vite
- Update `js_src` files to use module syntax and do a bit of general cleanup.
- Generate `lib/js/react.dev.js` and `lib/js/react.min.js` (with sourcemap for min version) 
- Remove corejs polyfilling (wdesk_login is the only place that needs to support ie11 and it has its own polyfilling)
  - React 16 Polyfill requirements: https://reactjs.org/docs/javascript-environment-requirements.html
    - import 'core-js/es/map'; 
      - Since chrome 38, https://caniuse.com/mdn-javascript_builtins_map
    - import 'core-js/es/set';
      - Since chrome 38, https://caniuse.com/mdn-javascript_builtins_set
  - Know required Polyfill's for dart side usage
    - import 'core-js/stable/reflect/delete-property'; 
      - Since chrome 49, https://caniuse.com/?search=Reflect.deleteProperty
    - import 'core-js/stable/object/assign'; 
      - Since chrome 45, https://caniuse.com/mdn-javascript_builtins_object_assign
   - Used by dart_env_dev.js
     - import 'core-js/stable/string/starts-with';
     - Since chrome 41, https://caniuse.com/mdn-javascript_builtins_string_startswith
- Leaving the `baseURI` polyfill in, just in case.

- Update Dart test setup to use `test_html_builder` to make it easier to swap template being used.
- Update CI matric to run tests against both react 17 and react 18 assets.
- Added explicit react 18 bundle tests 
- Added tags `react-17` (wont run when using react 18) and `react-18` (wont run when using react 17)
- Added test to explicitly check the version of react being used in tests.


build.yaml
```diff
@@ -1,10 +1,11 @@
 targets:
   $default:
     builders:
-      mockito:mockBuilder:
-        # Scope only to files declaring mocks, for performance.
-        generate_for:
-          - test/mockito.dart
```
kealjones-wk, line 7: we don't use mockito? and it was throwing build warnings that its unused

js_src/dart_env_dev.mjs
```diff
@@ -1,10 +1,3 @@
-if (typeof window.MemoryInfo == "undefined") {
-  if (typeof window.performance.memory != "undefined") {
-    window.MemoryInfo = function () {};
-    window.MemoryInfo.prototype = window.performance.memory.__proto__;
-  }
-}
-
```
kealjones-wk: https://groups.google.com/a/dartlang.org/g/announce/c/YtDxtYwvuBA/m/y_oScJcVAgAJ?pli=1

its fixed in chrome 87 and dart 2.9.3 we shouldn't need this anymore right @greglittlefield-wf

build.yaml
```diff
@@ -1,10 +1,11 @@
...
+      test_html_builder:
+        options:
+          templates:
+            "test/templates/html_template.html":
+              - "test/**_test.dart"
```
kealjones-wk, line 8: Moved to using test_html_builder to make it easier to swap js files that were being used for testing with react 17 and 18 in ci

lib/src/react_client/private_utils.dart
```diff
@@ -57,7 +57,7 @@ void validateJsApi() {
     registerComponent2(() => _DummyComponent2());
     _isJsApiValid = true;
```
greglittlefield-wf, line 58: #idea One thing I was thinking of that would be helpful for QAing this PR, but also would be helpful in consumer repos. While we have this dual-version setup, it might be nice to log the version of React being used so we can easily validate what tests are running with

Unfortunately just a `console.log` in the JS bundle won't make its way to the Dart test output, and it needs to be a `print` statement triggered from a test Zone.

I was just thinking, that most usages of this package indirectly call this function pretty early on, and this might actually be a good place to stick it... For example:
```dart
    if (inReactDevMode) print('React version: ${React.version}');
    _isJsApiValid = true;
```

That might need a little more discovery and could be done in a follow-up though. Hmm yeah and thinking about it, one downside is it would get logged for every test file that gets run, which might not be ideal... 

Yeahh maybe we do that as a follow-up haha

greglittlefield-wf (commented):
This looks amazing! Just one tiny comment on the README.

Didn't get a chance to QA, but I can do that next week!

.github/workflows/dart_ci.yml
```diff
@@ -49,6 +50,11 @@ jobs:
           if [[ "$DART_VERSION" =~ ^3 ]]; then
             ./tool/delete_dart_2_only_files.sh
           fi
+      - name: Switch to React 17 Test HTML
+        if: ${{ matrix.react == 17 }}
+        run: |
+          mv test/templates/html_template.html test/templates/html_template-old.html
+          mv test/templates/html_template-react17.html test/templates/html_template.html
```
greglittlefield-wf, line 57: Oh one more thing I thought of: could we add some automated validation that this swap works, and tests are running against the React version we expect?

E.g., a test with some tags like described here: https://github.com/Workiva/react_testing_library/blob/dac2b64da94f6d1e2486a5a0b98caadaa40d72b6/test/unit/react_version_test.dart#L14-L23

greglittlefield-wf (approved):
- Changes look good
    - React-redux version matches between yarn.lock/pnpm.lock - upgrading this requires a separate effort
    - No changes to existing (React 17) JS files
- JS assets and lockfile are up to date
- Dual version-testing setup
    - Verified tests run with the correct versions
    - Verified test presets work and fail correctly when the wrong version is being used
- over_react CI passes when consuming React 18 assets (along with RTL dual-support branch): https://github.com/Workiva/over_react/pull/977 
- Examples look good

+10

mitchellsharma-wk (approved):
+1

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #419, opened 2025-06-18
Add non-minified React 18 prod bundle

sydneyjodon-wk:


greglittlefield-wf (commented):
One comment; this results in non-react code, like react-redux and react-dart JS, not being minified, but unfortunately doesn't seem to affect React code.

That might be worth moving forward with, but I feel it might not provide too much value to consumers. 🤷  Thoughts?

Otherwise, all the changes look good!

PR #422, opened 2025-08-07
FED-4060 Set up gha-dart-oss

sydneyjodon-wk:
We should set up gha-dart-oss so we can:
- take advantage of shared workflows
- use the publish workflow to auto-publish on merge 🎉 

## QA instructions
- [ ] CI passes
- [ ] All steps previously run in CI are still running even if some of them are in the shared jobs now
- [ ] When this merges, follow along to make sure the package actually publishes


sydneyjodon-wk:
@Workiva/release-management-p

